PR #262, opened 2016-10-25
*: report status to TPR

xiang90:


xiang90:
We will try another approach once kubectl apply works.

Issue #328, opened 2016-11-03
doc: mention GA outgoing traffic


xiang90:

xiang90:
Closed by #330

PR #330, opened 2016-11-03
readme: add GA warning

xiang90:
/cc @philips


xiang90:
Fix https://github.com/coreos/etcd-operator/issues/328

PR #352, opened 2016-11-07
*: create client service for all etcd client endpoints

### LINKED ISSUE #351: cluster: create service for entire cluster
Today the etcd operator creates a service for each cluster member. But, it would be nice to have a service for the entire cluster so that as a client I can connect to a single DNS name to service discover the rest of the peer DNS names.

xiang90:
Fix #351

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -94,7 +94,11 @@ func new(kclient *unversioned.Client, name, ns string, spec *spec.ClusterSpec, s
 		status:    &Status{},
 	}
 	if isNewCluster {
-		var err error
+		err := c.createClientServiceLB()
+		if err != nil {
+			// todo: do not panic!
+			panic("todo:" + err.Error())
+		}
 		if spec.Seed == nil {
 			err = c.newSeedMember()
 		} else {
@@ -245,7 +249,7 @@ func (c *Cluster) migrateSeedMember() error {
 	etcdName := fmt.Sprintf("%s-%04d", c.name, c.idCounter)
 	m := &etcdutil.Member{Name: etcdName}
 
-	svc, err := k8sutil.CreateEtcdService(c.kclient, m.Name, c.name, c.namespace)
+	svc, err := k8sutil.CreateEtcdMemberService(c.kclient, m.Name, c.name, c.namespace)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


pkg/cluster/cluster.go
```diff
@@ -410,11 +414,36 @@ func (c *Cluster) delete() {
...
 	}
+
+	err = c.deleteClientServiceLB()
+	if err != nil {
+		// todo: do not panic!
+		panic(err)
+	}
+}
+
+func (c *Cluster) createClientServiceLB() error {
```
hongchaodeng: "LB" isn't needed. Can you remove it?

pkg/cluster/cluster.go
```diff
@@ -94,7 +94,11 @@ func new(kclient *unversioned.Client, name, ns string, spec *spec.ClusterSpec, s
 		status:    &Status{},
 	}
 	if isNewCluster {
-		var err error
+		err := c.createClientServiceLB()
+		if err != nil {
+			// todo: do not panic!
```
hongchaodeng: Add the todo into panic message.

pkg/cluster/cluster.go
xiang90: We need to capture that this service do roundrobin for N client endpoints. Do you have a better name for this? We already create Client service for each member today.

hongchaodeng: I think it's two separate issues here.

First of all, for internal code, I'm fine with the method name with "LB".

When using an etcd cluster, a name like "client service" is more intuitive to users. The client service for each member is ephemeral and for internal use only. We don't want user to use it.

pkg/cluster/cluster.go
hongchaodeng: This needs to be addressed

xiang90: I do not quite understand. Do you want me to simply add a TODO into the message? Or you want me to address the TODO in this PR? I leave a todo here since we panic all over the place already. It requires a dedicated effort to address the panic problem.

pkg/cluster/cluster.go
hongchaodeng: No change is needed here.

pkg/cluster/cluster.go
hongchaodeng: Just add todo into the message like this:
```
err = fmt.Sprintf("(TODO: don't panic) fail to create client service lb: %v",err)
panic(err) or log.Fatalf(err)
```

hongchaodeng:
LGTM after fixing nits

pkg/cluster/cluster.go
xiang90: sure

pkg/cluster/cluster.go
xiang90: OK.

xiang90:
@etcd-bot retest this please

Issue #349, opened 2016-11-07
e2e flake: backupnow failed: dial tcp ...: i/o timeout


hongchaodeng:
logs
```
time="2016-11-06T23:10:42Z" level=error msg="backupnow (test-etcd-16ss4-backup-tool:19999) request failed: Get http://test-etcd-16ss4-backup-tool:19999/backupnow: dial tcp 10.0.254.28:19999: i/o timeout" 
time="2016-11-06T23:10:42Z" level=error msg="check backup (test-etcd-16ss4-backup-tool:19999): unexpected status code (404 Not Found)" 
```

xiang90:
@hongchaodeng Which e2e test is this from

hongchaodeng:
Flakes in both disaster recovery tests where we create pvc

hongchaodeng:
Some logs from last failure:
```
e2e_test.go:224: fail to make a latest backup: fail to request backupnow: backupnow (10.0.174.234:19999) request failed: Get http://10.0.174.234:19999/backupnow: dial tcp 10.0.174.234:19999: getsockopt: connection refused
```

It seems like connectivity issue.

hongchaodeng:
@xiang90 

It's relying on service IP being propagated to jenkins slave machine. There could be a delay.

We can probably retry on such error.

hongchaodeng:
fixed via #370

PR #354, opened 2016-11-10
*: support cluster pause

xiang90:


xiang90:
@hongchaodeng 

> FAIL: TestDisasterRecovery2Members

test/e2e/e2e_test.go
```diff
@@ -151,6 +151,55 @@ func TestDisasterRecoveryAll(t *testing.T) {
...
+	}
+
+	testEtcd.Spec.Paused = true
+	if testEtcd, err = updateEtcdCluster(f, testEtcd); err != nil {
+		t.Fatalf("failed to pause control: %v", err)
+	}
+
+	// TODO: this is used to wait for the TPR to be updated.
+	// TODO: make this wait for reliable
+	time.Sleep(5 * time.Second)
```
hongchaodeng: Why we wait here? As long as API request returns, it should have been updated?

hongchaodeng:
Failure is related to https://github.com/coreos/etcd-operator/issues/349

test/e2e/e2e_test.go
xiang90: The full TPR update cycle has two phase. API server updates TPR, API server sends TPR updates to the watcher in our operator, the operator sends the update to the cluster monitor routine, the monitor routine finally update its TPR cache (the spec). We need to wait for the full cycle. Basically we can remove this once our operator writes status back. Now we do not.

hongchaodeng:
LGTM
Let me merge this. I will carry the burden to fix test flakes.

PR #358, opened 2016-11-11
cluster: use field logger

xiang90:
/cc @hongchaodeng


pkg/cluster/reconcile.go
```diff
@@ -240,29 +247,26 @@ func (c *Cluster) disasterRecovery(left etcdutil.MemberSet) error {
 	return c.restoreSeedMember()
 }
 
-func RequestBackupNow(httpClient *http.Client, addr string) bool {
+// TODO: make this private
```
xiang90: This is public since e2e needs this. We should not make a func public for testing. We need to keep internal stuff internal.

hongchaodeng:
lgtm

Issue #115, opened 2016-09-15
Problem with etcd in hostNetwork


caseydavenport:
Ran into this today while trying to test the changes from #112 

```
NAME                                READY     STATUS    RESTARTS   AGE
po/etcd-cluster-0000                0/1       Error     0          1m
```

```
2016-09-15 00:49:12.196100 I | etcdmain: etcd Version: 3.0.8
2016-09-15 00:49:12.196134 I | etcdmain: Git SHA: d40982f
2016-09-15 00:49:12.196174 I | etcdmain: Go Version: go1.6.3
2016-09-15 00:49:12.196204 I | etcdmain: Go OS/Arch: linux/amd64
2016-09-15 00:49:12.196232 I | etcdmain: setting maximum number of CPUs to 2, total number of available CPUs is 2
2016-09-15 00:49:12.196591 I | etcdmain: listening for peers on http://0.0.0.0:2380
2016-09-15 00:49:12.196725 I | etcdmain: listening for client requests on 0.0.0.0:2379
2016-09-15 00:49:12.213419 E | netutil: could not resolve host etcd-cluster-0000:2380
2016-09-15 00:49:12.236475 I | etcdmain: stopping listening for client requests on 0.0.0.0:2379
2016-09-15 00:49:12.236515 I | etcdmain: stopping listening for peers on http://0.0.0.0:2380
2016-09-15 00:49:12.236541 I | etcdmain: --initial-cluster must include etcd-cluster-0000=http://etcd-cluster-0000:2380 given --initial-advertise-peer-urls=http://etcd-cluster-0000:2380
```

Looks like it's failing to resolve etcd-cluster-0000.  I'll investigate.

xiang90:
Fixed by #366

bamb00:
Hi,

Is there a fix or workaround for the hostNetwork-true? I'm running  etcd 3.1.6.  What workaround I can apply from Kubernetes?

Thanks in Advance.

hongchaodeng:
@bamb00 
What's the context? What are you trying to do?

bamb00:
The issue is reference in @caseydavenport with the link from kubernetes (kubernetes/kubernetes#17406).  I'm just wondering is there a workaround for this issue.

PR #366, opened 2016-11-15
Support self hosted cluster

xiang90:
1. use host network
2. enable anti-affinity

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -103,7 +103,11 @@ func new(kclient *unversioned.Client, name, ns string, spec *spec.ClusterSpec, s
 			panic("todo:" + err.Error())
 		}
 		if spec.Seed == nil {
-			err = c.newSeedMember()
+			if c.spec.SelfHosted {
+				err = c.newSelfHostedSeedMember()
+			} else {
+				err = c.newSeedMember()
+			}
 		} else {
 			err = c.migrateSeedMember()
 		}
@@ -264,20 +268,18 @@ func (c *Cluster) migrateSeedMember() error {
 	c.logger.Info("created etcd service")
 
 	// Used the service IP as additional peer URL so that the seed member can access it.
-	m.AdditionalPeerURL = fmt.Sprintf("http://%s:2380", svc.Spec.ClusterIP)
-
-	mpurls := []string{fmt.Sprintf("http://%s:2380", m.Name), m.AdditionalPeerURL}
+	m.PeerURLs = []string{fmt.Sprintf("http://%s:2380", m.Name), fmt.Sprintf("http://%s:2380", svc.Spec.ClusterIP)}
 
 	initialCluster := make([]string, 0)
 	for _, purl := range seedMember.PeerURLs {
 		initialCluster = append(initialCluster, fmt.Sprintf("%s=%s", seedMember.Name, purl))
 	}
-	for _, purl := range mpurls {
+	for _, purl := range m.PeerURLs {
 		initialCluster = append(initialCluster, fmt.Sprintf("%s=%s", m.Name, purl))
 	}
 
 	pod := k8sutil.MakeEtcdPod(m, initialCluster, c.name, "existing", "", c.spec)
-	pod = k8sutil.PodWithAddMemberInitContainer(pod, m.Name, mpurls, c.spec)
+	pod = k8sutil.PodWithAddMemberInitContainer(pod, c.spec.Seed.MemberClientEndpoints, m.Name, m.PeerURLs, c.spec)
 
 	if _, err := c.kclient.Pods(c.namespace).Create(pod); err != nil {
 		return err
@@ -330,7 +332,7 @@ func (c *Cluster) migrateSeedMember() error {
 
 	c.logger.Infof("updating the peer urls (%v) for the member %x", resp.Members[0].PeerURLs, resp.Members[0].ID)
 
-	m.AdditionalPeerURL = ""
+	m.PeerURLs = nil
 	for {
 		ctx, cancel = context.WithTimeout(context.Background(), constants.DefaultRequestTimeout)
 		_, err = etcdcli.MemberUpdate(ctx, resp.Members[0].ID, []string{m.PeerAddr()})
@@ -386,8 +388,10 @@ func (c *Cluster) updateMembers(etcdcli *clientv3.Client) error {
 		}
 
 		c.members[m.Name] = &etcdutil.Member{
-			Name: m.Name,
-			ID:   m.ID,
+			Name:       m.Name,
+			ID:         m.ID,
+			ClientURLs: m.ClientURLs,
+			PeerURLs:   m.PeerURLs,
 		}
 	}
 	return nil
--- a/pkg/cluster/reconcile.go
+++ b/pkg/cluster/reconcile.go
@@ -127,6 +127,10 @@ func (c *Cluster) resize() error {
 	}
 
 	if c.members.Size() < c.spec.Size {
+		if c.spec.SelfHosted {
+			return c.addOneSelfHostedMember()
+		}
+
 		return c.addOneMember()
 	}
 
@@ -206,6 +210,10 @@ func removeMember(clientURLs []string, id uint64) error {
 }
 
 func (c *Cluster) disasterRecovery(left etcdutil.MemberSet) error {
+	if c.spec.SelfHosted {
+		return errors.New("self-hosted cluster cannot be recovered from disaster")
+	}
+
 	if c.spec.Backup == nil {
 		c.logger.Errorf("fail to do disaster recovery: no backup policy has been defined.")
 		return errNoBackupExist
--- a/pkg/cluster/self_hosted.go
+++ b/pkg/cluster/self_hosted.go
@@ -1,0 +1,115 @@
+// Copyright 2016 The etcd-operator Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cluster
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/coreos/etcd-operator/pkg/util/constants"
+	"github.com/coreos/etcd-operator/pkg/util/k8sutil"
+	"github.com/coreos/etcd/clientv3"
+	"github.com/pborman/uuid"
+)
+
+func (c *Cluster) addOneSelfHostedMember() error {
+	newMemberName := fmt.Sprintf("%s-%04d", c.name, c.idCounter)
+	c.idCounter++
+
+	peerURL := "http://$(MY_POD_IP):2380"
+	initialCluster := append(c.members.PeerURLPairs(), newMemberName+"="+peerURL)
+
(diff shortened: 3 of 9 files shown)


xiang90:
/cc @hongchaodeng @caseydavenport 

Also fix https://github.com/coreos/etcd-operator/issues/115

hongchaodeng:
@xiang90 
If you want to run specific test on jenkins, you can just change the hack/test:
https://github.com/coreos/etcd-operator/blob/master/hack/test#L48

xiang90:
@hongchaodeng Test failure is unrelated to this pr.

pkg/cluster/reconcile.go
```diff
@@ -244,6 +248,7 @@ func (c *Cluster) disasterRecovery(left etcdutil.MemberSet) error {
 		}
 	}
 	c.members = nil
+	// MAYBE: restore a self-hosted etcd cluster.
 	return c.restoreSeedMember()
```
hongchaodeng: panic if self hosted?

pkg/cluster/self_hosted.go
```diff
@@ -0,0 +1,112 @@
...
+func (c *Cluster) newSelfHostedSeedMember() error {
+	newMemberName := fmt.Sprintf("%s-%04d", c.name, c.idCounter)
+	initialCluster := []string{newMemberName + "=http://$(MY_POD_IP):2380"}
+
+	pod := k8sutil.MakeSelfHostedEtcdPod(newMemberName, initialCluster, c.name, "new", uuid.New(), c.spec)
+	err := k8sutil.CreateAndWaitPod(c.kclient, c.namespace, pod, 30*time.Second)
+	if err != nil {
+		return err
+	}
+	c.idCounter++
```
hongchaodeng: This would introduce the wrong ID counter problem.

xiang90: right. will fix.

pkg/cluster/reconcile.go
xiang90: i do not want to panic the operator no matter what happened.

pkg/cluster/self_hosted.go
```diff
@@ -0,0 +1,112 @@
...
+	c.logger.Infof("added a self-hosted member (%s)", newMemberName)
+	return nil
+}
+
+func (c *Cluster) newSelfHostedSeedMember() error {
+	newMemberName := fmt.Sprintf("%s-%04d", c.name, c.idCounter)
+	initialCluster := []string{newMemberName + "=http://$(MY_POD_IP):2380"}
+
+	pod := k8sutil.MakeSelfHostedEtcdPod(newMemberName, initialCluster, c.name, "new", uuid.New(), c.spec)
+	err := k8sutil.CreateAndWaitPod(c.kclient, c.namespace, pod, 30*time.Second)
```
hongchaodeng: Why we wait here?

pkg/cluster/self_hosted.go
```diff
@@ -0,0 +1,112 @@
...
+	c.idCounter++
+
+	// update pod status to get its IP
+	pod, err = c.kclient.Pods(c.namespace).Get(pod.Name)
+	if err != nil {
+		return err
+	}
+
+	cfg := clientv3.Config{
+		Endpoints: []string{"http://" + pod.Status.PodIP + ":2379"},
```
hongchaodeng: If pod.Status.PodIP is necessary here, we need to loop wait it or reconcile it.

pkg/cluster/self_hosted.go
```diff
@@ -0,0 +1,112 @@
...
+		// waiting for the etcd pod to run.
+		DialTimeout: 10 * constants.DefaultDialTimeout,
+	}
+	etcdcli, err := clientv3.New(cfg)
+
```
hongchaodeng: remove blank line

pkg/cluster/self_hosted.go
xiang90: we need to wait for the pod to be ready to get podIP. see your comment https://github.com/coreos/etcd-operator/pull/366/files/a44f1fb9a71b60318f73f22bfb5d4baf28b087b6?diff=unified#r88142464

pkg/cluster/self_hosted.go
xiang90: sure

pkg/cluster/self_hosted.go
```diff
@@ -0,0 +1,112 @@
...
+	defer etcdcli.Close()
+
+	// wait for the new pod to start and add itself into the etcd cluster.
+	// TODO: do not wait forever.
+	for {
+		err = c.updateMembers(etcdcli)
+		if err == nil {
+			break
+		}
+		time.Sleep(5 * time.Second)
```
hongchaodeng: log error

pkg/cluster/self_hosted.go
xiang90: why? if the pod is up and running, the POD IP should be OK.

/cc @yifan-gu Is this true?

pkg/cluster/self_hosted.go
```diff
@@ -0,0 +1,112 @@
...
+
+	if err != nil {
+		return err
+	}
+	defer etcdcli.Close()
+
+	// wait for the new pod to start and add itself into the etcd cluster.
+	// TODO: do not wait forever.
+	for {
+		err = c.updateMembers(etcdcli)
```
hongchaodeng: Why updateMembers() here?
It should go through reconcile loop and update itself, right?

pkg/cluster/self_hosted.go
yifan-gu: I think @xiang90 is right, the network is setup before the pod is actually running, even it's running in hostnetwork, the pod IP will be non-empty since kubelet will amend it with the host ip.

pkg/util/k8sutil/k8sutil.go
```diff
@@ -285,6 +285,7 @@ func MakeEtcdPod(m *etcdutil.Member, initialCluster []string, clusterName, state
 	if state == "new" {
 		commands = fmt.Sprintf("%s --initial-cluster-token=%s", commands, token)
 	}
+
```
hongchaodeng: revert this?

pkg/util/k8sutil/k8sutil.go
```diff
@@ -297,44 +298,9 @@ func MakeEtcdPod(m *etcdutil.Member, initialCluster []string, clusterName, state
...
-						// probe every 60 seconds
-						PeriodSeconds: 60,
-						// failed for 3 minutes
-						FailureThreshold: 3,
-					},
-					VolumeMounts: []api.VolumeMount{
-						{Name: "etcd-data", MountPath: etcdDir},
-					},
-				},
+				etcdContainer(commands, cs.Version),
```
hongchaodeng: Can we keep it as was in this PR?
Refactor this in another PR?

pkg/cluster/self_hosted.go
hongchaodeng: @yifan-gu @xiang90 
I need to double check this.
Last time my experience in bootkube, same using hostnetwork, wasn't providing the PodIP in pod status at the beginning.
Maybe it's related to "Running".

pkg/util/k8sutil/k8sutil.go
xiang90: this is not a refactor. now we have two functions that share the same code block. we need to avoid the unnecessary duplication.

pkg/util/k8sutil/k8sutil.go
xiang90: sure

pkg/cluster/self_hosted.go
hongchaodeng: I think it's fine here.

My bootkube problem is different due to static pod nature and kubelet sync-ing pod delay.

xiang90:

pkg/util/k8sutil/self_hosted.go
```diff
@@ -0,0 +1,69 @@
...
+				"etcd_node":    name,
+				"etcd_cluster": clusterName,
+			},
+			Annotations: map[string]string{},
+		},
+		Spec: api.PodSpec{
+			Containers: []api.Container{
+				etcdContainer(commands, cs.Version),
+			},
+			RestartPolicy: api.RestartPolicyAlways,
```
hongchaodeng: Why do we make it "always restart"?
Shouldn't controller be responsible for reconciling the state?

hongchaodeng:
In last test run, operator crashes:
```
time="2016-11-16T18:09:12Z" level=info msg="deleting cluster" cluster-name=etcd-test-seed-rhgkt pkg=cluster
panic: grpc: timed out when dialing

goroutine 12 [running]:
panic(0x18e0700, 0xc420401460)
	/usr/local/go/src/runtime/panic.go:500 +0x1a1
github.com/coreos/etcd-operator/pkg/cluster.new(0xc4200d2360, 0xc4205c0500, 0xf, 0xc42000c071, 0xc, 0xc42064d280, 0xc420567980, 0xc4200d24d0, 0xc420274201, 0xc420270101)
	/go/src/github.com/coreos/etcd-operator/pkg/cluster/cluster.go:115 +0x3f1
github.com/coreos/etcd-operator/pkg/cluster.New(0xc4200d2360, 0xc4205c0500, 0xf, 0xc42000c071, 0xc, 0xc42064d280, 0xc420567980, 0xc4200d24d0, 0xc42032b650)
	/go/src/github.com/coreos/etcd-operator/pkg/cluster/cluster.go:77 +0x86
github.com/coreos/etcd-operator/pkg/controller.(*Controller).Run.func2(0xc4203d8de0, 0xc4200d2480)
	/go/src/github.com/coreos/etcd-operator/pkg/controller/controller.go:149 +0x6a5
created by github.com/coreos/etcd-operator/pkg/controller.(*Controller).Run
	/go/src/github.com/coreos/etcd-operator/pkg/controller/controller.go:169 +0x2d6
```

pkg/cluster/cluster.go
```diff
@@ -386,8 +388,10 @@ func (c *Cluster) updateMembers(etcdcli *clientv3.Client) error {
 		}
 
 		c.members[m.Name] = &etcdutil.Member{
```
hongchaodeng: Why not just:
```
c.members[m.Name] = m
```

xiang90: m's type is not etcdutil.Member

pkg/cluster/reconcile.go
hongchaodeng: Can we return an error here instead?
```
if c.spec.SelfHosted{
  return fmt.Errorf("unsupported cluster (%s) self hosted restore seed member", ...)
}
```
We can see the error from logs

pkg/util/k8sutil/self_hosted.go
xiang90: Not for self-hosted. If all the etcd pods go down, operator is down. etcd must restart on its own. If there is a disaster, users are on their own to figure it out for self-hosted cluster.

pkg/cluster/cluster.go
hongchaodeng: It's a pointer, right?
```
c.members[m.Name] = *m
```

hongchaodeng: Right.
I saw from a different method.

pkg/cluster/self_hosted.go
```diff
@@ -0,0 +1,112 @@
...
+	// update pod status to get its IP
+	pod, err = c.kclient.Pods(c.namespace).Get(pod.Name)
+	if err != nil {
+		return err
+	}
+
+	cfg := clientv3.Config{
+		Endpoints: []string{"http://" + pod.Status.PodIP + ":2379"},
+		// wait for the just created etcd pod to be ready to accept requests
+		DialTimeout: 5 * constants.DefaultDialTimeout,
```
xiang90: At some point, i should change gRPC max backoff to 3 second or something. gRPC is too crazy here.

xiang90:
All fixed. CI still failed due to the PV thing.

hongchaodeng:
@xiang90 

Failure seems irrelevant. I have some ideas around on how to fix it.
Let's merge this and continue rest work.

PR #373, opened 2016-11-17
*: move migration into self hosted

xiang90:
(Updated)

Previously, we are migrating the boot etcd onto pod IP network. With this change, we are migrating onto host network.


xiang90:
@hongchaodeng PTAL.

pkg/cluster/self_hosted.go
```diff
@@ -113,3 +114,93 @@ func (c *Cluster) newSelfHostedSeedMember() error {
...
+	for _, purl := range bootMember.PeerURLs {
+		initialCluster = append(initialCluster, fmt.Sprintf("%s=%s", bootMember.Name, purl))
+	}
+
+	// create the first member inside Kubernetes for migration
+	newMemberName := fmt.Sprintf("%s-%04d", c.name, c.idCounter)
+	c.idCounter++
+
+	peerURL := "http://$(MY_POD_IP):2380"
+	initialCluster = append(initialCluster, newMemberName+"="+peerURL)
```
hongchaodeng: Can we abstract the above part into a method like
```
newMemberName, initialCluster := makeMember(bootMemberEndpoint)
```

xiang90: which part?

pkg/cluster/self_hosted.go
```diff
@@ -113,3 +114,93 @@ func (c *Cluster) newSelfHostedSeedMember() error {
...
+	pod := k8sutil.MakeSelfHostedEtcdPod(newMemberName, initialCluster, c.name, "existing", "", c.spec)
+	pod = k8sutil.PodWithAddMemberInitContainer(pod, []string{endpoint}, newMemberName, []string{peerURL}, c.spec)
+
+	err = k8sutil.CreateAndWaitPod(c.kclient, c.namespace, pod, 30*time.Second)
+	if err != nil {
+		return err
+	}
+
+	// update pod status to get its IP
+	pod, err = c.kclient.Pods(c.namespace).Get(pod.Name)
```
hongchaodeng: We can get rid of this by letting CreateAndWaitPod() return pod

pkg/cluster/self_hosted.go
```diff
@@ -113,3 +114,93 @@ func (c *Cluster) newSelfHostedSeedMember() error {
...
+	// delete the boot member from the etcd cluster.
+	ctx, cancel = context.WithTimeout(context.Background(), constants.DefaultRequestTimeout)
+	_, err = etcdcli.Cluster.MemberRemove(ctx, bootMemberID)
+	cancel()
+	if err != nil {
+		return fmt.Errorf("etcdcli failed to remove boot member: %v", err)
+	}
+
+	// TODO: do not wait forever.
+	for {
```
hongchaodeng: use
```
wait.PollImmediate(5*time.Second, 30 * time.Second, func()(bool,error){
  ...
  if err == nil {
    return true, nil
  }
  .. log ...
  return false, nil
})
```

pkg/cluster/self_hosted.go
```diff
@@ -113,3 +114,93 @@ func (c *Cluster) newSelfHostedSeedMember() error {
...
+		DialTimeout: constants.DefaultDialTimeout,
+	}
+	etcdcli, err = clientv3.New(cfg)
+	if err != nil {
+		return err
+	}
+
+	// delete the boot member from the etcd cluster.
+	ctx, cancel = context.WithTimeout(context.Background(), constants.DefaultRequestTimeout)
+	_, err = etcdcli.Cluster.MemberRemove(ctx, bootMemberID)
```
hongchaodeng: Can we abstract the above part into a method
```
removeBootMember()
```

hongchaodeng:
I'm OK with the workflow.
Just a few refactoring comments.

pkg/cluster/self_hosted.go
hongchaodeng: So I feel this func has X parts:
- prepare new member
- create pod/svc for new member
- remove boot member
- update membership

This four parts could be better refactored.

pkg/cluster/self_hosted.go
xiang90: that is on my todo list. i do not want to change that in this pr.

pkg/cluster/self_hosted.go
xiang90: this is no better than a for loop in my opinion.

pkg/cluster/self_hosted.go
hongchaodeng: @xiang90 
Maybe just comment the above four items on this method?
We can merge this PR and then refactor it.

xiang90: I will try to do it in this pr. The problem I do want to solve is that this func duplicates some code from the other two funcs in this file. I need to somehow reuse the code. but it would make this pr messier.

pkg/cluster/self_hosted.go
xiang90: We should probably stop depending on external util funcs. it is small, but when it breaks (like update dependencies) it causes problem to us. if we do want this kind of thing, we should write our own.

hongchaodeng: track: #375

pkg/cluster/self_hosted.go
xiang90: why? what is the difference from removing boot member and removing a normal member? this is just two lines of code. i think we should keep as it is.

hongchaodeng: OK. Fine with me.

pkg/cluster/self_hosted.go
xiang90: I think we should keep as it is now.

part1 is 4 lines of code and it is clear

part2 should be refactored with other funcs. basically we should hide these k8s calls when possible

part3 is 2 line of code and it is clear

part4 should be removed in the near future from this func once reconcile becomes a little smarter.

xiang90:
@hongchaodeng Can we get this merged?

hongchaodeng:
lgtm

PR #383, opened 2016-11-19
*: improve selfhosted test

xiang90:


hongchaodeng:
Test failure might be caused by left over pods from other tests.
Basically each pod will occupy one host.
I just clean up this morning and will rerun it.

hongchaodeng:
@etcd-bot retest this please

test/e2e/self_hosted_test.go
```diff
@@ -74,7 +75,8 @@ func TestCreateSelfHostedClusterWithBootMember(t *testing.T) {
 	defer e.Close()
 
 	<-e.Server.ReadyNotify()
-	fmt.Println("etcdserver is ready")
+
+	t.Log("etcdserver is ready")
```
hongchaodeng: t.Log() will only show after the test finished.
This is not really helpful: 1. debugging; 2. mixing logs with others that already use more real-time logging like fmt, logrus.

Can we keep fmt? I don't mind to change to logrus.

test/e2e/self_hosted_test.go
```diff
@@ -53,14 +53,15 @@ func TestCreateSelfHostedClusterWithBootMember(t *testing.T) {
 	os.RemoveAll(dir)
 	defer os.RemoveAll(dir)
 
+	host, _ := netutil.GetDefaultHost()
+
```
hongchaodeng: For debugging sake, can we print this and also change "hack/test" to only run this test?

test/e2e/self_hosted_test.go
xiang90: print in test is wrong. test should generate no output if it succeeds. if you want to debug, add debugging line and disable by default.

When test fails, it should print out log. So using https://golang.org/pkg/testing/#T.Log is correct. And this is also why t.log designed this way at the first place.

test/e2e/self_hosted_test.go
xiang90: we should print out nothing in test. i can add t.log if you wish to.

hongchaodeng: Let's not add t.Log here.
Could you check err?

test/e2e/self_hosted_test.go
hongchaodeng: OK

hongchaodeng:
I think the error might be caused by that self hosted pods in another test not "removed" yet due to async nature.
There has one pod created. We have exactly 4 worker nodes.
I need to further debug this.

hongchaodeng:
I need to rewrite e2e tests as mentioned in https://github.com/coreos/etcd-operator/issues/385#issuecomment-261753751 to pass the tests again.

The failure here isn't relevant and it would be good to have such refactor work in. Maybe we could just merge this? Do you have any opinion?

hongchaodeng:
@etcd-bot retest this please

hongchaodeng:
@etcd-bot retest this please

hongchaodeng:
@etcd-bot retest this please

Also fixed the firewall issue of backup pod port "19999".

hongchaodeng:
@etcd-bot retest this please

hongchaodeng:
@etcd-bot retest this please

I shut down proxy before. Now restarted. Should work.

hongchaodeng:
tests passed.
Merging.

PR #400, opened 2016-11-28
doc: add cluster status design

xiang90:
/cc @hongchaodeng

### DIFF
--- a/doc/design/cluster_status.md
+++ b/doc/design/cluster_status.md
@@ -1,0 +1,37 @@
+## Cluster status reporting
+
+An etcd cluster permanently fails when all its members are dead and no backup is available. A user might update etcd cluster TPR with invalid input or format. The operator needs to notify users about these bad events.
+
+A common way to do this in the Kuberentes world is through status field, like pod.status. The etcd operator will write out cluster status similar to pod status.
+
+```go
+type EtcdCluster struct {
+	unversioned.TypeMeta `json:",inline"`
+	api.ObjectMeta       `json:"metadata,omitempty"`
+	Spec                 ClusterSpec `json:"spec"`
+	Status               ClusterStatus `json:"status"`   
+}
+
+Type ClusterStatus struct {
+	Paused bool `json:"paused"`
+	...
+}
+```
+
+The etcd operator keeps the truth of the cluster status, and update the TPR item after each cluster reconciliation atomically. Note that users can:
+
+ - modify TPR item concurrently with the operator. 
+ - update TPR and overwrite the status filed with empty or bad input
+
+ The operator MUST handle these two cases gracefully. Thus, the operator MUST NOT overwrite the user input on spec field. The operator CANNOT trust the existing status filed in TPR.
+
+To not overwrite the user input, the etcd operator will set resource version, and do a compare and swap on resource version to update the status. If the compare and swap fails, the operator knows that the user updated the spec concurrently, and it will retry later after get the new spec.
+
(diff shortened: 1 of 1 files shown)


doc/design/cluster_status.md
```diff
@@ -0,0 +1,32 @@
+## Cluster status reporting
+
+An etcd cluster can permanently fail when all members are dead due to massive nodes failure. A user might update etcd cluster TPR with invalid input or format. The operator needs to notify users these bad events.
```
hongchaodeng: "when all members are dead due to massive nodes failure." ->
"when all members are dead and no backup is available."

hongchaodeng: Can we put the last sentence to first:

"The operator needs to notify users bad events. For example, ..."

doc/design/cluster_status.md
```diff
@@ -0,0 +1,32 @@
...
+	Status               ClusterStatus `json:"status"`   
+}
+
+Type ClusterStatus struct {
+	Paused bool `json:"paused"`
+	...
+}
+```
+
+The etcd operator keeps the truth of the cluster status, and update the TRP item after each cluster reconciliation atomically. Note that users can modify TPR item concurrently with the operator. The operator MUST NOT overwrite the user input. The user MIGHT update TPR with empty status. The operator MUST handle these two cases gracefully.
```
hongchaodeng: TRP -> TPR

hongchaodeng: How about:
"Note that we have two corner cases: 1. ...; 2. ..."

doc/design/cluster_status.md
```diff
@@ -0,0 +1,32 @@
...
+
+To not overwrite the user input, the etcd operator will set resource version, and do a compare and swap to update the status. If the compare and swap fails, the operator knows that the user updated the spec concurrently, and it will retry later after get the new spec.
```
hongchaodeng: "To avoid overwriting..."

hongchaodeng: I don't think k8s client supports CAS.

doc/design/cluster_status.md
```diff
@@ -0,0 +1,32 @@
...
+
+To not be affected by the potential empty or broken status accidently wrote by the user, the etcd operator will never read the status from TPR. It always keeps the source of truth in memory.
```
hongchaodeng: "accidentally written ..."

doc/design/cluster_status.md
xiang90: It supports CAS through resource version.

xiang90:

hongchaodeng:
Just for sync-up, can you clarify how to do CAS? I couldn't find any option in:
https://github.com/kubernetes/kubernetes/blob/850729bfaf192a491c38b9bc15744775e4a0f3b1/pkg/client/clientset_generated/release_1_5/typed/core/v1/pod.go#L34-L42

xiang90:
@hongchaodeng 

For any k8s object, it contains a struct called `ObjectMeta`. Inside this struct, there is a field called `ResourceVersion`. This is used for CAS operations. See the comments:

```
// An opaque value that represents the version of this resource. May be used for optimistic
// concurrency, change detection, and the watch operation on a resource or set of resources.
// Clients must treat these values as opaque and values may only be valid for a particular
// resource or set of resources. Only servers will generate resource versions.
```

We have tested it. It works.

hongchaodeng:
LGTM

PR #415, opened 2016-12-01
backup: abstract file backend as backend interface

xiang90:
prepare for a s3 backend.

### DIFF
--- a/pkg/backup/http.go
+++ b/pkg/backup/http.go
@@ -16,8 +16,8 @@ package backup
 
 import (
 	"fmt"
+	"io"
 	"net/http"
-	"path"
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


pkg/backup/http.go
```diff
@@ -87,5 +87,9 @@ func (b *Backup) serveSnap(w http.ResponseWriter, r *http.Request) {
 	if r.Method == http.MethodHead {
 		return
 	}
-	http.ServeFile(w, r, path.Join(b.backupDir, fname))
+
+	_, err = io.Copy(w, rc)
+	if err != nil {
+		logrus.Errorf("failed to write backup to %s: %v", r.RemoteAddr, err)
```
hongchaodeng: What happen to http response on this case?

xiang90: io copy fails. http connection probably fails. we do nothing i guess. but we can try to write a error to http.

xiang90: actually we cannot write http header as long as we ever write one byte in the body.

xiang90: so it does not worth it.

xiang90:
@hongchaodeng Any more comments or concerns that I should address?

hongchaodeng:
lgtm

PR #416, opened 2016-12-01
backup: switch on storage type

xiang90:


hongchaodeng:
lgtm

hongchaodeng:
FYI, we didn't parse it from flags yet.
We would need to parse it when we actually have another option.

xiang90:
@hongchaodeng 

I saw that todo. Can we add a flag -spec-json? And simply marshal json and pass it in?

xiang90:
@hongchaodeng Also I am not sure how are we going to update backup policy on fly. Or users can only choose backup policy when creating the cluster at first time?

xiang90:
```
	e2e_test.go:213: fail to make a latest backup: backupnow (:19999) failed: Get http://:19999/backupnow: dial tcp :19999: getsockopt: connection refused
```

@hongchaodeng any idea?

hongchaodeng:
@xiang90 
Forgot to check pod running.
Will fix it.

hongchaodeng:
@etcd-bot retest this please

PR #418, opened 2016-12-01
backup: add s3 backend

xiang90:
/cc @hongchaodeng

### DIFF
--- a/pkg/backup/s3.go
+++ b/pkg/backup/s3.go
@@ -1,0 +1,75 @@
+// Copyright 2016 The etcd-operator Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package backup
+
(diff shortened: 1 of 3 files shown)


pkg/backup/s3.go
```diff
@@ -0,0 +1,76 @@
...
+		return fmt.Errorf("failed to save snapshot: %v", err)
+	}
+	rc.Close()
+
+	// S3 put is atomic, so let's go ahead and put the key directly.
+	err = sb.S3.Put(key, tmpfile)
+	if err != nil {
+		return err
+	}
+	log.Printf("saved backup %s (size: %d) successfully", key, n)
```
hongchaodeng: use logrus?

xiang90: will do

pkg/backup/s3.go
```diff
@@ -0,0 +1,76 @@
...
+	defer func() {
+		tmpfile.Close()
+		os.Remove(tmpfile.Name())
+	}()
+
+	n, err := io.Copy(tmpfile, rc)
+	if err != nil {
+		return fmt.Errorf("failed to save snapshot: %v", err)
+	}
+	rc.Close()
```
hongchaodeng: This will double close().

xiang90: will io.copy close it for us?

xiang90: io copy takes reader not closer. so why this is double close?

hongchaodeng: This is closed outside in writeSnap()

xiang90: ok. probably we should pass in just reader then. i will clean this up after this pr.

xiang90:

hongchaodeng:
lgtm

PR #429, opened 2016-12-04
README: add a gif demo

philips:
I don't know if we should check this into git but it is only 516KB so it
seems OK.


xiang90:
I think it is OK to check in the demo since it is pretty small.

I think we will add more demos in the near future. For example, there is a youtube demo I probably want to link to soon. So can we make the demo its own section after `Namespace`?

hongchaodeng:
lgtm

PR #435, opened 2016-12-06
cluster: do not intilize cluster on restore case

xiang90:
Close #433

All we need to ensure is that we only initialize PVC and other stuff for a completely new cluster.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -102,22 +102,25 @@ func new(config Config, s *spec.ClusterSpec, stopC <-chan struct{}, wg *sync.Wai
 		status:  &Status{},
 	}
 
-	if err := c.prepareBackupAndRestore(); err != nil {
-		if err == errBackupUnsetRestoreSet {
-			c.logger.Errorf("%v. This is not allowed. Cluster to be dead.", err)
-		}
-		return nil, err
-	}
-
-	if isNewCluster && c.spec.Restore == nil {
-		// Note: For restore case, we will go through reconcile loop and disaster recovery.
-		if err := c.prepareSeedMember(); err != nil {
+	if isNewCluster {
+		err := c.prepareBackupAndRestore()
+		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/cluster/cluster.go
```diff
@@ -102,22 +102,25 @@ func new(config Config, s *spec.ClusterSpec, stopC <-chan struct{}, wg *sync.Wai
 		status:  &Status{},
 	}
 
-	if err := c.prepareBackupAndRestore(); err != nil {
-		if err == errBackupUnsetRestoreSet {
-			c.logger.Errorf("%v. This is not allowed. Cluster to be dead.", err)
+	if isNewCluster {
```
hongchaodeng: I don't think this is equivalent to "creating new cluster".
This var would better be named "IsControllerRefreshingInformation".

I'm mostly concerned with the case that the TPR object was created before etcd operator starts/restarts, and then etcd operator won't do anything.

xiang90: > I don't think this is equivalent to "creating new cluster".

New Cluster = controller receives new cluster event. Other than that, it is not a new cluster from the controller point of view.

> I'm mostly concerned with the case that the TPR object was created before etcd operator starts/restarts, and then etcd operator won't do anything.

If you really want to deal with this, you have to go back to what I have suggested. Check the existence of Backup deployment here.

hongchaodeng: I think the race is rare. Even under such race, we could guide that just delete the cluster and recreate it.

As a result, could we add a log to its "else" case that saying "restoring/recoving/refreshing states"?

xiang90: The logging can be very annoying since the resync is expected. The controller will spam this logging.

One thing we should do is to voluntarily stop the controller and re-watch from the head revision to avoid the out of date event. After we have that mechinism, we can add logging here.

hongchaodeng:
lgtm

PR #469, opened 2016-12-18
cluster: refactor backup deletion

xiang90:


xiang90:
@hongchaodeng After second thought, I feel we should make backup manager non-k8s related initially. We probably need to remove `PodSpecWithStorage(*api.PodSpec) *api.PodSpec` from the interface. 

Later on, we probably should make the backup manager manage different backup backends (pv, s3, etc..). Basically backup manager should not be an interface, but backup backend should be.

xiang90:
Or we will end up with duplicating quite a few k8s related logic in each manager.

hongchaodeng:
> Later on, we probably should make the backup manager manage different backup backends (pv, s3, etc..). Basically backup manager should not be an interface, but backup backend should be.

Sounds good.

hongchaodeng:
lgtm

PR #492, opened 2016-12-21
Updated etc to v3.1.0-rc.1

w0mbat:


hongchaodeng:
We won't update etcd version until 3.1 stable release.
We also need to update vendoring.

w0mbat:
@hongchaodeng ah ok - just thought that the RC would be more stable than the first alpha

hongchaodeng:
@w0mbat 
We haven't seen any major stability problem in etcd3 so far. That includes heavy testing with large scale deployment.

Of course, we encourage your deployment to upgrade to latest etcd version as much as possible. That's why we built etcd operator to do it.

Meanwhile, you can deploy your own yaml.

w0mbat:
@hongchaodeng ok sure - we already used it down-upstream with the rc version

Issue #290, opened 2016-10-29
Garbage services left due to member creation/deletion failure


hongchaodeng:
There could be cases that we didn't remove services successfully and controller continue to run smoothly:
- we create a service but didn't create the pod successfully,
- We failed to remove the service. But etcd pod exited correctly once removed from membership, or etcd pod gets removed due to liveness probe.

Controller will heal the cluster and create more pods and services if needed. Nonetheless, previous services might be useless and left in oblivion.

hongchaodeng:
fixed by #571

hongchaodeng:
Reopen. We need full GC to fix this.

hongchaodeng:
fixed via #578

hongchaodeng:
These two cases aren't fixed:

> - we create a service but didn't create the pod successfully
> - We failed to remove the service

PR #507, opened 2016-12-23
Delete unused services

fatduo:
In some cases, etcd instance's pod and service are not one-to-one.(stop etcd-operator, delete some pods then restart etcd-operator)
So there are many unused etcd services in k8s env.
```
/user/duo/etcd-operator# kubectl get endpoints --namespace=t1
NAME                                   ENDPOINTS                           AGE
etcd-cluster-with-backup               172.16.74.2:2379                    4h
etcd-cluster-with-backup-0000          <none>                              4h
etcd-cluster-with-backup-0002          <none>                              4h
etcd-cluster-with-backup-0004          <none>                              1h
etcd-cluster-with-backup-0005          <none>                              1h
etcd-cluster-with-backup-0006          172.16.74.2:2380,172.16.74.2:2379   1h
etcd-cluster-with-backup-backup-tool   172.16.25.2:19999                   4h
```
So, I think etcd-operator should delete the unused services.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -45,25 +45,24 @@ const (
 
 type clusterEvent struct {
 	typ  clusterEventType
-	spec spec.ClusterSpec
+	spec *spec.ClusterSpec
 }
 
 type Config struct {
-	Name          string
-	Namespace     string
 	PVProvisioner string
 	s3config.S3Context
 	KubeCli *unversioned.Client
 }
 
+// TODO: better rename to ClusterManager
 type Cluster struct {
 	logger *logrus.Entry
 
 	Config
 
-	status *Status
+	*spec.EtcdCluster
 
-	spec *spec.ClusterSpec
+	status *Status
 
 	idCounter int
 	eventCh   chan *clusterEvent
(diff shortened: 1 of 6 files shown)


fatduo:
@xiang90 @hongchaodeng

pkg/cluster/cluster.go
```diff
@@ -378,6 +384,34 @@ func (c *Cluster) deleteClientServiceLB() error {
...
+	s := labels.SelectorFromSet(map[string]string{
+		"etcd_cluster": c.Name,
+		"app":          "etcd",
+	})
+	r, err := labels.NewRequirement("etcd_node", selection.Exists, nil)
+	if err != nil {
+		return err
+	}
+
+	svcs, err := c.KubeCli.Services(c.Namespace).List(k8sapi.ListOptions{
```
hongchaodeng: Isn't this gonna list all services (including used ones) and then delete them?

hongchaodeng:
@fatduo 
Thanks for mentioning this.
We are developing GC which seems to solve this problem as well:
https://github.com/coreos/etcd-operator/issues/290

pkg/cluster/cluster.go
fatduo: This func get and delete all instances' service(func name may be bad). Its fine because I call this func at deleting cluster stage.

hongchaodeng: Ah. I see.
I think the problem is that we only list pods and then call removePodAndService() only on Pod name.
Instead of calling another func to delete services, could you rewrite that part in delete() to list and delete both pods and services?

PR #521, opened 2016-12-26
cluster: add initial gc implementation

xiang90:
/cc @hongchaodeng

### DIFF
--- a/pkg/controller/gc.go
+++ b/pkg/controller/gc.go
@@ -1,0 +1,133 @@
+package controller
+
+import (
+	"github.com/Sirupsen/logrus"
+	k8sapi "k8s.io/kubernetes/pkg/api"
+	"k8s.io/kubernetes/pkg/client/unversioned"
+	"k8s.io/kubernetes/pkg/labels"
+	"k8s.io/kubernetes/pkg/types"
+)
+
+type gc struct {
+	// reference to parent controller
+	controller *Controller
+
+	logger *logrus.Entry
+
+	k8s *unversioned.Client
+	ns  string
+}
+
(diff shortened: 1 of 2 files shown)


xiang90:
@fatduo Check this out. If you want to implement lazy gc, you would need to call collectCluster before creating a new cluster.

fatduo:
@xiang90  Fine. I have created a new PR #523

xiang90:
@fatduo I saw that one. But we want to get this pr merged and use gc.collectCluster to implement lazy gc.

fatduo:
@xiang90 OK. Pls discard that PR. I saw the gc code. Did we have ownerReferences between tpr etcdcluster and it's pods/svcs?

xiang90:
@hongchaodeng is working on resource tracking

pkg/controller/gc.go
```diff
@@ -0,0 +1,133 @@
...
+		}),
+	}
+
+	return gc.collectResources(option, map[types.UID]bool{clusterUID: true})
+}
+
+// fullyCollect collects resources that are created by the controller,
+// but does not belong to any running etcd clusters.
+func (gc *gc) fullyCollect() error {
+	clusters, _, err := gc.controller.getClustersFromTPR()
```
hongchaodeng: IIRC, this will call a list to API server, which could return less fresh data.

We should get the list of clusters from controller directly. Controller has a map of clusters.

xiang90: List without RV will serve from etcd. Freshness will be ensured. We can improve this by serving from operator cache. But now, we do not really store the UID. We can improve this later.

hongchaodeng:
lgtm

PR #530, opened 2016-12-27
spec: rich status

xiang90:
/cc @hongchaodeng

### DIFF
--- a/pkg/spec/cluster.go
+++ b/pkg/spec/cluster.go
@@ -16,6 +16,7 @@ package spec
 
 import (
 	"errors"
+	"time"
 
 	"k8s.io/kubernetes/pkg/api"
 	"k8s.io/kubernetes/pkg/api/unversioned"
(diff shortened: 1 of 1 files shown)


pkg/spec/cluster.go
```diff
@@ -92,9 +93,49 @@ func (c *ClusterSpec) Validate() error {
...
+	ClusterPhaseRunning = "Running"
+	ClusterPhaseFailed  = "Failed"
+)
+
+type ClusterCondition struct {
+	Type ClusterConditionType `json:"type"`
+
+	Reason string `json:"reason"`
+
+	TransitionTime time.Time `json:"transitionTime"`
```
hongchaodeng: What's TransitionTime?

pkg/spec/cluster.go
```diff
@@ -92,9 +93,49 @@ func (c *ClusterSpec) Validate() error {
...
+	ClusterConditionUpgrading = "Upgrading"
+)
+
 type ClusterStatus struct {
+	// Phase is the cluster running phase
+	Phase  ClusterPhase `json:"phase"`
+	Reason string       `json:"reason"`
+
+	// Condition keeps ten most recent cluster conditions
+	Conditions []ClusterCondition `json:"conditions"`
```
hongchaodeng: Why do we need to report an array? What is the array of cond used for?

xiang90: We want to keep a list of conditions.

When user changes size from 3 to 5.

The cluster will turn from condition Ready to ScalingUp. After the cluster has 5 member, the condition will turn from ScalingUp back to Ready.

This is a common practice in k8s world.

pkg/spec/cluster.go
xiang90: When the cluster transited to this condition.

hongchaodeng:
lgtm

PR #531, opened 2016-12-27
cluster: report size change conditions

xiang90:


hongchaodeng:
@etcd-bot retest this please

pkg/spec/cluster.go
```diff
@@ -146,3 +146,44 @@ func (cs *ClusterStatus) SetVersion(v string) {
...
+		Type:           ClusterConditionReady,
+		TransitionTime: time.Now(),
+	}
+
+	if len(cs.Conditions) == 0 {
+		cs.appendCondition(c)
+		return
+	}
+
+	lastc := cs.Conditions[len(cs.Conditions)-1]
```
hongchaodeng: It's unsafe to use this index here.
Maybe:
```go
var lastc *ClusterCondition
if le := len(cs.Conditions); le > 0 {
  lastc = &cs.Conditions[le-1]
}
if lastc != nil && ...
```

pkg/spec/cluster.go
```diff
@@ -146,3 +146,44 @@ func (cs *ClusterStatus) SetVersion(v string) {
...
+
+func (cs *ClusterStatus) AppendScalingDownCondition() {
+	c := ClusterCondition{
+		Type:           ClusterConditionScalingDown,
+		TransitionTime: time.Now(),
+	}
+	cs.appendCondition(c)
+}
+
+func (cs *ClusterStatus) SetReadyCondition() {
```
hongchaodeng: Why this is "Set" but other two "Append"?
Maybe change all of them "Set"?

xiang90: They are different. Append appends the condition to the end of the list.

Set sets the last condition to the desired one. So it only appends if the last condition is not Ready; If the last condition is Ready, it does nothing.

pkg/spec/cluster.go
xiang90: not sure what you mean. i think line172 already does what you want?

hongchaodeng:
lgtm

PR #536, opened 2016-12-27
cluster: add scaleup,down,recovering,upgrading conditions

xiang90:


pkg/cluster/reconcile.go
```diff
@@ -162,6 +156,8 @@ func (c *Cluster) resize() error {
 }
 
 func (c *Cluster) addOneMember() error {
+	c.status.AppendScalingUpCondition()
```
hongchaodeng: We also have self hosted path.

xiang90: what do you mean? can you provide more details?

hongchaodeng: If we call AppendScalingUpCondition() in addOneMember(), we also need to do the same in addOneSelfHostedMember().

xiang90: right. done.

hongchaodeng:
@etcd-bot retest this please

xiang90:

hongchaodeng:
lgtm

PR #538, opened 2016-12-27
cluster: report paused info

xiang90:


pkg/spec/cluster.go
```diff
@@ -20,7 +20,6 @@ import (
 	"time"
 
 	"k8s.io/kubernetes/pkg/api"
-	"k8s.io/kubernetes/pkg/api/meta/metatypes"
```
hongchaodeng: don't remove this?

xiang90:
@hongchaodeng PTAL. tests fail due to known flaky.

pkg/spec/cluster.go
```diff
@@ -163,6 +166,14 @@ func (cs *ClusterStatus) SetPhaseFailed() {
 	cs.Phase = ClusterPhaseFailed
 }
 
+func (cs *ClusterStatus) PausedControl() {
+	cs.ControlPaused = true
+}
+
+func (cs *ClusterStatus) Control() {
```
hongchaodeng: ResumeControl()?

pkg/spec/cluster.go
```diff
@@ -163,6 +166,14 @@ func (cs *ClusterStatus) SetPhaseFailed() {
 	cs.Phase = ClusterPhaseFailed
 }
 
+func (cs *ClusterStatus) PausedControl() {
```
hongchaodeng: PauseControl()?

xiang90: sure.

pkg/spec/cluster.go
xiang90: if you look at the code, we actually update the status in every round. we do not really care about the previous state of `paused`. Resume assumes the previous state is paused. So it is not suitable.

hongchaodeng: ok

hongchaodeng:
lgtm

PR #556, opened 2016-12-30
k8sutil: add checkpoint annotation for selfhosted etcd

xiang90:
/cc @aaronlevy

### DIFF
--- a/pkg/util/k8sutil/self_hosted.go
+++ b/pkg/util/k8sutil/self_hosted.go
@@ -24,6 +24,10 @@ import (
 	"k8s.io/kubernetes/pkg/api/meta/metatypes"
 )
 
+const (
+	shouldCheckpointAnnotation = "checkpointer.alpha.coreos.com/checkpoint" // = "true"
+)
+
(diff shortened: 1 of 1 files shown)


hongchaodeng:
Failed test
```
--- FAIL: TestEtcdUpgrade (50.91s)
    upgrade_test.go:46: failed to create 3 members etcd cluster: still failing after 6 retries
FAIL
```

It seems like there is a pod always pending, i.e. never starts running.
```
skip reconciliation: running ([test-etcd-4r27n-0000 test-etcd-4r27n-0001]), pending ([test-etcd-4r27n-0002])
```
Seems more like k8s issue.

hongchaodeng:
@etcd-bot retest this please

hongchaodeng:
defer to @aaronlevy for last check

hongchaodeng:
feel free to merge this

aaronlevy:
One question: what about pods which are created when I do something like expand from 1node to 3? Will it use this same spec or is this self-hosted pod special?

Otherwise, code lgtm

xiang90:
> One question: what about pods which are created when I do something like expand from 1node to 3? Will it use this same spec or is this self-hosted pod special?

self hosted is a cluster, not just one pod or container. so it will apply to all.

aaronlevy:
sgtm

PR #576, opened 2017-01-03
spec: allow image to be set in ClusterSpec

lclarkmichalek:
This makes it possible for the user to run a custom build of etcd. It does not interact with version management in any way, meaning that users will need to modify the Version field to roll out a new image.

### DIFF
--- a/pkg/cluster/upgrade.go
+++ b/pkg/cluster/upgrade.go
@@ -29,7 +29,7 @@ func (c *Cluster) upgradeOneMember(m *etcdutil.Member) error {
 		return fmt.Errorf("fail to get pod (%s): %v", m.Name, err)
 	}
 	c.logger.Infof("upgrading the etcd member %v from %s to %s", m.Name, k8sutil.GetEtcdVersion(pod), c.cluster.Spec.Version)
-	pod.Spec.Containers[0].Image = k8sutil.MakeEtcdImage(c.cluster.Spec.Version)
+	pod.Spec.Containers[0].Image = k8sutil.MakeEtcdImage(c.cluster.Spec)
 	k8sutil.SetEtcdVersion(pod, c.cluster.Spec.Version)
 	_, err = c.config.KubeCli.Pods(c.cluster.Namespace).Update(pod)
 	if err != nil {
--- a/pkg/spec/cluster.go
+++ b/pkg/spec/cluster.go
@@ -60,6 +60,9 @@ type ClusterSpec struct {
 	// equal to the expected version.
 	Version string `json:"version"`
 
+	// Image is the name of the docker image used for pods created by the etcd controller.
+	Image string `json:"image,omitempty"`
+
 	// Backup defines the policy to backup data of etcd cluster if not nil.
 	// If backup policy is set but restore policy not, and if a previous backup exists,
 	// this cluster would face conflict and fail to start.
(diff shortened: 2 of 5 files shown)


hongchaodeng:
@lclarkmichalek 
What's your use case? Why do you need a custom image?

lclarkmichalek:
I've a couple of patches for etcd I've been developing, and would like to be able to use the etcd-operator to run my patched etcd. This'd make it easier for me to test out my changes in my kube based environments, which use the etcd-operator to run etcd currently.

hongchaodeng:
@lclarkmichalek 
We provide a way for advanced user to try and build etcd operator to run their custom images:
https://github.com/coreos/etcd-operator/blob/master/hack/release/README.md
Actually we run our test infrastructure against both quay and gcr.

xiang90:
We cannot accept this patch at the moment. We are not ready to allow users to run arbitrary images. Upgrading, backup are all version aware and depends on our image naming. Our current focus is to stabilize the core functionality of the operator. And we cannot really provide a lot of flexibility now.

As @hongchaodeng mentioned, you can build your own operator to try out new ideas.

We will reevaluate this after we feel the operator itself is stable.

euank:
@xiang90 perhaps the API for that could be 'experimental', e.g. by renaming the field to `ExperimentalCustomImage` to make it clear that it shouldn't generally be used and isn't necessarily going to exist forever. Would that alleviate your concern?

I'd expect people that run their own private registry (e.g. a hosted quay) would also want something like this.

xiang90:
> I'd expect people that run their own private registry (e.g. a hosted quay) would also want something like this.

we probably can config the operator to talk to different registry, but still verify the image and relies on naming. would that be a better first step?

lclarkmichalek:
I can give an exact use case, if that's useful?

I'm building https://www.etcd-stg.genevieve.io/ (probably not working, 100% for fun), as a playground to test etcd features useful when you have hostile tenants on the cluster (resource quotas, chroots, that kind of thing). As this is all running on kube, testing those features is much easier if I can quickly spin up a cluster using the etcd-operator. Before upstreaming my patches, I'd like to run them on my stg/prd clusters to test them out against a broader audience.

Currently, I'm publishing an image named by git commit, but I could easily shift that to a version number based system. What if
```yaml
Version: v3.0.1
Image: foo.com/etcd:v3.0.1
```
were changed to
```yaml
Version: v3.0.1
ImageRoot: foo.com/etcd
```
and I add a check before starting nodes to ensure that etcd binary in that image is the same version as the version specified in the `Version` field?

xiang90:
One idea I had is to separate them into

```go
Version
BaseImage
```

Similar to what you just proposed.

But again, I do not think we are going to do it right away (accepting your pr immediately).

hongchaodeng:
@lclarkmichalek 
It sounds like you are building some experimental tools or services on top of etcd operator. That's welcome. My point is that for doing such stuff you need to build your custom image and maintain it on your own.

We can't accept this kind of change right now because it breaks our codebase. For example, what does it mean to upgrade from "foo:v3.0" to "foo:v4.0"? We need some careful thoughts and design considerations on here. Please consider creating an issue first and discuss your thoughts.

lclarkmichalek:
Cool, thanks, hope to see this in a future version then

Issue #627, opened 2017-01-07
e2e flake: recovery_test.go:138: pods "test-etcd-1nids-0000" not found


hongchaodeng:
https://jenkins-etcd.prod.coreos.systems/job/etcd-operator/1138/console
```
--- FAIL: TestFailureRecovery (0.00s)
    --- FAIL: TestFailureRecovery/failure_recovery (0.00s)
        --- FAIL: TestFailureRecovery/failure_recovery/3_members_(all)_down (46.41s)
        	util.go:199: created etcd cluster: test-etcd-1nids
        	util.go:111: waiting size (3), etcd pods: []
        	util.go:111: waiting size (3), etcd pods: []
        	util.go:111: waiting size (3), etcd pods: [test-etcd-1nids-0000 test-etcd-1nids-0001]
        	util.go:111: waiting size (3), etcd pods: [test-etcd-1nids-0000 test-etcd-1nids-0001]
        	util.go:111: waiting size (3), etcd pods: [test-etcd-1nids-0000 test-etcd-1nids-0001 test-etcd-1nids-0002]
        	recovery_test.go:138: pods "test-etcd-1nids-0000" not found
        	util.go:214: pod (test-etcd-1nids-0003): status (Pending), cmd ([/bin/sh -c sleep 5; /usr/local/bin/etcd --data-dir=/var/etcd/data --name=test-etcd-1nids-0003 --initial-advertise-peer-urls=http://test-etcd-1nids-0003:2380 --listen-peer-urls=http://0.0.0.0:2380 --listen-client-urls=http://0.0.0.0:2379 --advertise-client-urls=http://test-etcd-1nids-0003:2379 --initial-cluster=test-etcd-1nids-0003=http://test-etcd-1nids-0003:2380 --initial-cluster-state=new --initial-cluster-token=e33c7455-aadd-4428-9461-adbc08a2f87f])
```

hongchaodeng:
I looked into the logs again.
It seems that operator might have removed it due to disaster recovery.

hongchaodeng:
fixed via #636

PR #631, opened 2017-01-08
cluster: fix failure handling code logic

xiang90:
The previous code logic is unclear and error-prone since the if
condidtion describes the intended behavior not the state condition.

Changing the condition to state condition, and handling the condition
correctly.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -205,17 +205,18 @@ func (c *Cluster) send(ev *clusterEvent) {
 }
 
 func (c *Cluster) run(stopC <-chan struct{}, wg *sync.WaitGroup) {
-	needDeleteCluster := true
+	clusterFailed := false
 
 	defer func() {
-		if needDeleteCluster {
-			c.logger.Infof("deleting cluster")
+		if clusterFailed {
+			c.reportFailedStatus()
+
+			c.logger.Infof("deleting the failed cluster")
 			c.delete()
 		}
+
 		close(c.stopCh)
 		wg.Done()
-
-		c.reportFailedStatus()
 	}()
 
 	c.status.SetPhase(spec.ClusterPhaseRunning)
@@ -224,7 +225,6 @@ func (c *Cluster) run(stopC <-chan struct{}, wg *sync.WaitGroup) {
 	for {
 		select {
 		case <-stopC:
-			needDeleteCluster = false
 			return
 		case event := <-c.eventCh:
 			switch event.typ {
@@ -235,9 +235,13 @@ func (c *Cluster) run(stopC <-chan struct{}, wg *sync.WaitGroup) {
 				// TODO: we can't handle another upgrade while an upgrade is in progress
 				c.logger.Infof("spec update: from: %v to: %v", c.cluster.Spec, event.cluster.Spec)
 				c.cluster = event.cluster
+
 			case eventDeleteCluster:
+				c.logger.Infof("cluster is deleted by the user")
+				clusterFailed = true
 				return
 			}
+
 		case <-time.After(reconcileInterval):
 			if c.cluster.Spec.Paused {
 				c.status.PauseControl()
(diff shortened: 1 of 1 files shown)


pkg/cluster/cluster.go
```diff
@@ -235,9 +235,13 @@ func (c *Cluster) run(stopC <-chan struct{}, wg *sync.WaitGroup) {
 				// TODO: we can't handle another upgrade while an upgrade is in progress
 				c.logger.Infof("spec update: from: %v to: %v", c.cluster.Spec, event.cluster.Spec)
 				c.cluster = event.cluster
+
 			case eventDeleteCluster:
+				c.status.SetReason("cluster is deleted by the user")
```
hongchaodeng: TPR is already deleted. Setting status isn't helpful.

pkg/cluster/cluster.go
```diff
@@ -235,9 +235,13 @@ func (c *Cluster) run(stopC <-chan struct{}, wg *sync.WaitGroup) {
...
+				clusterFailed = true
```
hongchaodeng: We probably need another state called "clusterDeleted" and wouldn't need to report failed on that.

pkg/cluster/cluster.go
xiang90: this is just to make code clear. probably i should change this to a logging message?

pkg/cluster/cluster.go
xiang90: we can do that in another pr.

pkg/cluster/cluster.go
hongchaodeng: There is another logging message
```
"deleting the failed cluster"
```

We'd better have one message for either "failed deleting" and "normal deleting". But definitely not two together.

xiang90:
@hongchaodeng fixed. PTAL.

hongchaodeng:
> We'd better have one message for either "failed deleting" and "normal deleting". But definitely not two together.

I probably want this eventually. But I'm fine with this PR. So LGTM here

pkg/cluster/cluster.go
```diff
@@ -205,17 +205,18 @@ func (c *Cluster) send(ev *clusterEvent) {
...
 
 func (c *Cluster) run(stopC <-chan struct{}, wg *sync.WaitGroup) {
-	needDeleteCluster := true
+	clusterFailed := false
 
 	defer func() {
-		if needDeleteCluster {
-			c.logger.Infof("deleting cluster")
+		if clusterFailed {
+			c.reportFailedStatus()
```
hongchaodeng: Make this async?

This will block controller unboundedly

xiang90: why cluster will block the controller? i do not think it ever should. if cluster blocks controller, then there is something wrong somewhere else i guess.

hongchaodeng: reportFailedStatus() is unbounded.

On restore, controller will wait all cluster run() exit, using the passed "wg".

xiang90: > On restore, controller will wait all cluster run() exit, using the passed "wg".

this is wrong, and we should fix it. controller should never wait for a specific cluster.

xiang90: also i am not super concerned about this case right now. if reporting fails, then the cluster cannot connect to API server. Controller cannot do anything meaningful either. But we should fix the root problem: a cluster event should never blocks the controller.

hongchaodeng: > a cluster event should never blocks the controller.

This is not only cluster event, but also controller restore case. Waiting all exits only happens on controller restore.

> if reporting fails, then the cluster cannot connect to API server.

Add a comment?

hongchaodeng:
@etcd-bot retest this please

hongchaodeng:
@etcd-bot retest this please

https://github.com/coreos/etcd-operator/issues/627

PR #632, opened 2017-01-08
*: simplify status recovery logic

xiang90:
1. do not recover any failed cluster

2. always recover status from the existing etcd cluster TPR item if possible

3. make status the source of truth after the cluster struct is materialized

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -65,6 +65,7 @@ type Cluster struct {
 	cluster *spec.EtcdCluster
 
 	// in memory state of the cluster
+	// status is the source of truth after Cluster struct is materialized.
 	status        *spec.ClusterStatus
 	memberCounter int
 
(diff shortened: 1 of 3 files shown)


pkg/cluster/cluster.go
```diff
@@ -118,20 +121,17 @@ func (c *Cluster) setup() error {
...
 		shouldCreateCluster = true
 	case spec.ClusterPhaseCreating:
 		return errors.New("cluster failed to be created")
 	case spec.ClusterPhaseRunning:
 		shouldCreateCluster = false
-	case spec.ClusterPhaseFailed:
-		return errors.New("cluster already failed")
+
+	default:
+		return fmt.Errorf("unexpected cluster state: %s", c.status.Phase)
```
hongchaodeng: state -> phase?

hongchaodeng:
lgtm

PR #642, opened 2017-01-11
remove dead pod in one reconcile round

icyxieex:
* the case: we have 5 members at first, 2 die and user set `sepc.size = 3`

### DIFF
--- a/pkg/cluster/reconcile.go
+++ b/pkg/cluster/reconcile.go
@@ -92,13 +92,17 @@ func (c *Cluster) reconcileMembers(running etcdutil.MemberSet) error {
 		c.logger.Infof("Disaster recovery")
 		return c.disasterRecovery(L)
 	}
-
-	c.logger.Infof("replacing one dead member with a new member")
-
-	if err := c.removeDeadMember(c.members.Diff(L).PickOne()); err != nil {
-		return err
+	
+	deadMembers := c.members.Diff(L)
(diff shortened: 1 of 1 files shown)


hongchaodeng:
Why do you want to remove all at once?

hongchaodeng:
Any use case or concern? Please elaborate.

hongchaodeng:
@icyxieex 
Any update here? Or should we close it?

icyxieex:
the case: we have 5 members at first, 2 die , and then the user set sepc.size = 3
firstly, reconcileMembers() will remove one deadMember,we have 4 members> sept.size,
next one random member will be removed,with the dead member may not be treated。

xiang90:
> we have 4 members> sept.size,
next one random member will be removed,with the dead member may not be treated。

We should prioritize the removal in this case. we remove dead member first. Sounds good?

icyxieex:
@hongchaodeng  I think we should remove all dead members in this case

xiang90:
@icyxieex Each reconcile should only do one action unless there is a strong need. Your use case can be solved by what I proposed.

icyxieex:
@xiang90 OK, I close it

xiang90:
@icyxieex Can you open an issue to track the use case you mentioned above though? It would be great if you can solve it the way I proposed.

icyxieex:
I open an issue #648 , is it right? can you assign it to me @xiang90

PR #643, opened 2017-01-11
client: add initial implementation

xiang90:
We need to update everything to use client-go or this client library wont work.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -2,21 +2,168 @@ package client
 
 import (
 	"context"
+	"net/http"
 
 	"github.com/coreos/etcd-operator/pkg/spec"
+	"github.com/coreos/etcd-operator/pkg/util/constants"
+
+	"k8s.io/client-go/1.5/pkg/api"
+	"k8s.io/client-go/1.5/pkg/api/unversioned"
+	"k8s.io/client-go/1.5/pkg/api/v1"
+	"k8s.io/client-go/1.5/pkg/runtime"
+	"k8s.io/client-go/1.5/pkg/runtime/serializer"
+	"k8s.io/client-go/1.5/rest"
 )
 
 // Operator operators etcd clusters atop Kubernetes.
 type Operator interface {
-	// Create creates an etcd cluster with the given spec.
+	// Create creates an etcd cluster.
 	Create(ctx context.Context, name string, spec *spec.ClusterSpec) error
 
-	// Delete deletes the etcd cluster with given name
+	// Delete deletes the etcd cluster.
 	Delete(ctx context.Context, name string) error
 
 	// Update updates the etcd cluster with the given spec.
 	Update(ctx context.Context, name string, spec *spec.ClusterSpec) error
 
-	// Status gets the status of the etcd cluster with the given name.
-	Status(ctx context.Context, name string) (error, *spec.ClusterStatus)
+	// Get gets the etcd cluster information.
+	Get(ctx context.Context, name string) (*spec.EtcdCluster, error)
+
+	// List lists all etcd clusters.
+	List(ctx context.Context) (*spec.EtcdClusterList, error)
+}
+
+var (
+	groupversion = unversioned.GroupVersion{
+		Group:   "coreos.com",
+		Version: "v1",
+	}
+)
+
+func init() {
+	schemeBuilder := runtime.NewSchemeBuilder(
+		func(scheme *runtime.Scheme) error {
+			scheme.AddKnownTypes(
+				groupversion,
+				&spec.EtcdCluster{},
+				&spec.EtcdClusterList{},
+				&v1.ListOptions{},
+				&v1.DeleteOptions{},
+			)
+			return nil
+		})
+	schemeBuilder.AddToScheme(api.Scheme)
+}
+
+type operator struct {
+	tprClient *rest.RESTClient
+	tprName   string
+	ns        string
+}
+
+func NewOperator(namespace string) (Operator, error) {
+	cfg, err := rest.InClusterConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	configureClient(cfg)
+
+	tprclient, err := rest.RESTClientFor(cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	return &operator{
+		tprClient: tprclient,
+		tprName:   constants.TPRName,
+		ns:        namespace,
+	}, nil
+
+}
+
+func (o *operator) Create(ctx context.Context, name string, cspec *spec.ClusterSpec) error {
+	cluster := &spec.EtcdCluster{
+		ObjectMeta: v1.ObjectMeta{
+			Name: name,
+		},
+		Spec: cspec,
+	}
+
+	err := o.tprClient.Post().
+		Resource(o.tprName).
+		Namespace(o.ns).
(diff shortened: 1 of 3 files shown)


client/client.go
```diff
@@ -2,21 +2,141 @@ package client
...
 	Update(ctx context.Context, name string, spec *spec.ClusterSpec) error
 
-	// Status gets the status of the etcd cluster with the given name.
-	Status(ctx context.Context, name string) (error, *spec.ClusterStatus)
+	// Get gets the etcd cluster information.
+	Get(ctx context.Context, name string) (*spec.EtcdCluster, error)
+}
+
+type operator struct {
+	tprClient *rest.RESTClient
```
hongchaodeng: tprClient -> restcli

client/client.go
```diff
@@ -2,21 +2,141 @@ package client
...
+	tprName   string
```
hongchaodeng: Isn't tpr name const?

client/client.go
```diff
@@ -2,21 +2,141 @@ package client
...
+		tprClient: tprclient,
+		tprName:   tpr,
+		ns:        namespace,
+	}, nil
+
+}
+
+func (o *operator) Create(ctx context.Context, name string, cspec *spec.ClusterSpec) error {
+	cluster := &spec.EtcdCluster{
+		ObjectMeta: api.ObjectMeta{
```
hongchaodeng: We will use version types only, e.g. "v1. ObjectMeta"

client/client.go
xiang90: `tprClient` is what suggested in client-go example. I would rather stick to it.

client/client.go
hongchaodeng: This also applies to all others using "api."

xiang90: any reason they use api here: https://github.com/kubernetes/client-go/blob/master/examples/third-party-resources/main.go#L142-L155?

client/client.go
xiang90: i changed it to use the const. but we limit users to use the fixed tpr name for no good reason rather than initial simplicity. and we will probably make it configurable soon. so users can have multiple operators under the same namespace.

client/client.go
hongchaodeng: So there are two kinds of APIs here: one is within k8s and one is our TPR.

(My guess) The code you mentioned seems to be registering the TPR type. For TPR type, we would api.ListOptions/DeleteOptions . I'm not sure what's api.Codecs . We should ask the question to client-go.

xiang90: @hongchaodeng If I understand this correctly, /api maps to k8s internal types. v1 will eventually get converted to api types when the requests reach k8s api server. so we should only use v1 externally. 

so i am not sure if it is correct to change ALL api to v1 as you suggested. since the code i mentioned is provided by client-go. i am not sure if i should treat client type registration external or internal. 

i keep what client-go suggests for now.

xiang90:
@hongchaodeng All fixed. With the new vendor client-go, everything works now.

pkg/controller/controller.go
```diff
@@ -37,7 +37,7 @@ import (
 )
 
 const (
-	tprName = "etcd-cluster.coreos.com"
+	TPRName = "etcd-cluster.coreos.com"
```
hongchaodeng: Why change this?

xiang90: Client needs to know the TPR name to operator. see your comment here https://github.com/coreos/etcd-operator/pull/643#discussion_r95631853

hongchaodeng: It seems that you also define `tprName` in client package.. I'm confused.

xiang90: I should remove that one

xiang90: @hongchaodeng removed.

client/client.go
```diff
@@ -2,21 +2,167 @@ package client
...
 
 	// Update updates the etcd cluster with the given spec.
 	Update(ctx context.Context, name string, spec *spec.ClusterSpec) error
 
-	// Status gets the status of the etcd cluster with the given name.
-	Status(ctx context.Context, name string) (error, *spec.ClusterStatus)
+	// Get gets the etcd cluster information.
+	Get(ctx context.Context, name string) (*spec.EtcdCluster, error)
+
+	// List lists all the etcd clusters.
```
hongchaodeng: all ~~the~~ etcd

client/client.go
```diff
@@ -2,21 +2,164 @@ package client
...
+		Do().Into(clusters)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return clusters, nil
+}
+
+func configureClient(config *rest.Config) {
```
hongchaodeng: We should call this func only once.
Let's call it in init()?

xiang90: i do not understand this. this func also modify config. it should be called when there is a new rest config.

hongchaodeng: Wait.. I rethink about this func. I think part of it should be called to mutate config. The scheme registering part should be called only once.

xiang90: will it panic if we call it multiple times? if not, i would rather keep as it is (stick with the tpr example in client/go). if it panics, we need to fix both the trp example and this.

hongchaodeng: See https://github.com/kubernetes/client-go/issues/59

hongchaodeng: This is using global var api.Codecs, api.Scheme.
Let's create new instance and mutate it instead as:
https://github.com/atlassian/smith/blob/25c5183c6718806f38761bc5f1df3ab3d6a8a22b/pkg/resources/client.go#L67-L76

xiang90: ok. will fix.

client/client.go
xiang90: will fix.

client/client.go
xiang90: i move the scheme registration to init. there should not be any race in this way. i do not want to create an instance since it is tedious to maintain it.

pkg/controller/controller.go
hongchaodeng: Can we move this to pkg/util/constants/ ? Otherwise client will import the whole controller package.

client/client.go
hongchaodeng: Why is it tedious? It's not much difference from this code.

xiang90: @hongchaodeng https://github.com/atlassian/smith/blob/36fadf8337c1e0cddfb35a5d398259d46eb0d706/pkg/app/app.go#L34

you need to start passing the scheme around for codec.

hongchaodeng: I'm fine with this anyway.

client/client.go
```diff
@@ -2,21 +2,168 @@ package client
...
+	// Get gets the etcd cluster information.
+	Get(ctx context.Context, name string) (*spec.EtcdCluster, error)
+
+	// List lists all etcd clusters.
+	List(ctx context.Context) (*spec.EtcdClusterList, error)
+}
+
+var (
+	groupversion = unversioned.GroupVersion{
+		Group:   "k8s.io",
```
hongchaodeng: it's "coreos.com"

xiang90:
@hongchaodeng All fixed.

hongchaodeng:
lgtm

PR #644, opened 2017-01-12
cluster: improve selfhosted cluster reliability

xiang90:
/cc @hongchaodeng

### DIFF
--- a/pkg/cluster/member.go
+++ b/pkg/cluster/member.go
@@ -15,6 +15,8 @@
 package cluster
 
 import (
+	"fmt"
+
 	"github.com/coreos/etcd-operator/pkg/spec"
 	"github.com/coreos/etcd-operator/pkg/util/etcdutil"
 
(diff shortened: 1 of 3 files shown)


pkg/cluster/member.go
```diff
@@ -30,6 +32,10 @@ func (c *Cluster) updateMembers(known etcdutil.MemberSet) error {
 	for _, m := range resp.Members {
 		var name string
 		if c.cluster.Spec.SelfHosted != nil {
+			if m.PeerURLs[0] == c.cluster.Spec.SelfHosted.BootMemberClientEndpoint {
+				return fmt.Errorf("skipping update members for self hosted cluster: waiting for the boot member to be removed...")
```
hongchaodeng: "boot member (%s)"

hongchaodeng:
lgtm

pkg/cluster/self_hosted.go
```diff
@@ -114,14 +114,16 @@ func (c *Cluster) migrateBootMember() error {
 		return err
 	}
 
-	delay := 20 * time.Second
-	c.logger.Infof("wait %v before removing the boot member", delay)
-	time.Sleep(delay)
+	go func() {
+		delay := 60 * time.Second
```
xiang90: @hongchaodeng I am not really sure if 60 is necessary. But let's keep it this way for safety? We can shorten this in the future if bootkube can work with shorter timeout?

pkg/cluster/member.go
```diff
@@ -31,10 +33,16 @@ func (c *Cluster) updateMembers(known etcdutil.MemberSet) error {
 		var name string
 		if c.cluster.Spec.SelfHosted != nil {
 			name = m.Name
-			if len(name) == 0 {
+			if len(name) == 0 || len(m.ClientURLs) == 0 {
```
hongchaodeng: Why do we need to check ClientURL?

hongchaodeng: Oh I see. Boot member client endpoint.

xiang90: we are going to access m.ClientURL[0] later. we should check if the server finished publishing its client url. or we might panic.

pkg/cluster/self_hosted.go
hongchaodeng: Add a TODO?

hongchaodeng:
lgtm

Issue #649, opened 2017-01-14
glide is broken


hongchaodeng:
The glide in jenkins is actually now using some weird cache.

After deleting the cache:
```
$ rm -rf ~/.glide/
$ glide install --strip-vendor --skip-test
[ERROR]	Error scanning k8s.io/client-go/1.4/rest: open /Users/hdeng/.glide/cache/src/https-k8s.io-client-go/1.4/rest: no such file or directory
[ERROR]	This error means the referenced package was not found.
[ERROR]	Missing file or directory errors usually occur when multiple packages
[ERROR]	share a common dependency and the first reference encountered by the scanner
[ERROR]	sets the version to one that does not contain a subpackage needed required
[ERROR]	by another package that uses the shared dependency. Try setting a
[ERROR]	version in your glide.yaml that works for all packages that share this
[ERROR]	dependency.
```

Even if I manually added back client-go v1.4.0, it would still fail:
```
04:27:45 package github.com/coreos/etcd-operator/test/e2e (test)
04:27:45 	imports github.com/coreos/etcd/embed
04:27:45 	imports github.com/coreos/etcd/etcdserver
04:27:45 	imports github.com/coreos/etcd/alarm
04:27:45 	imports github.com/coreos/etcd/mvcc/backend
04:27:45 	imports github.com/boltdb/bolt: cannot find package "github.com/boltdb/bolt" in any of:
04:27:45 	/home/jenkins/workspace/build-etcd-operator-flaketest-734/gopath/src/github.com/coreos/etcd-operator/vendor/github.com/boltdb/bolt (vendor tree)
04:27:45 	/usr/local/go/src/github.com/boltdb/bolt (from $GOROOT)
04:27:45 	/home/jenkins/workspace/build-etcd-operator-flaketest-734/gopath/src/github.com/boltdb/bolt (from $GOPATH)
```

Even previous command will fail:
```
$ glide install
[ERROR]	Error scanning k8s.io/client-go/1.4/rest: open /Users/hdeng/.glide/cache/src/https-k8s.io-client-go/1.4/rest: no such file or directory
[ERROR]	This error means the referenced package was not found.
[ERROR]	Missing file or directory errors usually occur when multiple packages
[ERROR]	share a common dependency and the first reference encountered by the scanner
[ERROR]	sets the version to one that does not contain a subpackage needed required
[ERROR]	by another package that uses the shared dependency. Try setting a
[ERROR]	version in your glide.yaml that works for all packages that share this
[...]

PR #650, opened 2017-01-14
Revert "*: add glide install options"

hongchaodeng:
Reverts coreos/etcd-operator#646

The build becomes flaky today. I tried that reverting this will make build stable again. Let's not break the master first.

As mentioned in https://github.com/coreos/etcd-operator/issues/649, it's still broken.

### DIFF
--- a/hack/jenkins
+++ b/hack/jenkins
@@ -27,7 +27,7 @@ cleanup() {
 
 trap cleanup EXIT
 
-glide install --strip-vendor --skip-test 1>/dev/null
+glide install 1>/dev/null
 
 GIT_VERSION=$(git rev-parse HEAD)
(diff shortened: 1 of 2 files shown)


xiang90:
I do not understand this. why glide options can affect our build? Can you confirm this commit is the bad one?

hongchaodeng:
@xiang90 
Yes.
With this commit https://github.com/coreos/etcd-operator/commit/ca626ece46a52605c5de5c40f4d38dfb33d4e80f, CI test turned green: https://jenkins-etcd.prod.coreos.systems/job/etcd-operator-flaketest/

xiang90:
LGTM. But we need to figure out why glide does not work well.

PR #654, opened 2017-01-16
make sure deadMembers will be removed before resize()

### LINKED ISSUE #648: remove dead members firstly
* use case: we have 5 members at first, 2 die , and then the user sets sepc.size = 3
reconcileMembers() will remove one dead member, we have 4 members> sept.size,
next one random member will be removed with the dead member may not be treated.
* as @xiang90 proposed, we should remove the dead member at each reconcile

icyxieex:
fix #648

### DIFF
--- a/pkg/cluster/reconcile.go
+++ b/pkg/cluster/reconcile.go
@@ -93,13 +93,9 @@ func (c *Cluster) reconcileMembers(running etcdutil.MemberSet) error {
 		return c.disasterRecovery(L)
 	}
 
-	c.logger.Infof("replacing one dead member with a new member")
-
-	if err := c.removeDeadMember(c.members.Diff(L).PickOne()); err != nil {
-		return err
-	}
-
-	return c.resize()
+	c.logger.Infof("removing one dead member")
+	// remove dead members that doesn't have any running pods before doing resizing.
+	return c.removeDeadMember(c.members.Diff(L).PickOne())
 }
 
 func (c *Cluster) resize() error {


hongchaodeng:
Can you changed the commit message to be something more descriptive?

hongchaodeng:
The title "remove dead members firstly" is confusing. What does it mean by "firstly"? Can you change that to be more descriptive?

pkg/cluster/reconcile.go
```diff
@@ -95,11 +95,17 @@ func (c *Cluster) reconcileMembers(running etcdutil.MemberSet) error {
 
 	c.logger.Infof("replacing one dead member with a new member")
```
hongchaodeng: Can you change the logging?
It seems not true with this change

hongchaodeng:
I rethink about the whole process. I don't understand how could we have the problem?

Let's say I have five members, A0 ... A4, and two died, A0 and A1. If we also set "spec.size = 3", then it will at each reconcile round
- remove A0, check (members.Size() = 4) > (spec.size = 3)
- remove A1, check  (members.Size() = 3) == (spec.size = 3)

It won't have the problem as you mentioned, right?

icyxieex:
@hongchaodeng 
remove A0, then call resize(),and then removeOneMember() will delete one random member.
check (members.Size() =3) == (spec.size = 3)

now  A1 won't be treated

hongchaodeng:
> after remove A0, it will call resize(),and then removeOneMember(), right?

Yes. Now members is A1...A4 .

> removeOneMember() maybe remove another living member, right?

It won't call removeOneMember(). Instead, it will call addOneMember().

hongchaodeng:
Sorry. Now I understand, it seems like it will have "member.size > spec.size"

pkg/cluster/reconcile.go
```diff
@@ -93,13 +93,18 @@ func (c *Cluster) reconcileMembers(running etcdutil.MemberSet) error {
 		return c.disasterRecovery(L)
 	}
 
-	c.logger.Infof("replacing one dead member with a new member")
+	deadMembers := c.members.Diff(L)
+	c.logger.Infof("remove one dead member")
```
hongchaodeng: removing

pkg/cluster/reconcile.go
```diff
@@ -93,13 +93,18 @@ func (c *Cluster) reconcileMembers(running etcdutil.MemberSet) error {
...
 
-	if err := c.removeDeadMember(c.members.Diff(L).PickOne()); err != nil {
+	if err := c.removeDeadMember(deadMembers.PickOne()); err != nil {
 		return err
 	}
 
-	return c.resize()
+	if deadMembers.Size() == 0 {
```
hongchaodeng: We don't need this. Just return nil here.

pkg/cluster/reconcile.go
hongchaodeng: Can you add a comment here:
```
// We will remove dead members that doesn't have any running pods before doing resizing.
```

pkg/cluster/reconcile.go
```diff
@@ -93,13 +93,18 @@ func (c *Cluster) reconcileMembers(running etcdutil.MemberSet) error {
 		return c.disasterRecovery(L)
 	}
 
-	c.logger.Infof("replacing one dead member with a new member")
+	deadMembers := c.members.Diff(L)
+	c.logger.Infof("remove one dead member")
 
-	if err := c.removeDeadMember(c.members.Diff(L).PickOne()); err != nil {
+	if err := c.removeDeadMember(deadMembers.PickOne()); err != nil {
```
hongchaodeng: Just keep original line:
```
c.members.Diff(L).PickOne()
```

Since we don't need to reuse `deadMembers`.

pkg/cluster/reconcile.go
```diff
@@ -93,10 +93,10 @@ func (c *Cluster) reconcileMembers(running etcdutil.MemberSet) error {
 		return c.disasterRecovery(L)
 	}
 
-	c.logger.Infof("replacing one dead member with a new member")
-
-	if err := c.removeDeadMember(c.members.Diff(L).PickOne()); err != nil {
-		return err
+	c.logger.Infof("removing one dead member")
+	// remove dead members that doesn't have any running pods before doing resizing.
+	if c.members.Diff(L).Size() > 0 {
```
hongchaodeng: I mean the following:
```
	return c.removeDeadMember(c.members.Diff(L).PickOne())
```

Because if we come to this point, it means that we must have dead members and must have majority alive.

icyxieex:
@hongchaodeng 
Was it right now？

hongchaodeng:
@icyxieex 
Please see my above comment. Just one minor nit.
One more item: please squash your commits into one. After that I will lgtm.

hongchaodeng:
lgtm

hongchaodeng:
Thanks @icyxieex

PR #681, opened 2017-01-27
backup: add v1 prefix to backup http endpoint

xiang90:
/cc @hongchaodeng

### DIFF
--- a/pkg/util/http.go
+++ b/pkg/util/http.go
@@ -22,7 +22,7 @@ func MakeBackupURL(host, version string) *url.URL {
 	u := &url.URL{
 		Scheme: "http",
 		Host:   host,
-		Path:   "/backup",
+		Path:   "/v1/backup",
 	}
 	q := u.Query()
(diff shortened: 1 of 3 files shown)


hongchaodeng:
lgtm

pkg/util/http.go
```diff
@@ -22,7 +22,7 @@ func MakeBackupURL(host, version string) *url.URL {
 	u := &url.URL{
 		Scheme: "http",
 		Host:   host,
-		Path:   "/backup",
+		Path:   "/v1/backup",
```
hongchaodeng: Why not use the const?

xiang90: there is the import cycle. i think you originally move this out to avoid import cycles. also /backup, /backupnow are still not using as const. it makes no sense to have backup http client related stuff in util actually.

we need a large code refactor to sort this out. see https://github.com/coreos/etcd-operator/issues/682

PR #689, opened 2017-01-27
s3: add version prefix

### LINKED ISSUE #673: versioning backup sidecar API
Clearly, it becomes a user facing API once we want to enable user driven disaster recovery when K8S itself is down.

xiang90:
Fix #673 

@hongchaodeng What is the plan for PV though? Can you fix that bit?

### DIFF
--- a/pkg/backup/s3/s3.go
+++ b/pkg/backup/s3/s3.go
@@ -20,10 +20,15 @@ import (
 	"path"
 	"strings"
 
+	"github.com/coreos/etcd-operator/pkg/backup/env"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
-	"github.com/coreos/etcd-operator/pkg/backup/env"
+)
+
+const (
+	v1 = "v1/"
 )
 
 // S3 represents AWS S3 service.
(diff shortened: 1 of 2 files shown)


hongchaodeng:
lgtm
Let's create another issue to discuss and handle PV case.

pkg/backup/s3/s3.go
```diff
@@ -61,7 +65,7 @@ func New(bucket, prefix string, option session.Options) (*S3, error) {
 func (s *S3) Put(key string, rs io.ReadSeeker) error {
 	_, err := s.client.PutObject(&s3.PutObjectInput{
 		Bucket: aws.String(s.bucket),
-		Key:    aws.String(s.prefix + key),
+		Key:    aws.String(path.Join(v1, s.prefix, key)),
```
hongchaodeng: another option is to prepend "v1" to "prefix" field. That should be easier. There is one special case in CopyPrefix() to handle.

pkg/backup/s3/s3.go
```diff
@@ -104,16 +109,16 @@ func (s *S3) List() ([]string, error) {
 func (s *S3) list(prefix string) ([]string, error) {
 	resp, err := s.client.ListObjects(&s3.ListObjectsInput{
 		Bucket: aws.String(s.bucket),
-		Prefix: aws.String(prefix),
+		Prefix: aws.String(path.Join(v1, prefix)),
```
hongchaodeng: This is wrong..
prefix ends with "/".
With path.Join(), the ending "/" is gone.

hongchaodeng: I sent out https://github.com/coreos/etcd-operator/pull/690 to help make this logic contained and simpler..

hongchaodeng:
lgtm

Issue #677, opened 2017-01-26
grace period should be non-zero for etcd pods


hongchaodeng:
ref: https://kubernetes.io/docs/user-guide/pods/#termination-of-pods

Currently, etcd operator deletes pods with zero grace period. This gives etcd server very little time to do graceful shutdown (e.g. completing current requests, releasing any remote resources like locks, connections).

We should change the grace period to non-zero like 5s as a better practice.

One problem I see is that on extreme cases we might fail to create another etcd pod with the same name. That's fine, we would retry again.

PR #716, opened 2017-02-01
cluster: remove pod with grace period

xiang90:
Give time to etcd to react on SIGTERM. This allows leader transfer to
happen before etcd gets SIGKILLed.

Fix https://github.com/coreos/etcd-operator/issues/677


hongchaodeng:
lgtm

Issue #355, opened 2016-11-11
create changelog


xiang90:
let's keep track on any user facing changes for each release.

PR #717, opened 2017-02-01
CHANGELOG: initial template

xiang90:
We need to keep the change log after beta release.

Fix https://github.com/coreos/etcd-operator/issues/355


hongchaodeng:
lgtm

Issue #624, opened 2017-01-07
Do not panic if too many unhandled events


hongchaodeng:
We rarely receive that many events for a cluster.
But this is definitely something we should fix:
```Go
func (c *Cluster) send(ev *clusterEvent) {
	select {
	case c.eventCh <- ev:
	case <-c.stopCh:
	default:
		panic("TODO: too many events queued...")
	}
}
```

PR #721, opened 2017-02-01
cluster: remove panic

xiang90:
After second thought, I believe we should simply block the
operator if one of the cluster cannot make progress.

It is either caused by a bug in client-go (unlimited request waiting)
or in our code (forget to set timeout?). If that happens, the best solution
(or simplest solution) would be restarted the operator. That should be
handled at the controller layer.

So in short, we should just let it blocking when it blocks.

Fix https://github.com/coreos/etcd-operator/issues/624

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -211,9 +211,11 @@ func (c *Cluster) Delete() {
 func (c *Cluster) send(ev *clusterEvent) {
 	select {
 	case c.eventCh <- ev:
+		l, ecap := len(c.eventCh), cap(c.eventCh)
+		if l > int(float64(ecap)*0.8) {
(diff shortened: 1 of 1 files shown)


pkg/cluster/cluster.go
```diff
@@ -212,8 +212,6 @@ func (c *Cluster) send(ev *clusterEvent) {
 	select {
 	case c.eventCh <- ev:
```
hongchaodeng: We should check the length of channel and have a watermark warning. This will help us have better insights into how operator runs.

xiang90: i can print out a warning here.

xiang90:
@hongchaodeng fixed. We will add a metrics for this soon.

xiang90:
@etcd-bot retest this please

xiang90:
can we get this merged

hongchaodeng:
lgtm

PR #723, opened 2017-02-01
test: enable race detector

### LINKED ISSUE #641: enable race detector in e2e test
we should enable race detector in our test soon as the operator becomes more concurrent.

xiang90:
Fix #641


hongchaodeng:
lgtm

Issue #615, opened 2017-01-06
GC and backup manager overlap-manages Backup RS


hongchaodeng:
See https://github.com/coreos/etcd-operator/issues/612

hongchaodeng:
@xiang90 
I think we should remove that responsibility from backup manager

xiang90:
Hmm... I guess it is fine after we added the notFound error checking.

hongchaodeng:
@xiang90 Just to clarify. This is not full GC, but lazy GC.
Basically, we call gc.CollectCluster() first, then bm.cleanup(). So they are overlapping each time we are deleting cluster

hongchaodeng:
Yep.
There are two motivations: 1. efficiency as you mentioned; 2. we aren't GC-ing backup service. I'm inclined to combining these two and also put backup service into GC.

xiang90:
We can do this after we fix our e2e?

hongchaodeng:
Sure. Just to create this issue to track

PR #726, opened 2017-02-02
cluster: let gc take care of removing k8s resources

xiang90:
Fix https://github.com/coreos/etcd-operator/issues/615

### DIFF
--- a/pkg/cluster/backup_manager.go
+++ b/pkg/cluster/backup_manager.go
@@ -132,16 +132,10 @@ func (bm *backupManager) createBackupService() error {
 }
 
 func (bm *backupManager) cleanup() error {
-	cl, c := bm.cluster, bm.config
-	err := k8sutil.DeleteBackupReplicaSetAndService(c.KubeCli, cl.Name, cl.Namespace)
+	// Only need to delete storage, Kubernetes related resources will be deleted by the GC.
+	err := bm.s.Delete()
 	if err != nil {
-		return fmt.Errorf("fail to delete backup ReplicaSet and Service: %v", err)
-	}
-	bm.logger.Infof("backup replica set and service deleted")
-
-	err = bm.s.Delete()
-	if err != nil {
-		return fmt.Errorf("fail to delete store: %v", err)
+		return fmt.Errorf("fail to delete backup storage: %v", err)
 	}
 	return nil
 }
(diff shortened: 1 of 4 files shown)


pkg/cluster/cluster.go
```diff
@@ -365,16 +365,15 @@ func (c *Cluster) Update(e *spec.EtcdCluster) {
...
-
 	if c.bm != nil {
-		err := c.bm.cleanup()
-		if err != nil {
+		if err := c.bm.cleanup(); err != nil {
 			c.logger.Errorf("cluster deletion: backup manager failed to cleanup: %v", err)
 		}
 	}
+
+	if err := c.gc.CollectCluster(c.cluster.Name, garbagecollection.NullUID); err != nil {
```
hongchaodeng: Probably not move this down here.
We should delete backup RS first before deleting pvc

xiang90: I moved this back. But still cluster.Delete looks strange now. I might want to not use gc.collect to remove k8s related resources in the future. Reusing that little code makes the code read worse.

pkg/cluster/backup_manager.go
```diff
@@ -132,16 +132,10 @@ func (bm *backupManager) createBackupService() error {
 }
 
 func (bm *backupManager) cleanup() error {
-	cl, c := bm.cluster, bm.config
-	err := k8sutil.DeleteBackupReplicaSetAndService(c.KubeCli, cl.Name, cl.Namespace)
```
hongchaodeng: Also delete DeleteBackupReplicaSetAndService() method in k8sutil.go?

hongchaodeng:
change waitBackupDeleted() in e2e/util.go to also check deletionTimestamp. Just like backup pod:
https://github.com/coreos/etcd-operator/blob/master/test/e2e/util.go#L292-L308

hongchaodeng:
I think we also need to change gc.go to delete rs with orphanChild option to be `false`:

https://github.com/coreos/etcd-operator/blob/master/pkg/util/k8sutil/backup.go#L257-L262

Setting grace period to 0 will still leave 30s gp for backup pod though.

xiang90:
@etcd-bot retest this please.

hongchaodeng:
lgtm

Issue #652, opened 2017-01-15
controller restore (due to rev compaction) doesn't need to "kill" clusters


hongchaodeng:
Currently, if APIServer returns error like StatusGone (i.e. rev compacted), controller will need to restart:
- signals all cluster "manager" to delete (or "exit").
- waits them
- relists all clusters and restarts those cluster "managers"
- watches again.

During this restart, some knowledge will be lost (e.g. membership) and need to reinstated. We can probably optimize this without "killing" all clusters.

xiang90:
#727 solves this in another way. We do not really need to restart controller in every out of date event. We only restart when necessary. So we do not need to worry about the inefficiency of the state build that much. Of course, we can still try to optimize it (not killing clusters even if restart happens) in the future.

/cc @hongchaodeng

PR #727, opened 2017-02-02
controller: best effort recovery on watch failure

xiang90:
If this patch works, then operator will not restart when there is not external issues.

Fix https://github.com/coreos/etcd-operator/issues/652.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -32,7 +32,6 @@ import (
 
 	"github.com/Sirupsen/logrus"
 	"k8s.io/client-go/1.5/kubernetes"
-	"k8s.io/client-go/1.5/pkg/api/unversioned"
 	"k8s.io/client-go/1.5/pkg/api/v1"
 	v1beta1extensions "k8s.io/client-go/1.5/pkg/apis/extensions/v1beta1"
 )
@@ -48,11 +47,6 @@ var (
 	initRetryWaitTime = 30 * time.Second
 )
 
-type rawEvent struct {
-	Type   string
-	Object json.RawMessage
-}
-
 type Event struct {
 	Type   string
 	Object *spec.EtcdCluster
@@ -60,10 +54,14 @@ type Event struct {
 
 type Controller struct {
 	logger *logrus.Entry
-
 	Config
-	clusters    map[string]*cluster.Cluster
-	stopChMap   map[string]chan struct{}
+
+	// TODO: combine the three cluster map.
+	clusters map[string]*cluster.Cluster
+	// Kubernetes resource version of the clusters
+	clusterRVs map[string]string
+	stopChMap  map[string]chan struct{}
+
 	waitCluster sync.WaitGroup
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controller/controller.go
```diff
@@ -288,16 +296,29 @@ func (c *Controller) monitor(watchVersion string) (<-chan *Event, <-chan error)
 				}
 
 				if st != nil {
-					if st.Code == http.StatusGone { // event history is outdated
-						errCh <- ErrVersionOutdated // go to recovery path
+					resp.Body.Close()
```
hongchaodeng: This could double Close().

hongchaodeng: Sorry.. It seems like it always break in this path.

pkg/controller/util.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+
+	"github.com/coreos/etcd-operator/pkg/spec"
+
+	"k8s.io/client-go/1.5/pkg/api/unversioned"
+)
+
```
hongchaodeng: Move `rawEvent` to this package?

hongchaodeng:
test failure due to operator pod was shut down early:
```
"2017-02-02T06:20:49Z"  shutting down, got signal: Terminated
```

e2e log:
```
time="2017-02-02T06:27:25Z" level=info msg="e2e teardown successfully" 
```

xiang90:
> test failure due to operator pod was shut down early:

Can you explain this more with context? I do not quite catch it.

hongchaodeng:
At "06:27:25', e2e finished and would do cleanup -- operator pod should have been shut down at this time.

From operator logs, it was shutdown way earlier -- "2017-02-02T06:20:49Z". This is also when the first test started to fail.

xiang90:
> From operator logs, it was shutdown way earlier 

Can you get the operator log to check why it shutdown?

hongchaodeng:
> check why it shutdown

I can only verify that it wasn't any Fatalf() call from within operator.

xiang90:
@etcd-bot retest this please

pkg/controller/util.go
xiang90: done.

hongchaodeng:
@etcd-bot retest this please

xiang90:
@hongchaodeng Have you got a chance to test it manually? basically we have to write to other endpoints (no the endpoint operator watches on) to trigger both timeout + out of date error.

hongchaodeng:
@xiang90 
I am working on triggering such error.
How does writing to endpoints would help trigger that?

xiang90:
> How does writing to endpoints would help trigger that?

We need to bump resource version and not increase the resource version of the watch. So that once the the watch gets timed out, rewatch on the old resource version will trigger the out of date error.

hongchaodeng:
Manually verified that this will re-watch and continue to work on version compacted case.

Can you also add a logging if controller decided to rebuild the state? It's somewhat expensive and logging would help users to have a clue.

pkg/controller/controller.go
```diff
@@ -288,16 +291,29 @@ func (c *Controller) monitor(watchVersion string) (<-chan *Event, <-chan error)
...
+
```
hongchaodeng: Can you add a logging here StatusGone?
It's not normal path. If LIST-ing/rebuilding cache is expensive, users might need to know what happen?

xiang90:
I can add some debug logging. I am not a big fan of adding logging when users need to take zero action. The aborted watch is expected for etcd operator as a long running process. If we believe user should know about the cost, we can add metrics.

hongchaodeng:
After a second thought, I think metrics is the right way to go.
Let's move forward this, and create an issue to track what metrics we need to add.

PR #766, opened 2017-02-11
client: simple backup client

xiang90:


client/experimentalclient/backup.go
```diff
@@ -0,0 +1,58 @@
...
+
+func NewBackup(c *http.Client, clusterName string) Backup {
+	return &backupClient{
+		client: c,
+		url:    k8sutil.BackupServiceAddr(clusterName),
+	}
+}
+
+func (b backupClient) Request(ctx context.Context) error {
+	req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("http://%s/backupnow", path.Join(b.url, backup.APIV1)), nil)
```
hongchaodeng: maybe `MethodPost`? Although we know it doesn't matter.

xiang90: it matters. the method should be a simple get not post.

hongchaodeng:
@etcd-bot retest this please

hongchaodeng:
@etcd-bot retest this please

client/experimentalclient/client.go
```diff
@@ -1,4 +1,4 @@
-package experimental
+package experimentalclient
```
hongchaodeng: Just for testing. Can you not change this?
I doubt the init() has some impact that we didn't know.

xiang90: we should not call anything just experimental. it is anti go pattern. client-> experimentalclient. we need to repeat the it. client.Backup makes sense, but experimental.Backup does not make sense.

hongchaodeng: I have figured out why the CI failed.
I can reproduce it by importing this package.
The init() has unexpected impact.

xiang90: i am going to comment out the init for now. i will investigate it more when we plan to use the operator client.

client/experimentalclient/client.go
```diff
@@ -0,0 +1,161 @@
...
+	Update(ctx context.Context, name string, spec spec.ClusterSpec) error
+
+	// Get gets the etcd cluster information.
+	Get(ctx context.Context, name string) (*spec.EtcdCluster, error)
+
+	// List lists all etcd clusters.
+	List(ctx context.Context) (*spec.EtcdClusterList, error)
+}
+
+// var (
```
hongchaodeng: Why not remove them?

test/e2e/util.go
```diff
@@ -81,9 +81,8 @@ func makeBackup(f *framework.Framework, clusterName string) error {
 		return fmt.Errorf("no backup pod found")
 	}
 
-	// We are assuming pod ip is accessible from test machine.
-	addr := fmt.Sprintf("%s:%d", podList.Items[0].Status.PodIP, constants.DefaultBackupPodHTTPPort)
-	err = cluster.RequestBackupNow(addr)
+	// We are assuming Kubernetes network is accessible from test machine.
+	err = cluster.RequestBackup(clusterName)
```
hongchaodeng: This isn't `clusterName`.
It should be based on pod ip.

xiang90:
@hongchaodeng PTAL. I do not think we should remove current TPR implementation, and replace it with raw HTTP implementation. Or we are going backwards. We should figure out the cause of the failure. Also it should not be a blocker of this PR.

client/experimentalclient/client.go
```diff
@@ -41,6 +41,13 @@ var (
 )
 
 func init() {
+	// FIXME: if we add this scheme, Kubernetes client will not work
+	// correctly. Fix it before use operator trp client.
```
hongchaodeng: trp -> tpr

client/experimentalclient/backup.go
```diff
@@ -0,0 +1,65 @@
...
+	// Or it returns an error.
+	Request(ctx context.Context) error
+}
+
+type backupClient struct {
+	client *http.Client
+	url    string
+}
+
+func NewBackup(c *http.Client, clusterName string) Backup {
```
hongchaodeng: clusterName -> url

client/experimentalclient/backup.go
```diff
@@ -0,0 +1,65 @@
...
+
+func NewBackupWithURL(c *http.Client, url string) Backup {
+	return &backupClient{
+		client: c,
+		url:    url,
+	}
+}
+
+func (b backupClient) Request(ctx context.Context) error {
+	req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("http://%s/backupnow", path.Join(b.url, backup.APIV1)), nil)
```
hongchaodeng: We should probably move http's APIV1 to here.

hongchaodeng:
> "github.com/coreos/etcd-operator/pkg/backup"
> "github.com/coreos/etcd-operator/pkg/util/k8sutil"

We should get rid of the above dep in backup client. See my comments.

client/experimentalclient/backup.go
xiang90: This is the clusterName not url. See NewBackupWithURL if you want to use URL

client/experimentalclient/backup.go
xiang90: what do you mean? can you give me an example?

xiang90:
> We should get rid of the above dep in backup client. See my comments.

k8sutil is fine. Yes, we should not rely on Backup. I need to separate out the API between backup client/ server to backupapi pkg. it should not live in client library like what Kubernetes does (k8s is probably wrong).

I will do it after this PR.

client/experimentalclient/backup.go
hongchaodeng: APIV1 is put here as `BackupAPIV1`, or `experimentalclient/backup.APIV1` .
backup/ imports it and uses APIv1.

client/experimentalclient/backup.go
hongchaodeng: It's fine as long as you want to have k8sutil in this package. Otherwise, we should remove this func.

client/experimentalclient/backup.go
xiang90: No. This is not what should happen. backup should not import its client and its client should not import the backup server pkg either. there needs to be a backupAPI pkg. we will do that in another pr.

client/experimentalclient/backup.go
xiang90: importing a util is OK. we can important it from everywhere. the current problem is that the util pkg imports have too many deps.

client/experimentalclient/backup.go
hongchaodeng: > backup should not import its client and its client should not import the backup server pkg either.

I don't catch that. Isn't this PR importing the backup server pkg right now.

xiang90: see my last two sentences.

hongchaodeng: Made sense.  Please put a TODO.

hongchaodeng:
lgtm

Issue #722, opened 2017-02-01
operator: add watchdog


xiang90:
Operator can gracefully recover from a restart.

To program defensively, we should add a watchdog feature to simply kill the operator itself
if it detects a potential deadlock or unbounded blocking.

xiang90:
fixed by #775

PR #775, opened 2017-02-15
controller: add panicTimer

xiang90:
Panic the controller when it blocks on one operation for too long.
The operation should be non-blocking. Kubernetes should be able to
restart the operator when it panics.

This is just a defensive approach.


xiang90:
fix https://github.com/coreos/etcd-operator/issues/722

hongchaodeng:
lgtm

Issue #777, opened 2017-02-15
Complete failure?


djMax:
We had a cluster with 3 nodes and backup configured. We had some sort of catastrophic event (still unclear what). First, all 3 etcd pods went away and it made no attempt to restart them. Then the backup pod went away after some hours, and no attempt to recreate. The volumes are still there, is there any way to recover?

xiang90:
@djMax Yes. As long as the volume is still there, etcd-operator can recover it automatically. You can create a new cluster with restore policy set: https://github.com/coreos/etcd-operator/blob/master/pkg/spec/cluster.go#L93.

You need to fill in the original storage type and the cluster name.

Please try this first. There is a backup recovery plan too. But let's see if this works for you.

xiang90:
Also can you get the log of the etcd operator for us? We would like to investigate into the root cause.

djMax:
Me too. I think the operator pod was restarted at some point and I can't find logs from the time period. Now it just says:

```
time="2017-02-15T15:30:52Z" level=info msg="etcd-operator Version: 0.2.0"
time="2017-02-15T15:30:52Z" level=info msg="Git SHA: 0ed2c67"
time="2017-02-15T15:30:52Z" level=info msg="Go Version: go1.7.5"
time="2017-02-15T15:30:52Z" level=info msg="Go OS/Arch: linux/amd64"
time="2017-02-15T15:31:37Z" level=info msg="use in cluster client from k8s library"
time="2017-02-15T15:31:38Z" level=info msg="finding existing clusters..." pkg=controller
time="2017-02-15T15:31:38Z" level=info msg="starts running from watch version: 13732244" pkg=controller
time="2017-02-15T15:31:38Z" level=info msg="start watching at 13732244" pkg=controller```

And there IS a valid cluster, but it doesn't seem to care.

**Comment by xiang90** [comment]
@djMax Can you get the information of that cluster in yaml format via kubectl? (<Yourcluster> -o yaml)

**Comment by djMax** [comment]
Sorry, I can't do the translation of that code to yaml yet. Is that just

```
apiVersion: "coreos.com/v1"
kind: "EtcdCluster"
metadata:
  name: "etcd-cluster"
spec:
  size: 3
  version: "3.1.0"
  restore: true
  backup:
    backupIntervalInSecond: 30
    maxBackups: 5
    volumeSizeInMB: 512
    storageType: "PersistentVolume"```

djMax:
[...]

PR #778, opened 2017-02-15
Pin operator version

djMax:
As discussed in #777


example/deployment-aws.yaml
```diff
@@ -12,6 +12,7 @@ spec:
       containers:
       - name: etcd-operator
         image: quay.io/coreos/etcd-operator
```
xiang90: @djMax change this to quay.io/coreos/etcd-operator:v0.0.2?

we will update this deployment example when we bump etcd operator version.

Issue #485, opened 2016-12-20
doc that removing operator deployment wont affect any running clusters


xiang90:
Users might have the impression that removing etcd operator == removing all managed etcd clusters by the operator.

We intentionally do not remove any existing etcd clusters no matter what happen to the operator. Or an accidental deletion of the operator will remove ALL clusters, it is VERY DANGEROUS. 

A managed etcd will only be deleted if user deleted the TPR, and operator receives the deletion request. Or nothing will happen.

PR #782, opened 2017-02-15
README: clarify that deletion of the operator will not delete etcd

xiang90:
Fix https://github.com/coreos/etcd-operator/issues/485

/cc @hasbro17 @hongchaodeng

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,8 @@ NAME                      DESCRIPTION             VERSION(S)
 cluster.etcd.coreos.com   Managed etcd clusters   v1beta1
 ```
 
+Note that the etcd clusters managed by the etcd operator will not be deleted if the operator fails or the deployment of the operator is deleted. This is an intentional design to prevent accidental operator failure from killing all the etcd clusters.
+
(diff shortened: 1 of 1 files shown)


hasbro17:
LGTM

README.md
```diff
@@ -48,6 +48,8 @@ NAME                      DESCRIPTION             VERSION(S)
 cluster.etcd.coreos.com   Managed etcd clusters   v1beta1
 ```
 
+Note that the etcd clusters managed by the etcd operator will not be deleted if the operator fails or the deployment of the operator is deleted. This is an intentional design to prevent accidental operator failure from killing all the etcd clusters.
```
hongchaodeng: make it shorter, just:

Note: Deleting deployment of etcd operator will not kill etcd clusters.

xiang90: there is also the failure part which is different from the deleting deployment. Also i want to point it is by design not a flaw or something. we should do it since people asked about why before.

hongchaodeng: ok.

PR #785, opened 2017-02-15
backup: expose backup service status

xiang90:
/cc @hongchaodeng @hasbro17 

I am going to add a method to get the status in the backup client library.

### DIFF
--- a/pkg/backup/backup.go
+++ b/pkg/backup/backup.go
@@ -21,6 +21,7 @@ import (
 	"path"
 	"time"
 
+	"github.com/coreos/etcd-operator/pkg/backup/backupapi"
 	"github.com/coreos/etcd-operator/pkg/backup/env"
 	"github.com/coreos/etcd-operator/pkg/backup/s3"
 	"github.com/coreos/etcd-operator/pkg/spec"
@@ -38,6 +39,8 @@ import (
 
 const (
 	PVBackupV1 = "v1"
+
+	maxRecentBackupStatusCount = 10
 )
 
 type Backup struct {
@@ -51,6 +54,9 @@ type Backup struct {
 	be backend
 
 	backupNow chan chan error
+
+	// recentBackupStatus keeps the statuses of 'maxRecentBackupStatusCount' recent backups.
+	recentBackupsStatus []backupapi.BackupStatus
 }
 
 func New(kclient kubernetes.Interface, clusterName, ns string, policy spec.BackupPolicy, listenAddr string) *Backup {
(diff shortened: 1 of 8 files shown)


pkg/backup/backend.go
```diff
@@ -17,7 +17,11 @@ package backup
 import "io"
 
 type backend interface {
-	save(version string, snapRev int64, rc io.Reader) error
+	save(version string, snapRev int64, rc io.Reader) (size int64, err error)
```
hongchaodeng: version -> etcdVersion
rc -> reader/r

mind add a comment to this method:
```
// save saves the snapshot from reader. The snapshot has given etcdVersion and snapRev.
// It returns the size of the snapshot saved.
save(...
```

pkg/backup/backupapi/types.go
```diff
@@ -0,0 +1,37 @@
...
+	// the backup service
+	RecentBackup *BackupStatus `json:"recentBackup,omitempty"`
+
+	// Backups is the totoal number of existing backups
+	Backups int `json:"backups"`
+}
+
+type BackupStatus struct {
+	// Size is the size of the backup.
+	Size int64 `json:"size"`
```
hongchaodeng: Also add revision?

pkg/backup/http.go
```diff
@@ -34,6 +35,7 @@ const (
 func (b *Backup) startHTTP() {
 	http.HandleFunc(backupapi.APIV1+"/backup", b.serveSnap)
 	http.HandleFunc(backupapi.APIV1+"/backupnow", b.serveBackupNow)
+	http.HandleFunc(backupapi.APIV1+"/status", b.serveStatus)
```
hongchaodeng: We also need to update backup_service.md . Are you going to do it in separate PR?

pkg/backup/backup.go
```diff
@@ -198,7 +204,22 @@ func (b *Backup) writeSnap(m *etcdutil.Member, rev int64) error {
...
+		return err
+	}
+
+	bs := backupapi.BackupStatus{
+		Size:     n,
+		Version:  resp.Version,
+		TimeTook: time.Since(start),
+	}
+	b.recentBackupsStatus = append(b.recentBackupsStatus, bs)
+	if len(b.recentBackupsStatus) > 10 {
```
hongchaodeng: define const of 10.

pkg/backup/http.go
xiang90: i will update it in this pr.

pkg/backup/backupapi/types.go
xiang90: not for now. i do not want to add more fields until we have a concrete use case.

xiang90:
@hongchaodeng PTAL

hongchaodeng:
lgtm

PR #794, opened 2017-02-17
k8sutil: change make -> new

xiang90:
Let us follow the go rules. Use new to indicate we
create a pointer.


xiang90:
ah. it is already fixed.

PR #796, opened 2017-02-17
controller: change event as const

xiang90:
/cc @hongchaodeng @hasbro17

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/pkg/api/v1"
 	v1beta1extensions "k8s.io/client-go/pkg/apis/extensions/v1beta1"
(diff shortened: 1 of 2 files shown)


pkg/controller/controller.go
```diff
@@ -46,6 +46,15 @@ var (
 	initRetryWaitTime = 30 * time.Second
 )
 
+const (
+	// Add event type from Kubernetes watch response
+	watchEventAdd = "ADDED"
```
hongchaodeng: This is already defined in client-go watch/
```
const (
	Added    EventType = "ADDED"
	Modified EventType = "MODIFIED"
	Deleted  EventType = "DELETED"
	Error    EventType = "ERROR"

	DefaultChanSize int32 = 100
)
```

xiang90: ok. i will wait your client 2.0 pr get merged then.

xiang90:
@hongchaodeng PTAL

hongchaodeng:
lgtm

Issue #719, opened 2017-02-01
report test code coverage


xiang90:

hasbro17:
The unit test runs in hack/test unittest_pass will now generate go coverage reports and then send those reports to the [coreos/etcd-operator](https://codecov.io/gh/coreos/etcd-operator) codecov repo.

Still need to configure a few settings for codecov commenting on PRs.

PR #804, opened 2017-02-21
hack/test: generate code coverage report

hasbro17:
hack/test: generate code coverage report

Creates coverage report to be consumed by codecov

#719 

/cc @hongchaodeng @xiang90

### DIFF
--- a/hack/test
+++ b/hack/test
@@ -49,6 +49,7 @@ function e2e_pass {
 }
 
 function unittest_pass {
+	# coverage.txt is the combined coverage report consumed by codecov
 	echo "mode: atomic" > coverage.txt
 	TEST_PKGS=`go list ./pkg/...`
 	for pkg in $TEST_PKGS
--- a/hack/test
+++ b/hack/test
@@ -49,8 +49,16 @@ function e2e_pass {
 }
 
 function unittest_pass {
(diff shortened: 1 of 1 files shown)


hack/test
```diff
@@ -49,8 +49,16 @@ function e2e_pass {
 }
 
 function unittest_pass {
+	echo "mode: atomic" > coverage.txt
```
xiang90: why do we need to output this to coverage.txt?

xiang90:
LGTM

hack/test
```diff
@@ -49,8 +49,16 @@ function e2e_pass {
...
 	TEST_PKGS=`go list ./pkg/...`
-	go test ${TEST_PKGS} --race
+	for pkg in $TEST_PKGS
+	do
+		go test -race -covermode=atomic -coverprofile=profile.out $pkg
+		if [ -f profile.out ]; then
+			tail -n +2 profile.out >> coverage.txt
```
xiang90: comment that we output coverage for coverage.io's consumption?

hack/test
hasbro17: This may not be strictly necessary. It's just how `go test -covermode=<mode>` normally generates a coverage report file with the mode at the top.

PR #810, opened 2017-02-22
hack/test: call codecov bash script from inside unittest

### LINKED ISSUE #719: report test code coverage

hasbro17:
hack/test: add codecov reporting script

Calls the script to send the coverage reports to codecov everytime the unit tests are run.
Previously part of the jenkins bash script.

fix #719 

/cc @hongchaodeng @xiang90

### DIFF
--- a/hack/test
+++ b/hack/test
@@ -60,6 +60,8 @@ function unittest_pass {
 			rm profile.out
 		fi
 	done
+	# Send reports to codecov, CODECOV_TOKEN env must be present in Jenkins
+	curl -s https://codecov.io/bash | bash
 }
 
 for p in $PASSES; do


hack/test
```diff
@@ -60,6 +60,8 @@ function unittest_pass {
 			rm profile.out
 		fi
 	done
+	# Send reports to codecov, CODECOV_TOKEN env must be present in Jenkins
+	curl -s https://codecov.io/bash | bash
```
hongchaodeng: What happened if $CODECOV_TOKEN is not set?

hongchaodeng:
@etcd-bot retest this please

hack/test
hasbro17: The codecov bash uploader will output an HTTP 400 error message but the unit test output script is not affected and continues as normal to either print `test success ===` or `test fail ===` based on the unit tests.

Or I can explicitly check for the presence of $CODECOV_TOKEN and not call the codecov script if that seems better.

hongchaodeng:
lgtm

Issue #696, opened 2017-01-29
add backup status


xiang90:
we should add backup status into TPR including:

- [x] created time of most recent backup
- [X] size of the most recent backup
- [X] time spent on saving the most recent backup
- [X] number of backups
- [x] total size of backups

PR #816, opened 2017-02-23
*: add backup creation time

xiang90:


pkg/spec/backup.go
```diff
@@ -71,6 +73,9 @@ type BackupServiceStatus struct {
 }
 
 type BackupStatus struct {
+	// Creation time of the backup.
+	Created time.Time `json:"created"`
```
hongchaodeng: createdTime?

Reading "created" from json would be hard to interpret.

hongchaodeng:
xref: https://github.com/coreos/etcd-operator/issues/696

pkg/spec/backup.go
xiang90: OK.

xiang90:
PTAL

hongchaodeng:
lgtm

PR #819, opened 2017-02-23
*: add total size into backup status

xiang90:
Also change size from int64 to float64. It is not a breaking change since JSON does not differentiate about int and float (actually number is double type in JSON).

### DIFF
--- a/pkg/backup/backupapi/types.go
+++ b/pkg/backup/backupapi/types.go
@@ -21,16 +21,19 @@ type ServiceStatus struct {
 	// the backup service
 	RecentBackup *BackupStatus `json:"recentBackup,omitempty"`
 
-	// Backups is the totoal number of existing backups
+	// Backups is the totoal number of existing backups.
 	Backups int `json:"backups"`
+
+	// BackupSize is the total size of existing backups in MB.
+	BackupSize float64 `json:"backupSize"`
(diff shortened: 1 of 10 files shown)


pkg/backup/backupapi/types.go
```diff
@@ -21,16 +21,19 @@ type ServiceStatus struct {
 	// the backup service
 	RecentBackup *BackupStatus `json:"recentBackup,omitempty"`
 
-	// Backups is the totoal number of existing backups
+	// Backups is the totoal number of existing backups.
 	Backups int `json:"backups"`
+
+	// BackupSize is the total size of existing backups in MB.
+	BackupSize float64 `json:"backupSize"`
```
hongchaodeng: bakcupSizeInMB?

hongchaodeng:
ref: https://github.com/coreos/etcd-operator/issues/696

pkg/backup/backupapi/types.go
xiang90: we have size in backup status struct. i want to keep them consistent.

i also feel adding inMB is better. However, we cannot really change the one in backup status for compatibility reason. I do not think this is a big deal either.

hongchaodeng:
lgtm

PR #822, opened 2017-02-24
*: expose pv name to backup status

xiang90:


pkg/spec/backup.go
```diff
@@ -73,6 +73,9 @@ type BackupServiceStatus struct {
 
 	// BackupSize is the total size of existing backups in MB.
 	BackupSize float64 `json:"backupSize"`
+
+	// StorageStatus is the status of the storage the backup pod uses.
+	StorageStatus *StorageStatus `json:"storageStatus,omitempty"`
```
hongchaodeng: There are too many layer of status:
```
status:
  backupStatus:
    storageStatus:
      pvStatus:
```

This is good ehough:
```
status:
  backupStatus:
      pv:
        name:
```

If you agree, then I would suggest make this `inline`, and change the names too.

xiang90: OK. I will make this inline.

xiang90:
After second thought, I decided to not expose pv name though status. This is derivable from the PVC name. As kubernetes admin (this is mainly useful for self hosted etcd case), they should already have the knowledge to get the PV/disk name from bounded PVC.

PR #834, opened 2017-02-27
test/e2e: add env to skip tests

hasbro17:
Added an environment variable to skip 3 tests(ClusterRestore,SelfHosted and BackupStatus) that fail consistently on an AWS cluster. This is for ease of debugging on AWS.

/cc @hongchaodeng @xiang90

### DIFF
--- a/test/e2e/cluster_status_test.go
+++ b/test/e2e/cluster_status_test.go
@@ -27,7 +27,7 @@ import (
 
 func TestBackupStatus(t *testing.T) {
 	if os.Getenv(framework.EnvCloudProvider) == "aws" {
-		t.Skip("skipping...")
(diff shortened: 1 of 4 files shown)


test/e2e/cluster_status_test.go
```diff
@@ -25,6 +26,10 @@ import (
 )
 
 func TestBackupStatus(t *testing.T) {
+	if os.Getenv(framework.EnvCloudProvider) == "aws" {
```
xiang90: add todo to remove this skip.

xiang90:
LGTM

test/e2e/cluster_status_test.go
```diff
@@ -25,6 +26,10 @@ import (
...
+		t.Skip("skipping...")
```
hongchaodeng: Maybe add the todo in the message.

Issue #840, opened 2017-02-28
e2e: have rbac roles for etcd operator 


hongchaodeng:
Our e2e test against k8s master version cluster failed due to RBAC limits. We need to create roles for operator in e2e test if RBAC is enabled.

xiang90:
this is already resolved by #850

PR #850, opened 2017-03-02
ci: create role and binding if rbac present

hasbro17:
Create a ClusterRole and ClusterRoleBinding if k8s supports rbac.
Also change the rbac documentation for the correct verb to use for permissions to the configmap and secrets.

#840 

/cc @xiang90 @hongchaodeng

### DIFF
--- a/hack/ci/create_rbac_role
+++ b/hack/ci/create_rbac_role
@@ -1,0 +1,74 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# Create ClusterRole
+cat <<EOF | kubectl create -f -
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
+metadata:
+  name: etcd-operator
+rules:
+- apiGroups:
+  - etcd.coreos.com
+  resources:
+  - clusters
+  verbs:
+  - "*"
+- apiGroups:
+  - extensions
+  resources:
+  - thirdpartyresources
+  verbs:
+  - create
+- apiGroups:
+  - storage.k8s.io
+  resources:
+  - storageclasses
+  verbs:
+  - create
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  - services
+  - endpoints
+  - persistentvolumeclaims
+  verbs:
+  - "*"
+- apiGroups:
+  - extensions
+  resources:
+  - replicasets
+  verbs:
+  - "*"
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  - configmaps
+  verbs:
+  - get
+EOF
+
+# Create ClusterRoleBinding in namespace
+cat <<EOF | kubectl create -f -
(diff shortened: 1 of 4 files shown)


hack/ci/create_rbac_role
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# Create ClusterRole
+cat <<EOF | kubectl create -f -
```
xiang90: shall we do this in our go e2e/testing instead?

hack/test
```diff
@@ -44,7 +44,24 @@ function build_pass {
...
+	kubectl delete clusterrolebinding etcd-operator
+	kubectl delete clusterrole etcd-operator
+}
+
 function e2e_pass {
+	# Enable rbac if present
+	if kubectl --kubeconfig $KUBECONFIG get clusterrole 1> /dev/null ; then
+		trap rbac_cleanup EXIT
+		# Create role and binding
+		if "hack/ci/create_rbac_role"; then
```
hongchaodeng: Can we have the other file to be a util, e.g. rbac_util.sh? Then `source xxx`.

It should contain two funcs, setup_rbac() and cleanup_rbac().

hack/ci/create_rbac_role
hasbro17: Well it's more convenient as a bash script for immediate purposes so we can test this out with jenkins. Perhaps we can move this into test/e2e/utils.go later.

xiang90: ok. sgtm.

xiang90: maybe add a todo?

hack/ci/rbac_utils.sh
```diff
@@ -0,0 +1,82 @@
...
+function rbac_cleanup {
+	kubectl delete clusterrolebinding etcd-operator
+	kubectl delete clusterrole etcd-operator
+}
+
+function rbac_setup() {
+    # Create ClusterRole
+    cat <<EOF | kubectl create -f -
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
```
hasbro17: The lack of indentation is a little ugly but necessary since it's a heredoc and indenting would mess up the yaml file structure.

hack/ci/rbac_utils.sh
```diff
@@ -0,0 +1,82 @@
...
+  kind: ClusterRole
+  name: etcd-operator
+subjects:
+- kind: ServiceAccount
+  name: default
+  namespace: $TEST_NAMESPACE
+EOF
+
+    # Check if ClusterRole and ClusterRoleBinding created
+    kubectl --kubeconfig=$KUBECONFIG get clusterrole etcd-operator 1> /dev/null
```
hongchaodeng: Remove these two lines?

hongchaodeng:
lgtm

Update: We have tried it locally and the tests did run through. There is very high rate of flakiness k8s master. We are going to merge this and continue our work testing on k8s master.

PR #865, opened 2017-03-07
spec/ClusterStatus: track healthy pods

hasbro17:
Outline for keeping track of number of healthy pods etcd cluster.
My first thought are just to send HTTP GET requests to each running pod after the cluster is reconciled.

Is this the right place to do this?

/cc @hongchaodeng @xiang90

### DIFF
--- a/pkg/spec/cluster.go
+++ b/pkg/spec/cluster.go
@@ -193,6 +193,10 @@ type ClusterStatus struct {
 
 	// Size is the current size of the cluster
 	Size int `json:"size"`
+	// ReadyPods are the etcd pods that are ready to serve requests
+	ReadyPods []string `json:"readyPods"`
+	// UnreadyPods are the etcd pods not ready to serve requests
+	UnreadyPods []string `json:"unreadyPods"`
 	// CurrentVersion is the current cluster version
 	CurrentVersion string `json:"currentVersion"`
 	// TargetVersion is the version the cluster upgrading to.


pkg/cluster/cluster.go
```diff
@@ -301,6 +301,13 @@ func (c *Cluster) run(stopC <-chan struct{}) {
 				break
 			}
 
+			// Send a get request to each running pod at http://PodIP:2379/version
```
hongchaodeng: Please remove these in this PR

pkg/spec/cluster.go
```diff
@@ -193,6 +193,8 @@ type ClusterStatus struct {
 
 	// Size is the current size of the cluster
 	Size int `json:"size"`
+	// HealthySize is the number of pods passing the liveliness probe, HTTP GET on port 2379
```
hongchaodeng: We want to expose the healthy endpoints, not just size.
For example:
```
healthyEndpoints: [ip/host:port, ...]
```

xiang90: probably just name?

hongchaodeng:
> Is this the right place to do this?

In pollPods(), you can use something that I reverted in
https://github.com/coreos/etcd-operator/pull/807/files

Then update the status and eventually it will be updated.

pkg/spec/cluster.go
hasbro17: To be more general what about tracking the pods since we're already checking them?
`healthyPods []*v1.Pod`
Or just the names or endpoints? Your call.

xiang90: I like names better. users can inspect the pod state with k8s api as long as they have the pod name.

hongchaodeng: It's inconvenient to display `[]*Pod` in json.
It's preferred to use `healthyEndpoints []string`.
Since it's endpoints, better be `ip/hostname:port`.

xiang90: @hongchaodeng Why do we want to show endpoints here at all? As a user, why do i care about endpoint? All I want is to see the pod name, so i can inspect it. I can get the endpoint myself to query pod.

hongchaodeng: @xiang90 
Made sense, I'm replying to this comment

> To be more general what about tracking the pods since we're already checking them? Then we can get the names and endpoints both.
healthyPods []*v1.Pod

hasbro17:
@hongchaodeng PTAL. I'll push the actual implementation of the health check in pollPods() in a separate PR.

pkg/spec/cluster.go
```diff
@@ -193,6 +193,8 @@ type ClusterStatus struct {
 
 	// Size is the current size of the cluster
 	Size int `json:"size"`
+	// HealthyEndpoints are the 'PodIP:Port' endpoints passing the HTTP GET liveliness probe
+	HealthyEndpoints []string `json:"healthyEndpoints"`
```
hongchaodeng: I think our final conclusion is to use `ReadyPods []string`, probably also provide info for `UnreadyPods []string` as well. See https://github.com/coreos/etcd-operator/pull/865#discussion_r104585695

pkg/spec/cluster.go
```diff
@@ -193,6 +193,10 @@ type ClusterStatus struct {
 
 	// Size is the current size of the cluster
 	Size int `json:"size"`
+	// ReadyPods are the pods passing the HTTP GET liveliness probe
```
hongchaodeng: > the HTTP GET liveliness probe

That's internal thing users shouldn't know about.
Probably:
```
ReadyPods are the etcd pods that are ready to serve requests
```

hongchaodeng:
lgtm

PR #875, opened 2017-03-09
operator: remove duplicate config creation

GrapeBaBa:
This commit remove duplicate invocation
k8sutil.InClusterConfig() in operator main

Signed-off-by: grapebaba <281165273@qq.com>

### DIFF
--- a/pkg/util/k8sutil/k8sutil.go
+++ b/pkg/util/k8sutil/k8sutil.go
@@ -315,12 +315,7 @@ func InClusterConfig() (*rest.Config, error) {
 	return rest.InClusterConfig()
 }
 
-func NewTPRClient() (*rest.RESTClient, error) {
-	config, err := InClusterConfig()
-	if err != nil {
-		return nil, err
-	}
-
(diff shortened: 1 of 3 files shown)


xiang90:
@GrapeBaBa What is the problem with invoking `k8sutil.InClusterConfig()` twice?

pkg/util/k8sutil/k8sutil.go
```diff
@@ -315,12 +315,7 @@ func InClusterConfig() (*rest.Config, error) {
 	return rest.InClusterConfig()
 }
 
-func NewTPRClient() (*rest.RESTClient, error) {
```
hongchaodeng: It's by design to make NewTPRClient() self-contained.
Again, what's the problem to call InClusterConfig() twice?

GrapeBaBa: oh, I didn't recognize this design. Not big issue, just looked the code and suspect it can be more elegant.

GrapeBaBa:
Close this PR as it is not a issue.

Issue #806, opened 2017-02-21
user: don't run as root


philips:
The etcd operator container is running as the root user by default. There is no reason to do this, please change to a UID other than 0.

hongchaodeng:
@philips 
I'm just curious: most, if not all, Dockerfiles that I have seen are using default user account. What was the reason to change it? Any security concern?

philips:
The default to root behavior is a hold over for backwards compatibility from early Docker days AFAIK. There is even a warning in the doc: https://docs.docker.com/engine/reference/builder/#user

PR #881, opened 2017-03-10
Dockerfile: run image as non-root user

hasbro17:
The etcd-operator container image will run as a non-root user: etcd-operator.

Waiting to see if any tests break first so don't merge yet.

#806

### DIFF
--- a/hack/build/operator/Dockerfile
+++ b/hack/build/operator/Dockerfile
@@ -1,8 +1,11 @@
 FROM alpine:3.5
 
+RUN adduser -D etcd-operator
 RUN apk add --no-cache ca-certificates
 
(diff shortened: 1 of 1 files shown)


hack/build/operator/Dockerfile
```diff
@@ -1,8 +1,11 @@
 FROM alpine:3.5
 
+RUN adduser -D etcd-operator
 RUN apk add --no-cache ca-certificates
 
 ADD _output/bin/etcd-operator /usr/local/bin
 ADD _output/bin/etcd-backup /usr/local/bin
 
+USER etcd-operator
```
hongchaodeng: This is more than just etcd operator pod. We also need to think about:
- backup pod
- etcd pod

PR #899, opened 2017-03-13
spec: add cluster tls configuration

colhom:
\cc @xiang90 @hongchaodeng

### DIFF
--- a/doc/user/cluster_tls.md
+++ b/doc/user/cluster_tls.md
@@ -1,0 +1,68 @@
+# Cluster TLS guide
+
+Cluster TLS policy is configured on a per-cluster basis via the TPR spec provided to etcd-operator.
+
+```yaml
+apiVersion: "etcd.coreos.com/v1beta1"
+kind: "Cluster"
+metadata:
+  name: "example-etcd-cluster"
+spec:
+  ...
+  TLS:
+    ....
+```
+
+For a review etcd's TLS support and requirements, please read the relevant section from the [operations guide](https://coreos.com/etcd/docs/latest/op-guide/security.html).
+
+
+## Static cluster TLS Policy
+
+```yaml
+apiVersion: "etcd.coreos.com/v1beta1"
+kind: "Cluster"
+metadata:
+  name: "example-etcd-cluster"
+ spec:
+  ...
+  TLS:
+    static:
+      serverSecretName: server-tls-secret
+      clientSecretName: client-tls-secret
+
+```
+
+* **serverSecretName**: contains pem-encoded private keys and x509 certificates needed by the etcd server.
+
+  etcd-operator will mount this secret at `/etc/etcd-operator/server-tls` for each etcd member pod in the cluster.
+
+  The server TLS assets are expected to conform to the following structure:
+
+  ```text
+  /etc/etcd-operator/server-tls/
+
+       server-client-crt.pem
+       server-client-key.pem
+       ca-client-crt.pem
+
+       server-peer-crt.pem
+       server-peer-key.pem
+       ca-peer-crt.pem
+  ```
+
+  How these files are used by the etcd server is outlined in the [security flags section of the etcd admin guide](https://github.com/coreos/etcd/blob/master/Documentation/op-guide/configuration.md#security-flags).
+
+* **clientSecretName**: contains pem-encoded private-key and x509 certificates needed to access etcd client interface. This identity is used by `etcd-operator` and backup sidecar to access the cluster's client interface. The secret will be mounted `/etc/etcd-operator/client-tls`.
+
+  The client TLS assets are expected to conform to the following structure:
+
+  ```text
+  /etc/etcd-operator/client-tls/
+
+    client-crt.pem
+    client-key.pem
+    ca-crt.pem
+
+  ```
+
+  These files are similar too the `--cert-file`,`--key-file`, and `--ca-file` arguments (respectively) to `etcdctl`.
--- a/pkg/spec/cluster.go
+++ b/pkg/spec/cluster.go
@@ -99,6 +99,9 @@ type ClusterSpec struct {
 	// SelfHosted determines if the etcd cluster is used for a self-hosted
 	// Kubernetes cluster.
 	SelfHosted *SelfHostedPolicy `json:"selfHosted,omitempty"`
+
+	// etcd cluster TLS configuration
+	TLS *TLSPolicy `json:"TLS"`
 }
 
 // RestorePolicy defines the policy to restore cluster form existing backup if not nil.
--- a/pkg/spec/cluster_tls.go
+++ b/pkg/spec/cluster_tls.go
@@ -1,0 +1,29 @@
+// Copyright 2017 The etcd-operator Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package spec
+
+// TLSPolicy defines the TLS policy of an etcd cluster
+type TLSPolicy struct {
+	// StaticTLS enables user to generate static x509 certificates and keys,
+	// put them into Kubernetes secrets, and specify them into here.
+	Static *StaticTLS `json:"static"`
+}
+
+type StaticTLS struct {
+	// ServerSecretName contains peer-interface and client-interface server x509 key/cert, along with peer and client CA cert.
+	ServerSecretName string `json:"serverSecretName"`
+	// ClientSecretName contains etcd client key/cert, along with client CA cert.
+	ClientSecretName string `json:"clientSecretName"`
+}


example/example-etcd-cluster.yaml
```diff
@@ -5,3 +5,8 @@ metadata:
 spec:
   size: 3
   version: "3.1.2"
+  clusterTLS:
```
hongchaodeng: Please don't change example

pkg/spec/cluster.go
```diff
@@ -99,6 +99,9 @@ type ClusterSpec struct {
 	// SelfHosted determines if the etcd cluster is used for a self-hosted
 	// Kubernetes cluster.
 	SelfHosted *SelfHostedPolicy `json:"selfHosted,omitempty"`
+
+	// etcd cluster TLS configuration
+	ClusterTLS *ClusterTLSPolicy `json:"clusterTLS"`
```
hongchaodeng: `TLS *TLSPolicy`

pkg/spec/cluster_tls.go
```diff
@@ -0,0 +1,30 @@
+// Copyright 2016 The etcd-operator Authors
```
hongchaodeng: 2017

pkg/spec/cluster_tls.go
```diff
@@ -0,0 +1,30 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package spec
+
+type ClusterTLSPolicy struct {
+	Static *StaticTLS `json:"static"`
+	//TODO: add more TLS options here
```
hongchaodeng: Remove this todo.

pkg/spec/cluster_tls.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+type StaticTLS struct {
+	// Secret containing peer and client CA cert files
+	CASecretName string `json:"caSecretName"`
```
hongchaodeng: Remove this field. We should put client CA to client secret, same to others.

pkg/spec/cluster_tls.go
```diff
@@ -0,0 +1,30 @@
...
+	// Secret containing peer-interface and client-interface server x509 key/cert
+	// Using this approach means all nodes have the same "identity", which is less than ideal
+	NodeSecretName string `json:"nodeSecretName"`
```
hongchaodeng: ServerSecretName

pkg/spec/cluster_tls.go
```diff
@@ -0,0 +1,30 @@
...
+
+type ClusterTLSPolicy struct {
+	Static *StaticTLS `json:"static"`
+	//TODO: add more TLS options here
+}
+
+type StaticTLS struct {
+	// Secret containing peer and client CA cert files
+	CASecretName string `json:"caSecretName"`
+	// Secret containing peer-interface and client-interface server x509 key/cert
```
hongchaodeng: ServerSecretName contains ...

hongchaodeng: "." after "cert".

pkg/spec/cluster_tls.go
```diff
@@ -0,0 +1,30 @@
...
+	// Using this approach means all nodes have the same "identity", which is less than ideal
```
hongchaodeng: Remove this line. Spec comments minimally explain what it does and why, no limitation unless there are alternatives.

pkg/spec/cluster_tls.go
```diff
@@ -0,0 +1,30 @@
...
+	NodeSecretName string `json:"nodeSecretName"`
+	// Secret containing etcd client key/cert
+	ClientSecretName string `json:"clientSecretName"`
```
hongchaodeng: We need another doc or inside spec to further explain what exactly are we expecting in Server/ClientSecretName.

colhom:
@hongchaodeng thanks for the quick review. I've added some preliminary docs, PTAL

doc/user/cluster_tls.md
```diff
@@ -0,0 +1,73 @@
...
+Cluster TLS policy is configured on a per-cluster basis via the TPR spec provided to etcd-operator.
+
+```yaml
+apiVersion: "etcd.coreos.com/v1beta1"
+kind: "Cluster"
+metadata:
+  name: "example-etcd-cluster"
+spec:
+  ...
+  clusterTLS:
```
hongchaodeng: TLS

doc/user/cluster_tls.md
```diff
@@ -0,0 +1,73 @@
...
+
+	client-crt.pem
+	client-key.pem
+    ca-crt.pem
+
+  ```
+
+  These files represent the `--cert-file`,`--key-file`, and `--ca-file` arguments (respectively) to `etcdctl`.
+
+  It good to note that etcd-operator will also use these TLS secrets to access the etcd cluster, but will not mount the secret as a filesystem. This is so etcd-operator does not have to change it's own pod definition at runtime :smiley: .
```
hongchaodeng: It -> It is
it's own pod -> its own pod

pkg/spec/cluster.go
```diff
@@ -99,6 +99,9 @@ type ClusterSpec struct {
 	// SelfHosted determines if the etcd cluster is used for a self-hosted
 	// Kubernetes cluster.
 	SelfHosted *SelfHostedPolicy `json:"selfHosted,omitempty"`
+
+	// etcd cluster TLS configuration
+	TLS *TLSPolicy `json:"clusterTLS"`
```
hongchaodeng: json:clusterTLS -> json:TLS

pkg/spec/cluster_tls.go
```diff
@@ -0,0 +1,26 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package spec
+
+type TLSPolicy struct {
```
hongchaodeng: comment:
```
// TLSPolicy defines the TLS policy of an etcd cluster
```

pkg/spec/cluster_tls.go
```diff
@@ -0,0 +1,26 @@
...
+	Static *StaticTLS `json:"static"`
```
hongchaodeng: Comment:
```
// Static enables user to generate static x509 certificates and keys, put them into Kubernetes secrets, and specify them into here.
```

pkg/spec/cluster_tls.go
```diff
@@ -0,0 +1,26 @@
...
+}
+
+type StaticTLS struct {
```
hongchaodeng: Comment:
```
// StaticTLS enables user to generate static x509 certificates and keys, put them into Kubernetes secrets, and specify them into here.
```

doc/user/cluster_tls.md
hongchaodeng: > etcd-operator will also use these TLS secrets to access the etcd cluster, but will not mount the secret as a filesystem.

I don't understand what you are trying to prove here?

hongchaodeng:
cc @ericchiang @philips

xiang90:
LGTM in general.

doc/user/cluster_tls.md
```diff
@@ -0,0 +1,73 @@
...
+* **serverSecretName**: contains pem-encoded private keys and x509 certificates needed by the etcd server.
+
+  etcd-operator will mount this secret at `/etc/etcd-operator/server-tls` for each etcd member pod in the cluster.
+
+  The server TLS assets are expected to conform to the following structure:
+
+  ```text
+  /etc/etcd-operator/server-tls/
+
+       server-client-crt.pem
```
ericchiang: server-client-cert? What's this used for? I don't see an equivalent on the linked document below.

ericchiang: Also what are the expected SANS for these certificates, or does it not matter?

hongchaodeng: @ericchiang let me try to help answer first question:

> What's this used for? I don't see an equivalent on the linked document below.

These are directory/filename formats that we expect users to put their certs/keys in. It's not used yet because this is format spec now. How we use it are implementation details.

ericchiang: @hongchaodeng an external user has to provide these though, right? What should the extended usages be? A serving cert and client cert? Are these used by the underlying etcd cluster to server API traffic and should my clients therefore trust the CA these are generated from?

doc/user/cluster_tls.md
colhom: > I don't understand what you are trying to prove here?

I'm commenting that `etcd-operator` pod will consume the contents of the client tls secret, but will not mount the secret as a volume.

doc/user/cluster_tls.md
```diff
@@ -0,0 +1,73 @@
...
+       server-peer-crt.pem
+       server-peer-key.pem
+       ca-peer-crt.pem
+  ```
+
+  How these files are used by the etcd server is outlined in the [security flags section of the etcd admin guide](https://github.com/coreos/etcd/blob/master/Documentation/op-guide/configuration.md#security-flags).
+
+* **clientSecretName**: contains pem-encoded private-key and x509 certificates needed to access etcd client interface. This identity is used by `etcd-operator` to access the cluster's client interface.
+
+  etcd-operator will mount this secret at `/etc/etcd-operator/client-tls` for every pod that needs to access the etcd cluster's client interface. (eg: backup pod)
```
hongchaodeng: Combine this line to above paragraph.

Change it to:
```
The secret will be mounted `/etc/etcd-operator/client-tls`.
```

doc/user/cluster_tls.md
hongchaodeng: Right. etcd clients would need to trust the CA passed in.

Also @colhom here.

colhom: > Also what are the expected SANS for these certificates, or does it not matter?

On this topic, I really recommend not using IP SANS as we are now. It will really make assigning x509 identities to cluster pods a real pain.

  * Neither pod or service IP addresses are really reservable in Kubernetes. Which means you can't create the certificate until after the pod/service is created, which means you can't really pre-sign certificates in any way

  * The certificate will not be transferable if that etcd node finds itself at a new IP

  * There's no way to sign a certificate for an open-ended group of addresses. So you have to account for all possible hostnames when you sign the cert, which again is problematic. Using DNS you can sign a cert for `*.xxxx.svc.cluster.local`, which is in plain english "any service address in the xxxx kubernetes namespace". This is *really critical* stuff to be able to express to `etcd-operator` from the get-go.

To get past these problems, we need to make sure `etcd-operator`, `backup` and all the etcd member pods *only* address each other via fully-qualified cluster DNS name. For example:

`https://etcd-cluster-0002.my-namespace.svc.cluster.local`

This provides some much-needed flexibility to the end user, such as the ability to sign a generic certificate for all etcd member nodes in a reasonable way. Using IP SANS, this wouldn't really be possible as we would have to know the member IP addresses ahead of time.

A certificate could be generated for `*.my-namespace.svc.cluster.local`, and that certificate will only validate for things running in `my-namespace` on a Kubernetes cluster. I feel this is how a lot of external PKI will initially need to interact with managed components running on clusters, especially if we're not getting unique certificates signed on demand for new member pods.

colhom: @hongchaodeng I feel like we should condense this discussion into a separate issue, TLS work is dependent on it but it's orthogonal.

hongchaodeng: Sure.

This is not really a requested change in the docs. Please address other minor comments and let's move forward.

ericchiang: > I feel like we should condense this discussion into a separate issue, TLS work is dependent on it but it's orthogonal. 

Since you're expecting users to generate these, you should be giving instructions about _what_ to generate. Is that orthogonal to documenting this feature? How would someone use this doc without that information?

doc/user/cluster_tls.md
```diff
@@ -0,0 +1,68 @@
+# Cluster TLS Guide
```
xiang90: Cluster TLS guide (to follow coreos doc rules)

doc/user/cluster_tls.md
```diff
@@ -0,0 +1,68 @@
...
+spec:
+  ...
+  TLS:
+    ....
+```
+
+For a review etcd's TLS support and requirements, please read the relevant section from the [operations guide](https://coreos.com/etcd/docs/latest/op-guide/security.html).
+
+
+## Static Cluster TLS Policy
```
xiang90: Status cluster TLS policy

xiang90:
LGTM!

doc/user/cluster_tls.md
hongchaodeng: Static :)

doc/user/cluster_tls.md
colhom: @ericchiang we need to get https://github.com/coreos/etcd-operator/issues/901 figured out before we can write those instructions, as they will depend very much on how the components are addressing each other when establishing http(s) connections. This is definitely not in scope for this PR.

xiang90:
@colhom Thanks!

PR #920, opened 2017-03-21
Adding none pv provsioner and ignoring existing TP objects

uruddarraju:
There are cases where both s3 and gce pd are not offered and it should be possible to run the etcd operator without needing to register these storage classes. Hence, adding a new pv provider type `none`, which does not try to register the storage class.

The other change in this PR, is in case when the users using `etcd-operator` don't manage the k8s cluster and still want to use it. In this case, the cluster operator may register the cluster thirdparty object manually for the users, because the cluster operator is not inclined to give a clusterrolebinding to the operator, which can impact other controllers running inside the cluster that are also dependent on third party objects.

This is the spec that I used to test out the etcd operator:

```yaml
kind: Deployment
metadata:
  name: etcd-operator
spec:
  replicas: 1
  template:
    metadata:
      labels:
        name: etcd-operator
    spec:
      serviceAccountName: etcd-operator
      containers:
      - name: etcd-operator
        image: uruddarraju/etcd:pv-none
        command:
        - "/bin/sh"
        - "-c"
        - "/usr/local/bin/etcd-operator --pv-provisioner=none"
        env:
        - name: MY_POD_NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace
        - name: MY_POD_NAME
          valueFrom:
            fieldRef:
              fieldPath: metadata.name
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -40,6 +40,7 @@ var (
 	supportedPVProvisioners = map[string]struct{}{
 		"kubernetes.io/gce-pd":  {},
 		"kubernetes.io/aws-ebs": {},
+		"none":                  {},
 	}
 
 	ErrVersionOutdated = errors.New("requested version is outdated in apiserver")
@@ -247,15 +248,23 @@ func (c *Controller) initResource() (string, error) {
 				return "", err
 			}
 		} else {
-			return "", fmt.Errorf("fail to create TPR: %v", err)
+			_, listerr := k8sutil.GetClusterList(c.Config.KubeCli.CoreV1().RESTClient(), c.Config.Namespace)
+			if listerr != nil {
+				return "", fmt.Errorf("fail to create TPR and it does not exist in the api server: %v", err)
+			}
+			watchVersion, err = c.findAllClusters()
 		}
 	}
-	err = k8sutil.CreateStorageClass(c.KubeCli, c.PVProvisioner)
-	if err != nil {
-		if !k8sutil.IsKubernetesResourceAlreadyExistError(err) {
-			return "", fmt.Errorf("fail to create storage class: %v", err)
+
+	if c.PVProvisioner != "none" {
+		err = k8sutil.CreateStorageClass(c.KubeCli, c.PVProvisioner)
+		if err != nil {
+			if !k8sutil.IsKubernetesResourceAlreadyExistError(err) {
+				return "", fmt.Errorf("fail to create storage class: %v", err)
+			}
 		}
 	}
+
 	return watchVersion, nil
 }
 


uruddarraju:
cc @hongchaodeng

hongchaodeng:
> There are cases where both s3 and gce pd are not offered and it should be possible to run the etcd operator without needing to register these storage classes. Hence, adding a new pv provider type none, which does not try to register the storage class.

Can you create an issue to discuss this first?

> The other change in this PR, is in case when the users using etcd-operator don't manage the k8s cluster and still want to use it. In this case, the cluster operator may register the cluster thirdparty object manually for the users, because the cluster operator is not inclined to give a clusterrolebinding to the operator, which can impact other controllers running inside the cluster that are also dependent on third party objects.

I don't catch it. Can you create an issue to discuss this separately? Also give an example?

uruddarraju:
@hongchaodeng done.

pkg/controller/controller.go
```diff
@@ -40,6 +40,7 @@ var (
 	supportedPVProvisioners = map[string]struct{}{
 		"kubernetes.io/gce-pd":  {},
 		"kubernetes.io/aws-ebs": {},
+		"none":                  {},
```
xiang90: if user sets this to none, a cluster with backup PV enabled should fail on startup with a good failure reason. have you tried that?

uruddarraju: @xiang90 nope, let me add that in too. thanks for the catch.

pkg/controller/controller.go
```diff
@@ -247,15 +248,23 @@ func (c *Controller) initResource() (string, error) {
 				return "", err
 			}
 		} else {
-			return "", fmt.Errorf("fail to create TPR: %v", err)
+			_, listerr := k8sutil.GetClusterList(c.Config.KubeCli.CoreV1().RESTClient(), c.Config.Namespace)
```
hongchaodeng: What's this change for?

hongchaodeng:
> The other change in this PR, is in case when the users using etcd-operator don't manage the k8s cluster and still want to use it. In this case, the cluster operator may register the cluster thirdparty object manually for the users, because the cluster operator is not inclined to give a clusterrolebinding to the operator, which can impact other controllers running inside the cluster that are also dependent on third party objects.

Let's not mix it in this PR. Please create another issue for it.

xiang90:
@uruddarraju Let's just add the none provider support in this PR. I want to keep PR/commit clean and independent. Thanks for you contribution and patience in advance.

hongchaodeng:
@uruddarraju 
Are you going to continue and address the comments? Otherwise, please close this PR and we will take care of the issue.

uruddarraju:
@hongchaodeng I am sorry, did not get a chance to send a revision of this PR. I'll try to get something in by end of this week, else I'll close this one.

hongchaodeng:
@uruddarraju 
Closing this one. PR stayed too long.

Issue #901, opened 2017-03-15
use FQDN for addressing etcd member pods


colhom:
Right now we're using pod IP, which means the only way to do any sort of x509 validation is via IP SANS. This makes assigning x509 identities to cluster pods a real pain, and needs to get fixed in advance of further static cluster TLS work.

  * Neither pod or service IP addresses are really reservable in Kubernetes. Which means you can't create the certificate until after the pod/service is created, which means you can't really pre-sign certificates in any way and still do meaningful metadata validation.

  * The certificate will not be transferable if that etcd node finds itself at a new IP

  * There's no way to sign a certificate for an open-ended group of addresses. So you have to account for all possible hostnames when you sign the cert, which again is problematic. Using DNS you can sign a cert for `*.xxxx.svc.cluster.local`, which is in plain english "any service address in the xxxx kubernetes namespace". This is *really critical* stuff to be able to express to `etcd-operator` from the get-go.

To get past these problems, we need to make sure `etcd-operator`, `backup` and all the etcd member pods *only* address each other via FQDN. For example:

`https://etcd-cluster-0002.my-cluster.my-namespace.svc.cluster.local`

This provides some much-needed flexibility to the end user, such as the ability to sign a generic certificate for all etcd member nodes in a reasonable way. Using IP SANS, this wouldn't really be possible as we would have to know the member IP addresses ahead of time.

A wildcard certificate could be generated for `*.my-cluster.my-namespace.svc.cluster.local`, but that wildcard is restricted to only validate for etcd member endpoints running in `my-namespace` on a Kubernetes cluster.

We don't even need to add a service per etcd member pod, as it appears we can directly expose the pod IP via DNS A records with `pod.beta.kubernetes.io/hostname` and `pod.beta.kubernetes.io/subdomain` annotations. 

\cc @hongchaodeng @ericchiang

hongchaodeng:
How can we create a service like "etcd-cluster-0002.my-cluster"?

colhom:
@hongchaodeng https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#a-records-and-hostname-based-on-pods-hostname-and-subdomain-fields

ericchiang:
@hongchaodeng This would require the etcd operator to generate a unique service for each member (don't know if you could use endpoints instead). Per conversation in #899 I think creating predictable names for the members is a requirement for static TLS. How else would the admin know what common name or SANs to supply?

hongchaodeng:
[...]

PR #935, opened 2017-04-04
cluster: use fqdn as url

xiang90:
fix https://github.com/coreos/etcd-operator/issues/901

### DIFF
--- a/pkg/util/etcdutil/member.go
+++ b/pkg/util/etcdutil/member.go
@@ -15,13 +15,18 @@
 package etcdutil
 
 import (
+	"errors"
 	"fmt"
+	"net/url"
+	"regexp"
 	"strconv"
 	"strings"
 )
 
 type Member struct {
 	Name string
+	// Kubernetes namespace this member runs in.
+	Namespace string
 	// ID field can be 0, which is unknown ID.
 	// We know the ID of a member when we get the member information from etcd,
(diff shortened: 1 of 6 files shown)


xiang90:
@colhom 

probably we should just remove the member peer service thing and use https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/ instead?

colhom:
@xiang90 I would agree- the peer FQDN only lives as long as the pod, so there is no need technically for the service abstraction for the peer interface.

xiang90:
@hongchaodeng talked with @colhom. we will first get this simple pr merged, so we use domain names. then we switch to use pod sub domain name to get rid of the service domain name.

pkg/util/etcdutil/member.go
```diff
@@ -140,6 +148,20 @@ func GetCounterFromMemberName(name string) (int, error) {
 	return c, nil
 }
 
+func MemberNameFromPeerURL(pu string) (string, error) {
+	// url.Parse has very loose validation. We do our own validation.
+	if !validPeerURL.MatchString(pu) {
```
hongchaodeng: Also move `validPeerURL` to this file?

xiang90: fixed.

PR #937, opened 2017-04-04
*: use pod A record as fqdn for better grouping

xiang90:
With cluster name as subdomain grouping, we can do TLS
SAN wildcard more accurately.

/cc @colhom @hongchaodeng

### DIFF
--- a/pkg/util/etcdutil/member.go
+++ b/pkg/util/etcdutil/member.go
@@ -39,7 +39,7 @@ type Member struct {
 }
 
 func (m *Member) fqdn() string {
-	return fmt.Sprintf("%s.%s.svc.cluster.local", m.Name, m.Namespace)
+	return fmt.Sprintf("%s.%s.%s.pod.cluster.local", m.Name, clusterNameFromMemberName(m.Name), m.Namespace)
 }
 
 func (m *Member) ClientAddr() string {
@@ -167,4 +167,12 @@ func MemberNameFromPeerURL(pu string) (string, error) {
 
 func CreateMemberName(clusterName string, member int) string {
 	return fmt.Sprintf("%s-%04d", clusterName, member)
+}
+
+func clusterNameFromMemberName(mn string) string {
+	i := strings.LastIndex(mn, "-")
+	if i == -1 {
+		return mn
+	}
+	return mn[:i]
 }
--- a/pkg/util/etcdutil/member.go
+++ b/pkg/util/etcdutil/member.go
@@ -39,7 +39,7 @@ type Member struct {
 }
 
 func (m *Member) fqdn() string {
-	return fmt.Sprintf("%s.%s.%s.pod.cluster.local", m.Name, clusterNameFromMemberName(m.Name), m.Namespace)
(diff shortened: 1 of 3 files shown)


xiang90:
@hongchaodeng we can still leave the etcd service around... it has a better looking that this long stuff... @colhom  what do you think?

pkg/util/k8sutil/k8sutil.go
```diff
@@ -266,6 +266,11 @@ func NewEtcdPod(m *etcdutil.Member, initialCluster []string, clusterName, state,
 			Volumes: []v1.Volume{
 				{Name: "etcd-data", VolumeSource: v1.VolumeSource{EmptyDir: &v1.EmptyDirVolumeSource{}}},
 			},
+			// DNS A record: [m.Name].[clusterName].Namespace.pod.cluster.local.
+			// For example, etcd-0000 in default namesapce will have DNS name
+			// `etcd-0000.etcd.default.pod.cluster.local`.
```
colhom: From the docs, I believe that the url will be `etcd-0000.etcd.default.svc.cluster.local` even though the A record is a pod IP.

colhom:
@xiang90 what's the reasoning for leaving the etcd service around?

pkg/util/k8sutil/k8sutil.go
xiang90: @colhom see `1-2-3-4.default.pod.cluster.local`. this is the example in the official doc. i tried on my local cluster, it works in this way.

xiang90:
@colhom It is just a shortcut for non tls clusters.

pkg/util/k8sutil/k8sutil.go
colhom: Also I'd like to note that the DNS search for the pods in default NS includes `default.svc.cluster.local`, so you should still be able to lookup address of the member pod via `etcd-0000.etcd` from within the namespace. I don't think there is any reason to leave the service around.

xiang90: ok. i will remove the service in another pr.

pkg/util/etcdutil/member.go
```diff
@@ -168,3 +168,11 @@ func MemberNameFromPeerURL(pu string) (string, error) {
 func CreateMemberName(clusterName string, member int) string {
 	return fmt.Sprintf("%s-%04d", clusterName, member)
 }
+
+func clusterNameFromMemberName(mn string) string {
+	i := strings.LastIndex(mn, "-")
+	if i == -1 {
+		return mn
```
hongchaodeng: panic?
This is unexpected.

xiang90: yea. let's me make the ci green first.

pkg/util/k8sutil/k8sutil.go
hasbro17: @xiang90 We just verified this on k8s 1.6:
```
Given a Pod with the hostname set to “busybox-1” and the
subdomain set to “default-subdomain”, and a headless Service
named “default-subdomain” in the same namespace, the pod will
see it’s own FQDN as 
“busybox-1.default-subdomain.my-namespace.svc.cluster.local”.
```
So we can keep that service around and access the pod by `etcd-0000.etcd.default.svc.cluster.local`.

Issue #691, opened 2017-01-28
Provide storage class options in cluster backup spec


hongchaodeng:
Currently, we only allow GCE, AWS as input and set up storage class for users. This is simple and cloud based for most users, especially new ones.

For the future, if there are user requests, we will add storage class field in backup spec. This issue to track it.

EugeneKha:
We use glasterfs as persistence provider in Kubernetes. To make operator save backups to PV created by glusterfs we manually created storageClass named 'etcd-backup-gce-pd' that use glusterfs provider. Now we can use '--pv-provisioner=kubernetes.io/gce-pd' and PV is created in glusterfs cluster and operator works all right.

xiang90:
implemented by #1358. Thanks @jamiehannaford!

PR #949, opened 2017-04-07
Enable StorageClass as option

scheeles:
Now it possible to provide an own StorageClass with more detail options as an alternative to the PV-Provisioner

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -51,6 +51,7 @@ import (
 var (
 	analyticsEnabled bool
 	pvProvisioner    string
+	storageClass     string
 	namespace        string
 	name             string
 	awsSecret        string
@@ -74,6 +75,7 @@ func init() {
 	flag.BoolVar(&analyticsEnabled, "analytics", true, "Send analytical event (Cluster Created/Deleted etc.) to Google Analytics")
 
 	flag.StringVar(&pvProvisioner, "pv-provisioner", constants.PVProvisionerGCEPD, "persistent volume provisioner type")
+	flag.StringVar(&storageClass, "storageclass", "", "Storage class type")
 	flag.StringVar(&awsSecret, "backup-aws-secret", "",
 		"The name of the kube secret object that stores the AWS credential file. The file name must be 'credentials'.")
 	flag.StringVar(&awsConfig, "backup-aws-config", "",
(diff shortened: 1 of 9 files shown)


hongchaodeng:
@scheeles 

What is the purpose of adding storage class option? Any use case?

scheeles:
We want to have more control over the storage. On AWS e.g. we want to use `encrypted` and control the `type`. Additional we use the etcdOperator on Glusterfs and have backup in place.

hongchaodeng:
@scheeles Can you create an issue to discuss your use case first?

I think this feature needs some additional thoughts:
- It's duplicate to have both "pv-provisioner" and "storageclass" options.
- See https://github.com/coreos/etcd-operator/issues/691 . Even if we provide storage class, is it gonna be per cluster level? This is more flexible.
- See https://github.com/coreos/etcd-operator/issues/691#issuecomment-288184323 . There is a hack on naming "etcd-backup-gce-pd" to your custom storage class currently.

cmd/operator/main.go
```diff
@@ -68,6 +69,7 @@ func init() {
 	flag.BoolVar(&analyticsEnabled, "analytics", true, "Send analytical event (Cluster Created/Deleted etc.) to Google Analytics")
 
 	flag.StringVar(&pvProvisioner, "pv-provisioner", constants.PVProvisionerGCEPD, "persistent volume provisioner type")
+	flag.StringVar(&storageClass, "storageclass", "", "Storage class type")
```
xiang90: operator probably should die with error message when both storage class and pv-provisioner are provided.

also this flag needs a better help message. storage class type is too generic.

xiang90: the pv-provisioner flag message is also bad :(

pkg/cluster/backupstorage/pv.go
```diff
@@ -11,23 +11,25 @@ type pv struct {
 	clusterName   string
 	namespace     string
 	pvProvisioner string
+	storageClass  string
```
xiang90: you can remove the pvProvisioner string actually. i think we actually convert the provisioner to storageclass anyway

jamiehannaford:
@scheeles Are you still working on this? If not, I can cherry pick your work and fix the final details.

I'd like to add Cinder as a new PV option, so this PR would supersede my one (https://github.com/coreos/etcd-operator/pull/1351).

/cc @hongchaodeng

hongchaodeng:
@jamiehannaford 
This issue has been inactive for a while. Let's just close it and let you take over it.
Note that this is still changing global flag, not adding option in cluster spec.
Thanks for taking care of it!

PR #954, opened 2017-04-11
gc: tail 100 lines of pod logs before collections

colhom:
useful for debugging. we may want to have the pod logs print on the debug level?

\cc @hongchaodeng


hongchaodeng:
We shouldn't print pods' logs in operator log.
pods' logs should be managed by cluster level, e.g. using kubectl, ElasticSearch, etc.

PR #958, opened 2017-04-11
op_guide: highlight rbac setup

hasbro17:
We need to highlight the rbac setup part just before creating the operator since it's easy to skim past the RBAC docs link in the overview.
A few users have already faced this problem and with rbac being default in k8s 1.6 this will become more common.

[skip ci]

/cc @hongchaodeng @xiang90

### DIFF
--- a/doc/user/op_guide.md
+++ b/doc/user/op_guide.md
@@ -2,6 +2,8 @@
 
 ## Install etcd operator
 
+Before you create the etcd-operator make sure you setup the necessary [rbac rules](./rbac.md) if your Kubernetes version is 1.6 or higher.
+
 Create deployment:
 
 ```bash


doc/user/op_guide.md
```diff
@@ -2,6 +2,8 @@
 
 ## Install etcd operator
 
+Before you create the etcd-operator make sure you setup the necessary [rbac rules](https://github.com/coreos/etcd-operator/blob/a57d872e3e1a40b9250b1ca90f90ac57e50fd945/doc/user/rbac.md) if your Kubernetes version is 1.6 or higher.
```
hongchaodeng: use `[xxx](./rbac.md)`

hongchaodeng:
lgtm

PR #961, opened 2017-04-13
cmd: add self-signed TLS generation script

colhom:
depends on #960

### DIFF
--- a/cmd/operator-gen-selfsigned-tls/main.go
+++ b/cmd/operator-gen-selfsigned-tls/main.go
@@ -42,8 +42,7 @@ var (
 	//Certificate Authority
 	peerCACfg tlsutil.CACertConfig
 	//Node Server Certificates
-	peerCfg tlsutil.CertConfig
-
+	peerCfg                          tlsutil.CertConfig
 	peerCertDuration, peerCADuration time.Duration
 	prettyPrint                      bool
 	outputDir                        string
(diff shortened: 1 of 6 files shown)


hongchaodeng:
This is just one method of generate TLS certs. Ideally, we should put them in separate repo, have docs there, and link it here. Why put it in operator codebase?

Another question is that there is already a docs: https://coreos.com/os/docs/latest/generate-self-signed-certificates.html . Why write a new one instead of linking to it and write some automatic script around it?

colhom:
> Ideally, we should put them in separate repo, have docs there, and link it here. Why put it in operator codebase?

`pkg/util/tlsutil`: moving past static TLS as the only option, eventually etcd operator should have the ability to sign unique x509 certificates for member nodes on demand, using either remote signing API or via local access to `ca-key.pem`. In that case having `util/tlsutil` in the operator code base would be a first step towards that goal.

As for `cmd/operator-gen-selfsigned-tls`- we could definitely move that to `examples/` or `contrib/` folder.

> Another question is that there is already a docs: https://coreos.com/os/docs/latest/generate-self-signed-certificates.html . Why write a new one instead of linking to it and write some automatic script around it?

`./cmd/operator-gen-selfsigned-tls`: i see this as the "reference implementation for static TLS on etcd operator". Providing instructions for generating TLS certificates via `cfssl` is definitely an option, but it would require the user understand how their own cluster TPR translates to correct x509 metadata. Given that static TLS is the "simple" option, I feel we should have a targeted reference/quickstart script like this somewhere in the repo.

An idea I had about this- perhaps we could make the `operator-gen-selfsigned-tls` script parse the cluster TPR and generate the appropriate static TLS secrets? I feel this would make life easier for a lot of people that want to try etcd-operator TLS w/ minimal effort.

hongchaodeng:
After a second thought, I think this is great to have something that helps users to automatically generate secrets needed and passed them to cluster spec. For example, we could have a tool to run the following workflow:
```
$ ./gen-tls-secrets $NAMESPACE $CLUSTER_NAME
...
secret "etcd-server-peer-tls" created, pass it to "peerSecret".
```

However, the current code in `cmd/operator-gen-selfsigned-tls` is way too powerful -- it writes a bunch of go code to create CA certs, peer certs, etc. We shouldn't maintain something like this in operator repo. We should rely on `cfssl`, which is more stable.

The `gen-tls-secrets` could be a wrapper around `cfssl`, `kubectl`, etc.

Issue #966, opened 2017-04-19
Use Patch() when upgrading etcd pod


hongchaodeng:
Currently we use Update() when upgrading etcd pod:
```Go
_, err = c.config.KubeCli.CoreV1().Pods(ns).Update(pod)
```

We should change it to Patch().

xiang90:
@hongchaodeng i am going to take care of this.

PR #972, opened 2017-04-20
cluster: use patch instead of update to patch a pod object

xiang90:
fix https://github.com/coreos/etcd-operator/issues/966

/cc @hongchaodeng @aaronlevy

### DIFF
--- a/pkg/util/k8sutil/k8sutil.go
+++ b/pkg/util/k8sutil/k8sutil.go
@@ -35,11 +35,11 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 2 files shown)


pkg/util/k8sutil/k8sutil.go
```diff
@@ -338,3 +338,29 @@ func newLablesForCluster(clusterName string) map[string]string {
...
+		return nil, err
+	}
+	newData, err := json.Marshal(n)
+	if err != nil {
+		return nil, err
+	}
+	return strategicpatch.CreateTwoWayMergePatch(oldData, newData, datastruct)
+}
+
+func ClonePod(p *v1.Pod) *v1.Pod {
```
aaronlevy: FWIW there is an `api.Scheme.DeepCopy` - but this is fine too.

xiang90: fixed.

xiang90:
merging.

PR #991, opened 2017-04-24
hack/test: add unused and gosimple checks

### LINKED ISSUE #876: Add `unused` and `gosimple` check in CI
ref:
- https://github.com/dominikh/go-tools/tree/master/cmd/unused
- https://github.com/dominikh/go-tools/tree/master/cmd/gosimple

We can add the above tools to `unit` CI

hasbro17:
Added checks for `unused` and `gosimple`.
Fixes #876 

Also corrected changes pointed out by the above checks:
```
Checking unused...
unused checking failed:
pkg/cluster/cluster.go:85:2: field backupDir is unused (U1000)
pkg/cluster/cluster.go:390:19: func (*Cluster).deleteClientServiceLB is unused (U1000)
pkg/util/k8sutil/k8sutil.go:51:2: const annotationPrometheusScrape is unused (U1000)
pkg/util/k8sutil/k8sutil.go:52:2: const annotationPrometheusPort is unused (U1000)
Checking gosimple...
gosimple checking failed:
pkg/backup/s3/s3.go:73:2: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013)
pkg/backup/s3/s3.go:98:2: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013)
pkg/cluster/cluster.go:417:2: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013)
test/e2e/framework/framework.go:138:2: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013)
```
Waiting for tests to pass.

### DIFF
--- a/hack/ci/run
+++ b/hack/ci/run
@@ -14,6 +14,7 @@ mkdir -p $GOPATH
 cd $GOPATH
 export GOPATH=`pwd`
 echo "GOPATH: ${GOPATH}"
+export PATH=$GOPATH/bin:$PATH
 
 mkdir -p $GOPATH/src/github.com/coreos
 ln -s "${origpwd}" $GOPATH/src/github.com/coreos/etcd-operator
--- a/hack/test
+++ b/hack/test
@@ -29,7 +29,6 @@ function fmt_pass {
 		fi
 	done
 
-
 	echo "Checking govet..."
 	for file in $allfiles; do
 		checkRes=$(go vet $file)
@@ -38,6 +37,30 @@ function fmt_pass {
 			exit 255
 		fi
 	done
+
+	go get honnef.co/go/tools/cmd/gosimple || true
+	if which gosimple >/dev/null; then
+		echo "Checking gosimple..."
+		checkRes=$(gosimple $(go list ./client/... ./cmd/... ./pkg/... ./test/e2e/...)) || true
+			if [ -n "${checkRes}" ]; then
+				echo -e "gosimple checking failed:\n${checkRes}"
+				exit 255
+			fi
+	else
+		echo "Skipping gosimple: failed to install"
+	fi
+
+	go get honnef.co/go/tools/cmd/unused || true
+	if which unused >/dev/null; then
+		echo "Checking unused..."
+		checkRes=$(unused $(go list ./client/... ./cmd/... ./pkg/... ./test/e2e/...)) || true
+		if [ -n "${checkRes}" ]; then
+				echo -e "unused checking failed:\n${checkRes}"
+				exit 255
+		fi
+	else
+		echo "Skipping unused: failed to install"
+	fi
 }
 
 function build_pass {
--- a/pkg/backup/s3/s3.go
+++ b/pkg/backup/s3/s3.go
@@ -70,11 +70,7 @@ func (s *S3) Put(key string, rs io.ReadSeeker) error {
 		Body:   rs,
 	})
 
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return err
 }
 
 func (s *S3) Get(key string) (io.ReadCloser, error) {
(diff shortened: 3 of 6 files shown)


hack/test
```diff
@@ -5,7 +5,7 @@ set -o nounset
 set -o pipefail
 
 if [ -z "${PASSES-}" ]; then
-	PASSES="fmt build e2e unit cliente2e"
+	PASSES="fmt unused simple build e2e unit cliente2e"
```
hongchaodeng: Let's just add "unused" and "simple" check in "fmt"

hack/test
```diff
@@ -40,6 +40,26 @@ function fmt_pass {
 	done
 }
 
+function simple_pass {
+	echo "Checking gosimple..."
+	go get honnef.co/go/tools/cmd/gosimple
```
xiang90: if go get fails, we should probably ignore the tests (go get from honnef fails pretty frequently. i do not really want this to block our entire CI). alternatively, we do not go get this every time.

hack/test
```diff
@@ -40,6 +40,26 @@ function fmt_pass {
...
+	checkRes=$(gosimple $(go list ./pkg/... ./test/e2e/...)) || true
+		if [ -n "${checkRes}" ]; then
+			echo -e "gosimple checking failed:\n${checkRes}"
```
xiang90: do not fail if gosimple does not exist?

xiang90:
lgtm

hack/test
hongchaodeng: @hasbro17 
Check out etcd's example: https://github.com/coreos/etcd/blob/master/test#L252-L267

hack/test
```diff
@@ -38,6 +37,30 @@ function fmt_pass {
...
+	go get honnef.co/go/tools/cmd/gosimple || true
+	if which gosimple >/dev/null; then
+		echo "Checking gosimple..."
+		checkRes=$(gosimple $(go list ./pkg/... ./test/e2e/...)) || true
+			if [ -n "${checkRes}" ]; then
+				echo -e "gosimple checking failed:\n${checkRes}"
+				exit 255
+			fi
+	else
+		echo "Skipping gosimple..."
```
xiang90: Skipping gosimple: failed to install.

hack/test
```diff
@@ -38,6 +37,30 @@ function fmt_pass {
...
+	go get honnef.co/go/tools/cmd/unused || true
+	if which unused >/dev/null; then
+		echo "Checking unused..."
+		checkRes=$(unused $(go list ./pkg/... ./test/e2e/...)) || true
+		if [ -n "${checkRes}" ]; then
+				echo -e "unused checking failed:\n${checkRes}"
+				exit 255
+		fi
+	else
+		echo "Skipping unused..."
```
xiang90: Skipping unused: failed to install

hongchaodeng:
The check will always fail-skip. Please do
```
export PATH=$GOPATH/bin:$PATH
```
in `hack/ci/run`

hasbro17:
@etcd-bot retest this please

hasbro17:
@hongchaodeng tests are passing PTAL

hack/test
```diff
@@ -85,6 +108,8 @@ function cliente2e_pass {
 	./client/cliente2e/test
 }
 
+export PATH=$GOPATH/bin:$PATH
```
hongchaodeng: Please put this into `hack/ci/run` after exporting GOPATH

hack/test
```diff
@@ -38,6 +37,30 @@ function fmt_pass {
 			exit 255
 		fi
 	done
+
+	go get honnef.co/go/tools/cmd/gosimple || true
+	if which gosimple >/dev/null; then
+		echo "Checking gosimple..."
+		checkRes=$(gosimple $(go list ./pkg/... ./test/e2e/...)) || true
```
hongchaodeng: Why not checking `cmd/`?

hack/ci/run
```diff
@@ -14,6 +14,8 @@ mkdir -p $GOPATH
 cd $GOPATH
 export GOPATH=`pwd`
 echo "GOPATH: ${GOPATH}"
+export PATH=$GOPATH/bin:$PATH
+echo "PATH: ${PATH}"
```
hongchaodeng: PATH isn't useful to us. Let's not echo it.

Issue #978, opened 2017-04-21
e2e flake: wrong members


hongchaodeng:
failure:
```
        --- FAIL: TestSelfHosted/self_hosted/create_self_hosted_cluster_from_scratch (230.28s)
        	util.go:388: 2017-04-21 08:16:58.063161347 +0000 UTC created etcd cluster: test-etcd-zckv5
        	util.go:388: 2017-04-21 08:16:58.106539169 +0000 UTC waiting size (3), etcd pods: names ([]), nodes ([])
        	util.go:388: 2017-04-21 08:17:08.105124844 +0000 UTC waiting size (3), etcd pods: names ([test-etcd-zckv5-0000]), nodes ([gke-jenkins-default-pool-427b821a-c6xz])
        	util.go:388: 2017-04-21 08:17:18.12774646 +0000 UTC waiting size (3), etcd pods: names ([]), nodes ([])
        	util.go:388: 2017-04-21 08:17:28.104476363 +0000 UTC waiting size (3), etcd pods: names ([]), nodes ([])
        	util.go:388: 2017-04-21 08:17:38.102337214 +0000 UTC waiting size (3), etcd pods: names ([]), nodes ([])
        	util.go:388: 2017-04-21 08:17:48.102387109 +0000 UTC waiting size (3), etcd pods: names ([]), nodes ([])
        	util.go:388: 2017-04-21 08:17:58.101985103 +0000 UTC waiting size (3), etcd pods: names ([]), nodes ([])
        	util.go:388: 2017-04-21 08:18:08.102403378 +0000 UTC waiting size (3), etcd pods: names ([]), nodes ([])
        	util.go:388: 2017-04-21 08:18:18.102634182 +0000 UTC waiting size (3), etcd pods: names ([]), nodes ([])
        	util.go:388: 2017-04-21 08:18:28.102098682 +0000 UTC waiting size (3), etcd pods: names ([]), nodes ([])
        	util.go:388: 2017-04-21 08:18:38.102289623 +0000 UTC waiting size (3), etcd pods: names ([]), nodes ([])
        	util.go:388: 2017-04-21 08:18:48.10288512 +0000 UTC waiting size (3), etcd pods: names ([]), nodes ([])
        	util.go:388: 2017-04-21 08:18:58.103315281 +0000 UTC waiting size (3), etcd pods: names ([]), nodes ([])
[...]

PR #997, opened 2017-04-25
cluster: detect bad member in self hosted case

xiang90:
this should help debugging https://github.com/coreos/etcd-operator/issues/978

### DIFF
--- a/pkg/cluster/member.go
+++ b/pkg/cluster/member.go
@@ -16,6 +16,7 @@ package cluster
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/coreos/etcd-operator/pkg/spec"
 	"github.com/coreos/etcd-operator/pkg/util/etcdutil"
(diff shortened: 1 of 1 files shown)


hongchaodeng:
@etcd-bot retest this please

hongchaodeng:
@etcd-bot retest this please

known issue: https://github.com/coreos/etcd-operator/issues/978

hongchaodeng:
Issue #978 happens quite often.
Let's wait https://github.com/coreos/etcd-operator/pull/999 to merge and then retest this.

hongchaodeng:
@xiang90 
Can you rebase?

xiang90:
@hongchaodeng rebased.

pkg/cluster/member.go
```diff
@@ -38,6 +39,11 @@ func (c *Cluster) updateMembers(known etcdutil.MemberSet) error {
 				return errUnexpectedUnreadyMember
 			}
 
+			if !strings.HasPrefix(m.Name, c.cluster.Metadata.GetName()) {
```
hongchaodeng: This is not right.
The non-managed member could have a name not prefixed with cluster name.

xiang90: right. we need to compare the client url against the seed url. if it is one of the seed, we need to ignore it.

hongchaodeng: We actually just compare it below this.
Please put this check after that.

xiang90: fixed.

hongchaodeng:
lgtm

Issue #914, opened 2017-03-17
Option to set auto compaction retention policy


snarlysodboxer:
We would like to be able to set [this setting](https://github.com/coreos/etcd/blob/master/Documentation/op-guide/configuration.md#--auto-compaction-retention), otherwise we have to run jobs to periodically do compaction.

There are a couple others that would be nice to be able to set as well, like [v2](https://github.com/coreos/etcd/blob/master/Documentation/op-guide/configuration.md#--enable-v2), [cors](https://github.com/coreos/etcd/blob/master/Documentation/op-guide/configuration.md#--cors), and others.

xiang90:
Yes. This is something we will address soon.

xiang90:
@snarlysodboxer do you want to add this feature?

we need to:

1. introduce etcdPolicy into the cluster spec
2. add auto compaction as the first parameter

PR #1001, opened 2017-04-25
spec: support user defined evn

xiang90:
fix https://github.com/coreos/etcd-operator/issues/914.

/cc @hongchaodeng


hongchaodeng:
lgtm

PR #1002, opened 2017-04-25
k8sutil: fix env overwrite in self hosted cluster

xiang90:


pkg/util/k8sutil/self_hosted.go
```diff
@@ -77,9 +77,9 @@ func NewSelfHostedEtcdPod(name string, initialCluster []string, clusterName, ns,
 		commands = fmt.Sprintf("%s --initial-cluster-token=%s", commands, token)
 	}
 
-	var env []v1.EnvVar
+	env := []v1.EnvVar{envPodIP}
```
hongchaodeng: Why not do the same thing to normal cluster?

xiang90: we do not need to get podIP in normal cluster. so we never pass in env before.

hongchaodeng:
lgtm

Issue #872, opened 2017-03-08
Support custom labels in pod template


davidquarles:
This is piggybacking on #552, but [our use case](https://github.com/coreos/etcd-operator/issues/552#issuecomment-284895991) doesn't require direct access to the raw pod spec/template.

PR #1006, opened 2017-04-26
*: support customized labels

xiang90:
fix https://github.com/coreos/etcd-operator/issues/872

/cc @hongchaodeng @hasbro17


hongchaodeng:
lgtm

PR #1015, opened 2017-04-27
*: s3 backup path should include its namesapce

xiang90:


hongchaodeng:
lgtm

I think this is a good start. Let's use this to write an upgrade test.

PR #1023, opened 2017-04-28
e2e: safe update TPR instead of unconditional update

hasbro17:
Unconditional update will overwrite TPR and we should not use it.


hongchaodeng:
Commit title and comment should tell why you do it, not what you have done.

Change commit title:
```
e2e: safe update TPR instead of unconditional update

Unconditional update will overwrite TPR and we should not use it.
```

hongchaodeng:
Also remove the unconditionalUpdate() method in k8sutil

hongchaodeng:
I'm surprised that no one is using `UpdateClusterTPRObjectUnconditionally` but `unused` didn't fail it...

PR #1026, opened 2017-04-29
tpr_util,e2e/util: AtomicUpdateClusterTPR to retry safe update

hasbro17:
`AtomicUpdateClusterTPR` avoids overwriting the cluster TPR by fetching the latest cluster TPR and applying an update function to modify it before attempting to update the cluster TPR object.
 
This could fail if our fetched TPR becomes outdated before we can modify and update it. 
So the whole process is retried with exponential backoff for `Step` number of times until the update is successful, else we return an error.

Also changed is `updateEtcdCluster` in `e2e/util.go` which now accepts an update function in order to use `AtomicUpdateClusterTPR`.

@hongchaodeng I have not modified upgrade_test.go to use `AtomicUpdateClusterTPR` though it should be a similar change once it starts using `e2e/util.go`

### DIFF
--- a/pkg/util/k8sutil/tpr_util.go
+++ b/pkg/util/k8sutil/tpr_util.go
@@ -16,7 +16,6 @@ package k8sutil
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"net/http"
 	"time"
@@ -29,6 +28,10 @@ import (
 )
 
 // TODO: replace this package with Operator client
+
+// ClusterTPRUpdateFunc is a function to be used when atomically
+// updating a Cluster TPR.
+type ClusterTPRUpdateFunc func(*spec.Cluster)
 
 func WatchClusters(host, ns string, httpClient *http.Client, resourceVersion string) (*http.Response, error) {
 	return httpClient.Get(fmt.Sprintf("%s/apis/%s/%s/namespaces/%s/clusters?watch=true&resourceVersion=%s",
@@ -74,16 +77,32 @@ func GetClusterTPRObject(restcli rest.Interface, ns, name string) (*spec.Cluster
 	return readOutCluster(b)
 }
 
-// UpdateClusterTPRObject updates the given TPR object.
-// ResourceVersion of the object MUST be set or update will fail.
-func UpdateClusterTPRObject(restcli rest.Interface, ns string, c *spec.Cluster) (*spec.Cluster, error) {
-	if len(c.Metadata.ResourceVersion) == 0 {
-		return nil, errors.New("k8sutil: resource version is not provided")
-	}
-	return updateClusterTPRObject(restcli, ns, c)
+// AtomicUpdateClusterTPRObject will get the latest result of a cluster,
+// let user modify it, and update the cluster with modified result
+// The entire process would be retried if there is a conflict of resource version
+func AtomicUpdateClusterTPRObject(restcli rest.Interface, name, namespace string, maxRetries int, updateFunc ClusterTPRUpdateFunc) (*spec.Cluster, error) {
+	var updatedCluster *spec.Cluster
+	err := retryutil.Retry(1*time.Second, maxRetries, func() (done bool, err error) {
+		currCluster, err := GetClusterTPRObject(restcli, namespace, name)
+		if err != nil {
+			return false, err
+		}
+
+		updateFunc(currCluster)
+
+		updatedCluster, err = UpdateClusterTPRObject(restcli, namespace, currCluster)
+		if err != nil {
+			if apierrors.IsConflict(err) {
+				return false, nil
+			}
+			return false, err
+		}
+		return true, nil
+	})
+	return updatedCluster, err
 }
 
-func updateClusterTPRObject(restcli rest.Interface, ns string, c *spec.Cluster) (*spec.Cluster, error) {
+func UpdateClusterTPRObject(restcli rest.Interface, ns string, c *spec.Cluster) (*spec.Cluster, error) {
 	uri := fmt.Sprintf("/apis/%s/%s/namespaces/%s/clusters/%s", spec.TPRGroup, spec.TPRVersion, ns, c.Metadata.Name)
 	b, err := restcli.Put().RequestURI(uri).Body(c).DoRaw()
 	if err != nil {
--- a/test/e2e/upgradetest/upgrade_test.go
+++ b/test/e2e/upgradetest/upgrade_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/coreos/etcd-operator/pkg/spec"
 	"github.com/coreos/etcd-operator/pkg/util/k8sutil"
 	"github.com/coreos/etcd-operator/pkg/util/retryutil"
+	"github.com/coreos/etcd-operator/test/e2e/e2eutil"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/pkg/api/v1"
@@ -66,9 +67,11 @@ func TestResize(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	testClus.Spec.Size = 5
-	_, err = updateEtcdCluster(testClus)
-	if err != nil {
+
+	updateFunc := func(cl *spec.Cluster) {
+		cl.Spec.Size = 5
+	}
+	if _, err := e2eutil.UpdateEtcdCluster(testF.KubeCli, testClus, 10, updateFunc); err != nil {
 		t.Fatal(err)
 	}
 	_, err = waitUntilSizeReached(testClus.Metadata.Name, 5, 60*time.Second)
@@ -100,10 +103,6 @@ func createCluster() (*spec.Cluster, error) {
 		return nil, err
 	}
 	return res, nil
-}
-
-func updateEtcdCluster(cl *spec.Cluster) (*spec.Cluster, error) {
-	return k8sutil.UpdateClusterTPRObject(testF.KubeCli.CoreV1().RESTClient(), testF.KubeNS, cl)
 }
 
 func deleteCluster() error {
--- a/test/e2e/util.go
+++ b/test/e2e/util.go
@@ -232,10 +232,6 @@ func createCluster(t *testing.T, f *framework.Framework, cl *spec.Cluster) (*spe
 	return res, nil
 }
 
-func updateEtcdCluster(f *framework.Framework, c *spec.Cluster) (*spec.Cluster, error) {
-	return k8sutil.UpdateClusterTPRObject(f.KubeClient.CoreV1().RESTClient(), f.Namespace, c)
-}
-
 func deleteEtcdCluster(t *testing.T, f *framework.Framework, c *spec.Cluster) error {
 	podList, err := f.KubeClient.CoreV1().Pods(f.Namespace).List(k8sutil.ClusterListOpt(c.Metadata.Name))
 	if err != nil {
--- a/test/e2e/basic_test.go
+++ b/test/e2e/basic_test.go
@@ -19,7 +19,8 @@ import (
 	"testing"
 	"time"
 
-	"github.com/coreos/etcd-operator/pkg/util/k8sutil"
+	"github.com/coreos/etcd-operator/pkg/spec"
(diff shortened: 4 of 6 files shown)


pkg/util/k8sutil/tpr_util.go
```diff
@@ -74,16 +79,36 @@ func GetClusterTPRObject(restcli rest.Interface, ns, name string) (*spec.Cluster
...
-	if len(c.Metadata.ResourceVersion) == 0 {
-		return nil, errors.New("k8sutil: resource version is not provided")
-	}
-	return updateClusterTPRObject(restcli, ns, c)
+// AtomicUpdateClusterTPRObject takes an update function which is executed before attempting
+// to update the Cluster TPR object. Upon conflict or some other error the entire process is retried
+// again upto Step times until the update is successful.
+func AtomicUpdateClusterTPRObject(restcli rest.Interface, name, namespace string, updateFunc ClusterTPRModifier) (*spec.Cluster, error) {
+	var updatedCluster *spec.Cluster
+	err := wait.ExponentialBackoff(wait.Backoff{
```
hongchaodeng: Please use our own `retryutil.Retry`

pkg/util/k8sutil/tpr_util.go
```diff
@@ -74,16 +79,36 @@ func GetClusterTPRObject(restcli rest.Interface, ns, name string) (*spec.Cluster
...
 
-// UpdateClusterTPRObject updates the given TPR object.
-// ResourceVersion of the object MUST be set or update will fail.
-func UpdateClusterTPRObject(restcli rest.Interface, ns string, c *spec.Cluster) (*spec.Cluster, error) {
-	if len(c.Metadata.ResourceVersion) == 0 {
-		return nil, errors.New("k8sutil: resource version is not provided")
-	}
-	return updateClusterTPRObject(restcli, ns, c)
+// AtomicUpdateClusterTPRObject takes an update function which is executed before attempting
+// to update the Cluster TPR object. Upon conflict or some other error the entire process is retried
```
hongchaodeng: > Upon conflict or some other error the entire process is retried

Let's retry only if there is a resource conflict. Any other error should return.

pkg/cluster/cluster.go
```diff
@@ -495,13 +495,14 @@ func (c *Cluster) updateTPRStatus() error {
...
 	}
 
-	newCluster := c.cluster
-	newCluster.Status = c.status
-	newCluster, err := k8sutil.UpdateClusterTPRObject(c.config.KubeCli.Core().RESTClient(), c.cluster.Metadata.Namespace, newCluster)
+	updateFunc := func(cl *spec.Cluster) {
+		cl.Spec = c.cluster.Spec
+		cl.Status = c.status
+	}
+	newCluster, err := k8sutil.AtomicUpdateClusterTPRObject(c.config.KubeCli.Core().RESTClient(), c.cluster.Metadata.Name, c.cluster.Metadata.Namespace, updateFunc)
```
hongchaodeng: `updateTPRStatus()` is being retried in another method, `reportFailedStatus()` IIRC.

We need further thoughts on how to change existing `UpdateClusterTPRObject` before changing it.

Let's make this PR change only the test files, and deal with other places later in separate PR.

hasbro17:
upgrade_test.go now uses `UpdateEtcdCluster` test/e2e in order to do `AtomicUpdateClusterTPR `.

pkg/cluster/cluster.go
```diff
@@ -501,7 +501,6 @@ func (c *Cluster) updateTPRStatus() error {
 	if err != nil {
 		return err
 	}
-
```
hongchaodeng: Not change this?

pkg/util/k8sutil/tpr_util.go
```diff
@@ -16,20 +16,24 @@ package k8sutil
...
+	"github.com/Sirupsen/logrus"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/rest"
 )
 
 // TODO: replace this package with Operator client
 
+// ClusterTPRModifier is a modifier function to be used when atomically
+// updating a Cluster TPR.
+type ClusterTPRModifier func(*spec.Cluster)
```
hongchaodeng: Modifier is usually an interface.
For a func, we should call it ModifyFunc or UpdateFUnc.

pkg/util/k8sutil/tpr_util.go
```diff
@@ -74,16 +78,33 @@ func GetClusterTPRObject(restcli rest.Interface, ns, name string) (*spec.Cluster
...
 
-// UpdateClusterTPRObject updates the given TPR object.
-// ResourceVersion of the object MUST be set or update will fail.
-func UpdateClusterTPRObject(restcli rest.Interface, ns string, c *spec.Cluster) (*spec.Cluster, error) {
-	if len(c.Metadata.ResourceVersion) == 0 {
-		return nil, errors.New("k8sutil: resource version is not provided")
-	}
-	return updateClusterTPRObject(restcli, ns, c)
+// AtomicUpdateClusterTPRObject takes an update function which is executed before attempting
+// to update the Cluster TPR object. The entire process is retried
```
hongchaodeng: AtomicUpdateClusterTPRObject will get the latest result of a cluster, let user modify it, and update the cluster with modified result. The entire process would be retried if there is a conflict of resource version.

pkg/util/k8sutil/tpr_util.go
```diff
@@ -74,16 +78,33 @@ func GetClusterTPRObject(restcli rest.Interface, ns, name string) (*spec.Cluster
...
+// again until the update is successful or a non-conflict error is returned.
+func AtomicUpdateClusterTPRObject(restcli rest.Interface, name, namespace string, updateFunc ClusterTPRModifier) (*spec.Cluster, error) {
+	var updatedCluster *spec.Cluster
+	err := retryutil.Retry(5*time.Second, 5, func() (done bool, err error) {
```
hongchaodeng: 5s interval seems too long. Maybe use 1s?
Why retry only 5 times?

pkg/util/k8sutil/tpr_util.go
```diff
@@ -74,16 +78,33 @@ func GetClusterTPRObject(restcli rest.Interface, ns, name string) (*spec.Cluster
...
-// ResourceVersion of the object MUST be set or update will fail.
-func UpdateClusterTPRObject(restcli rest.Interface, ns string, c *spec.Cluster) (*spec.Cluster, error) {
-	if len(c.Metadata.ResourceVersion) == 0 {
-		return nil, errors.New("k8sutil: resource version is not provided")
-	}
-	return updateClusterTPRObject(restcli, ns, c)
+// AtomicUpdateClusterTPRObject takes an update function which is executed before attempting
+// to update the Cluster TPR object. The entire process is retried
+// again until the update is successful or a non-conflict error is returned.
+func AtomicUpdateClusterTPRObject(restcli rest.Interface, name, namespace string, updateFunc ClusterTPRModifier) (*spec.Cluster, error) {
```
hongchaodeng: Please align the name of "updateFunc" and "ClusterTPRModifier".

test/e2e/util.go
```diff
@@ -232,8 +233,8 @@ func createCluster(t *testing.T, f *framework.Framework, cl *spec.Cluster) (*spe
 	return res, nil
 }
 
-func updateEtcdCluster(f *framework.Framework, c *spec.Cluster) (*spec.Cluster, error) {
-	return k8sutil.UpdateClusterTPRObject(f.KubeClient.CoreV1().RESTClient(), f.Namespace, c)
+func UpdateEtcdCluster(kubeClient kubernetes.Interface, cl *spec.Cluster, updateFunc k8sutil.ClusterTPRModifier) (*spec.Cluster, error) {
```
hongchaodeng: upgradetest should not import the "e2e/".
Maybe move this to "e2eutil/".

pkg/util/k8sutil/tpr_util.go
```diff
@@ -74,16 +78,33 @@ func GetClusterTPRObject(restcli rest.Interface, ns, name string) (*spec.Cluster
...
+		currCluster, err := GetClusterTPRObject(restcli, namespace, name)
+		if err != nil {
+			return false, err
+		}
+
+		updateFunc(currCluster)
+
+		updatedCluster, err = UpdateClusterTPRObject(restcli, namespace, currCluster)
+		if err != nil {
+			if apierrors.IsConflict(err) {
```
hongchaodeng: Is there any way we can test deterministically that a conflict will return this error?

pkg/util/k8sutil/tpr_util.go
hasbro17: No reason. 5 just seemed like a reasonable number though I can increase it to 10 or something. The point being we would only be retrying a limited number of times.

hongchaodeng: This is a generic util.
If retry times is case by case, we should make it a parameter.

pkg/util/k8sutil/tpr_util.go
hasbro17: The code path for apimachinery and client-go indicate that a conflict error should be returned but I can write a manual test with client-go to force a conflict by:
- Getting the latest TPR
- Updating the TPR
- Updating the same TPR again with the previous resource version.
- Check if the error is classified by `IsConflict()`

pkg/util/k8sutil/tpr_util.go
```diff
@@ -74,16 +78,33 @@ func GetClusterTPRObject(restcli rest.Interface, ns, name string) (*spec.Cluster
...
+				logrus.Warningf("conflict updating cluster TPR resource, will try again: %v", err)
```
hongchaodeng: Also, this is a generic util. Shouldn't have `logrus.Warningf` in between.

pkg/util/k8sutil/tpr_util.go
hongchaodeng: Right.
This is a key path we might want to test. Maybe put the test into "cliente2e/"?
But since you have also verified the code path, I'm fine to write the test PR separately to minimize each PR.

hasbro17:
@hongchaodeng `UpdateEtcdCluster` has been moved to `test/e2e/e2eutil` so that both `test/e2e` and `test/e2e/upgradetest` can use the common function. With more cleanup of `test/e2e/util.go` more common functions might be moved here.

pkg/util/k8sutil/tpr_util.go
```diff
@@ -74,16 +77,32 @@ func GetClusterTPRObject(restcli rest.Interface, ns, name string) (*spec.Cluster
...
-// ResourceVersion of the object MUST be set or update will fail.
-func UpdateClusterTPRObject(restcli rest.Interface, ns string, c *spec.Cluster) (*spec.Cluster, error) {
-	if len(c.Metadata.ResourceVersion) == 0 {
-		return nil, errors.New("k8sutil: resource version is not provided")
-	}
-	return updateClusterTPRObject(restcli, ns, c)
+// AtomicUpdateClusterTPRObject will get the latest result of a cluster,
+// let user modify it, and update the cluster with modified result
+// The entire process would be retried if there is a conflict of resource version
+func AtomicUpdateClusterTPRObject(restcli rest.Interface, name, namespace string, updateFunc ClusterTPRUpdateFunc, maxRetries int) (*spec.Cluster, error) {
```
hongchaodeng: Convention: put func as last parameter.

test/e2e/e2eutil/util.go
```diff
@@ -0,0 +1,12 @@
+package e2eutil
```
hongchaodeng: better rename file to `tpr_util.go`

test/e2e/upgradetest/upgrade_test.go
```diff
@@ -66,11 +67,14 @@ func TestResize(t *testing.T) {
...
-	_, err = updateEtcdCluster(testClus)
-	if err != nil {
+
+	updateFunc := func(cl *spec.Cluster) {
+		cl.Spec.Size = 5
+	}
+	if _, err := e2eutil.UpdateEtcdCluster(testF.KubeCli, testClus, updateFunc, 10); err != nil {
 		t.Fatal(err)
 	}
+
```
hongchaodeng: remove this extra line?

PR #1027, opened 2017-05-01
e2e/util.go: refactor CreateCluster() into e2e/e2eutil for common use

hasbro17:
Refactored the `CreateCluster()` function and moved it to `e2eutil/tpr_util.go` since it's a common function.

upgrade_test.go is also changed to use the `CreateCluster()` func.

/cc @hongchaodeng

### DIFF
--- a/test/e2e/e2eutil/tpr_util.go
+++ b/test/e2e/e2eutil/tpr_util.go
@@ -1,12 +1,35 @@
 package e2eutil
 
 import (
+	"encoding/json"
+	"fmt"
+	"testing"
+	"time"
+
 	"github.com/coreos/etcd-operator/pkg/spec"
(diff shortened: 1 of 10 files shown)


test/e2e/e2eutil/tpr_util.go
```diff
@@ -1,12 +1,35 @@
...
+	"testing"
+	"time"
+
 	"github.com/coreos/etcd-operator/pkg/spec"
 	"github.com/coreos/etcd-operator/pkg/util/k8sutil"
 
 	"k8s.io/client-go/kubernetes"
 )
 
+func CreateCluster(t *testing.T, kubeClient kubernetes.Interface, cl *spec.Cluster, namespace string) (*spec.Cluster, error) {
```
hongchaodeng: parameters: put `namespace` before `cluster`? Better put native type first

PR #1031, opened 2017-05-02
e2e/util.go: refactor DeleteCluster() for common use between different frameworks

hasbro17:
Moved `DeleteEtcdCluster()` into `e2eutil/tpr_util.go` and made it more generic by not passing the framework.
The helper functions for `DeleteEtcdCluster()`,  `waitResourcesDeleted()` and `waitBackupDeleted()` have been moved to `e2eutil/util.go`.

In order to make `waitBackupDeleted()` support checking of different types of storage options in the future, a new wrapper struct `StorageCheckerOptions` is passed to `DeleteEtcdCluster()` which is then type asserted to the correct type of storage option.

### DIFF
--- a/test/e2e/e2eutil/tpr_util.go
+++ b/test/e2e/e2eutil/tpr_util.go
@@ -1,3 +1,17 @@
+// Copyright 2017 The etcd-operator Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package e2eutil
 
 import (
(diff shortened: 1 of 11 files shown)


hasbro17:
@hongchaodeng PTAL to see if the `StorageCheckerOptions` is the right way to pass the S3 and similar storage checking parameters to `waitBackupDeleted()`.

Also have to rebase with master after merging #1030

test/e2e/e2eutil/tpr_util.go
```diff
@@ -9,9 +9,19 @@ import (
...
 	"k8s.io/client-go/kubernetes"
 )
 
+type S3Options struct {
+	S3Cli    *s3.S3
+	S3Bucket string
+}
+
+type StorageCheckerOptions struct {
+	Options interface{}
```
hongchaodeng: We don't need this. Just put all fields here.

test/e2e/e2eutil/wait_util.go
```diff
@@ -0,0 +1,144 @@
+package e2eutil
```
hongchaodeng: Header?

hongchaodeng: Maybe rename this file to "util.go"? We can put all methods that's not very specific into here.

hasbro17: We do have a another wait related method `waitUntilSizeReached` that will be refactored here soon. I just wanted to logically group all the wait and retry type methods together.
Or we can rename it if you think that's better.

test/e2e/e2eutil/wait_util.go
```diff
@@ -0,0 +1,158 @@
+// Copyright 2016 The etcd-operator Authors
```
hongchaodeng: 2016 -> 2017

hasbro17: I'll change the licenses here but we'll need to update everywhere else sometime later.

hongchaodeng: We only check for new files coming in.

PR #1036, opened 2017-05-02
*: add pod policy to backup pod

### LINKED ISSUE #1009: Support backup sidecar's own NodeSelector
A backup sidecar pod seems to use the NodeSelector of cluster's pod. So, When I set anti-affinity with etcd pods, one etcd pod couldn't be scheduled due to a backup sidecar pod.

It would be good to give a flexibility to set the backup sidecar's own NodeSelector.

xiang90:
Fix #1009


hongchaodeng:
lgtm

PR #1043, opened 2017-05-02
upgradetest/TestResize: wait for cluster TPR, remove unnecessary waits

hasbro17:
Check for the creation of the TPR, after creating operator deployment, by polling for the presence of the API path:
`/apis/etcd.coreos.com/v1beta1/namespaces/<namespace>/clusters/`

No need to wait after deleting the cluster since `DeleteCluster()` already does `waitResourcesDeleted()` and `waitBackupDeleted()`.

Also no need to wait after updating the operator deployment since we do `WaitUntilSizeReached()` after changing cluster size.

### DIFF
--- a/test/e2e/upgradetest/upgrade_test.go
+++ b/test/e2e/upgradetest/upgrade_test.go
@@ -34,7 +34,10 @@ func TestResize(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	time.Sleep(10 * time.Second)
+	if err = k8sutil.WaitEtcdTPRReady(testF.KubeCli.CoreV1().RESTClient(), 3*time.Second, 30*time.Second, testF.KubeNS); err != nil {
+		t.Fatalf("failed to see cluster TPR get created in time: %v", err)
+	}
+
 	testClus, err := e2eutil.CreateCluster(t, testF.KubeCli, testF.KubeNS, e2eutil.NewCluster("upgrade-test-", 3))
 	if err != nil {
 		t.Fatal(err)
@@ -43,7 +46,6 @@ func TestResize(t *testing.T) {
 		if err := e2eutil.DeleteCluster(t, testF.KubeCli, testClus, &e2eutil.StorageCheckerOptions{}); err != nil {
 			t.Fatal(err)
 		}
-		time.Sleep(10 * time.Second)
 	}()
 	_, err = e2eutil.WaitUntilSizeReached(t, testF.KubeCli, 3, 60*time.Second, testClus)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


hasbro17:
A manual local test of TestResize passes with correct behavior, but we need to change `WaitUntilSizeReached()` to wait for ready members and not running pods since `upgradetest/TestResize` is supposed to test for the readiness of the cluster.

test/e2e/upgradetest/upgrade_test.go
```diff
@@ -34,7 +35,11 @@ func TestResize(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	time.Sleep(10 * time.Second)
+	tprAPIPath := fmt.Sprintf("/apis/%s/%s/namespaces/%s/clusters", spec.TPRGroup, spec.TPRVersion, testF.KubeNS)
+	if err = e2eutil.WaitForAPIPath(testF.KubeCli, tprAPIPath, 10*time.Second); err != nil {
```
hongchaodeng: Such method already exists -- `k8sutil.WaitEtcdTPRReady()`.

Please use that one. Potentially we can make the name better.

test/e2e/upgradetest/upgrade_test.go
```diff
@@ -43,7 +48,6 @@ func TestResize(t *testing.T) {
 		if err := e2eutil.DeleteCluster(t, testF.KubeCli, testClus, &e2eutil.StorageCheckerOptions{}); err != nil {
 			t.Fatal(err)
 		}
-		time.Sleep(10 * time.Second)
```
hongchaodeng: Don't delete this until it checks resources are all deleted.

hasbro17: I thought that `DeleteCluster()` already checks that all resources are deleted by calling `waitResourcesDeleted()` and `waitBackupDeleted()` which waits for the pods, services, backup side car and storage to be deleted. 
Why is the extra sleep needed? Perhaps for the TPR object to be deleted?

hongchaodeng: OK. You are right.

test/e2e/upgradetest/upgrade_test.go
```diff
@@ -34,7 +34,10 @@ func TestResize(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	time.Sleep(10 * time.Second)
+	if err = k8sutil.WaitEtcdTPRReady(testF.KubeCli.CoreV1().RESTClient(), 3*time.Second, 30*time.Second, testF.KubeNS); err != nil {
```
hongchaodeng: Why 3s and 30s?

hongchaodeng: Can you try and see what's the normal time for this?

hasbro17: It takes about 20 seconds from the time of creating the operator to the TPR to being visible at the api server. So a 3s interval check for a total of 30s seemed fine. The interval doesn't matter as much as the total timeout.

hongchaodeng:
LGTM

PR #1059, opened 2017-05-04
doc: access the client service using lb

### LINKED ISSUE #1016: document: write a service to lb over all etcd members.
we do not provide reliable member pod name, thus no reliable FQDN to put into etcd clients.

users normally should use the client service we created or they can create their own one to do load balancing over the etcd pods.

we should document both options.

hasbro17:
[skip ci]
Added documentation on how to access the client service from a pod within the cluster (similar to the README), and how to create a load balancer service to access the client API from outside the cluster.

I have not mentioned how to use an existing load balancer though I assume that should be simple enough by just setting the IP field.

fixes #1016 

/cc @xiang90 @hongchaodeng

### DIFF
--- a/doc/user/client_service.md
+++ b/doc/user/client_service.md
@@ -1,0 +1,63 @@
+## Client service
+For every etcd cluster created the etcd-operator will create an etcd client service in the same namespace with the name `<cluster-name>-client` .
+```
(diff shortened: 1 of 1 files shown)


hongchaodeng:
lgtm

PR #1063, opened 2017-05-05
e2e/upgradetest: 1 member recovery test

hasbro17:
- Added an `e2e/upgrade` test for the newer operator to successfully heal the cluster back to size after a single member failure.
- Moved `KillMembers()` from `e2e/util.go` to `e2e/e2eutil/util.go` to make it common between `e2e` and `e2e/upgradetest` tests.
- Changed the deployment delete grace period to zero, and wait for deployment pods to be deleted, so that the deployment from the previous test is gone before it conflicts with the next test's creation.

### DIFF
--- a/test/e2e/e2eutil/util.go
+++ b/test/e2e/e2eutil/util.go
@@ -27,9 +27,9 @@ import (
 	"k8s.io/client-go/pkg/api/v1"
 )
 
-func KillMembers(kubecli kubernetes.Interface, ns string, names ...string) error {
+func KillMembers(kubecli kubernetes.Interface, namespace string, names ...string) error {
 	for _, name := range names {
-		err := kubecli.CoreV1().Pods(ns).Delete(name, metav1.NewDeleteOptions(0))
+		err := kubecli.CoreV1().Pods(namespace).Delete(name, metav1.NewDeleteOptions(0))
 		if err != nil && !k8sutil.IsKubernetesResourceNotFoundError(err) {
 			return err
 		}
--- a/test/e2e/upgradetest/framework/framework.go
+++ b/test/e2e/upgradetest/framework/framework.go
@@ -15,9 +15,13 @@
 package framework
 
 import (
+	"time"
+
 	"github.com/coreos/etcd-operator/pkg/util/k8sutil"
+	"github.com/coreos/etcd-operator/pkg/util/retryutil"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/pkg/api/v1"
@@ -98,9 +102,25 @@ func (f *Framework) CreateOperator() error {
 
 func (f *Framework) DeleteOperator() error {
 	foreground := metav1.DeletePropagationForeground
-	return f.KubeCli.AppsV1beta1().Deployments(f.KubeNS).Delete("etcd-operator", &metav1.DeleteOptions{
+	err := f.KubeCli.AppsV1beta1().Deployments(f.KubeNS).Delete("etcd-operator", &metav1.DeleteOptions{
 		GracePeriodSeconds: func(t int64) *int64 { return &t }(0),
 		PropagationPolicy:  &foreground,
+	})
+	if err != nil {
+		return err
+	}
+
+	// Wait until the etcd-operator pod is actually gone and not just terminating
+	ls := labels.SelectorFromSet(map[string]string{"name": "etcd-operator"})
+	return retryutil.Retry(5*time.Second, 5, func() (bool, error) {
+		podList, err := f.KubeCli.CoreV1().Pods(f.Config.KubeNS).List(metav1.ListOptions{LabelSelector: ls.String()})
+		if err != nil {
+			return false, err
+		}
+		if len(podList.Items) == 0 {
+			return true, nil
+		}
+		return false, nil
 	})
 }
 
--- a/test/e2e/upgradetest/upgrade_test.go
+++ b/test/e2e/upgradetest/upgrade_test.go
@@ -72,3 +72,45 @@ func TestResize(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestHealOneMemberForOldCluster(t *testing.T) {
+	err := testF.CreateOperator()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer func() {
+		err := testF.DeleteOperator()
+		if err != nil {
+			t.Fatal(err)
+		}
+	}()
+	if err = k8sutil.WaitEtcdTPRReady(testF.KubeCli.CoreV1().RESTClient(), 3*time.Second, 30*time.Second, testF.KubeNS); err != nil {
+		t.Fatalf("failed to see cluster TPR get created in time: %v", err)
+	}
+
+	testEtcd, err := e2eutil.CreateCluster(t, testF.KubeCli, testF.KubeNS, e2eutil.NewCluster("upgrade-test-", 3))
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer func() {
+		if err := e2eutil.DeleteCluster(t, testF.KubeCli, testEtcd, &e2eutil.StorageCheckerOptions{}); err != nil {
+			t.Fatal(err)
+		}
+	}()
+	names, err := e2eutil.WaitUntilSizeReached(t, testF.KubeCli, 3, 60*time.Second, testEtcd)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = testF.UpgradeOperator()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if err := e2eutil.KillMembers(testF.KubeCli, testF.KubeNS, names[2]); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := e2eutil.WaitUntilSizeReached(t, testF.KubeCli, 3, 60*time.Second, testEtcd); err != nil {
+		t.Fatalf("failed to heal one member: %v", err)
+	}
+}


test/e2e/upgradetest/recovery_test.go
```diff
@@ -0,0 +1,74 @@
+// Copyright 2017 The etcd-operator Authors
```
hongchaodeng: It's not worth to create one file for each test.
Let's put this onto upgrade_test.go

test/e2e/upgradetest/resize_test.go
```diff
@@ -29,6 +29,9 @@ func TestResize(t *testing.T) {
 		if err != nil {
 			t.Fatal(err)
 		}
+		if err = e2eutil.WaitUntilDeploymentDeleted(testF.KubeCli, 3*time.Second, 30*time.Second, "etcd-operator", testF.KubeNS); err != nil {
```
hongchaodeng: Why wait for it?
Does it mean deleting deployment is async? Can you verify it?

hasbro17: Yes, when I ran both my tests together one after another, the the test that ran second failed because the previous deployment was still being deleted:
```
--- FAIL: TestResize (0.13s)
	resize_test.go:35: object is being deleted: deployments.extensions "etcd-operator" already exists
FAIL
FAIL	github.com/coreos/etcd-operator/test/e2e/upgradetest	96.266s
```

hongchaodeng: DeleteDeployment() doesn't set grace period, which is 30s by default. We can reduce that to 1s? What about 0s, what does it mean for deployment?

Do you know if deployment deleted means pod is also deleted?

test/e2e/upgradetest/recovery_test.go
hasbro17: It will become a long file by the end with all 5 upgrade test cases.
The issue is that the per test setup/cleanup is a little long and is mostly repeated for each case. Perhaps if we just abstract it we can shorten the test cases and put them in one file. 
The current setup/cleanup being:
```
	defer func() {
		err := testF.DeleteOperator()
		if err != nil {
			t.Fatal(err)
		}
		if err = e2eutil.WaitUntilDeploymentDeleted(testF.KubeCli, 3*time.Second, 30*time.Second, "etcd-operator", testF.KubeNS); err != nil {
			t.Fatalf("failed to delete deployment in time: %v", err)
		}
	}()
	err := testF.CreateOperator()
	if err != nil {
		t.Fatal(err)
	}
	if err = k8sutil.WaitEtcdTPRReady(testF.KubeCli.CoreV1().RESTClient(), 3*time.Second, 30*time.Second, testF.KubeNS); err != nil {
		t.Fatalf("failed to see cluster TPR get created in time: %v", err)
	}

	testClus, err := e2eutil.CreateCluster(t, testF.KubeCli, testF.KubeNS, e2eutil.NewCluster("upgrade-test-", 3))
	if err != nil {
		t.Fatal(err)
	}
	defer func() {
		if err := e2eutil.DeleteCluster(t, testF.KubeCli, testClus, &e2eutil.StorageCheckerOptions{}); err != nil {
			t.Fatal(err)
		}
	}()
	_, err = e2eutil.WaitUntilSizeReached(t, testF.KubeCli, 3, 60*time.Second, testClus)
	if err != nil {
		t.Fatal(err)
	}
	err = testF.UpgradeOperator()
	if err != nil {
		t.Fatal(err)
	}
```
Let me know if you're still sure we should put all the cases in one file?

hongchaodeng: > It will become a long file by the end with all 5 upgrade test cases.

It doesn't matter :)

> The issue is that the per test setup/cleanup is a little long and is mostly repeated for each case. Perhaps if we just abstract it we can shorten the test cases and put them in one file.

That's orthogonal. It doesn't matter if we put them in one file.

> Let me know if you're still sure we should put all the cases in one file?

yes.

test/e2e/upgradetest/resize_test.go
hasbro17: Okay changing the grace period to zero resolved the conflict.
It seems the deployment goes away immediately(as far as I could see) but the pod takes some more time to delete.
For the purposes of this test since the pod names are generated they don't conflict. Only the deployment needs to be gone before the next test.

hongchaodeng: The pod name is not unique. Moreover, there should not be two existing operator pods at any give time because they manage the same namespace.

My experience with `kubectl delete deployment` is that it will return after deployment and its pods are deleted. We might want the same behavior. The question is if we change grace period to 1 for deployment, what would happen? Is Delete() call sync? Otherwise how is kubectl implemented?

test/e2e/upgradetest/framework/framework.go
```diff
@@ -97,10 +97,7 @@ func (f *Framework) CreateOperator() error {
 }
 
 func (f *Framework) DeleteOperator() error {
-	foreground := metav1.DeletePropagationForeground
-	return f.KubeCli.AppsV1beta1().Deployments(f.KubeNS).Delete("etcd-operator", &metav1.DeleteOptions{
-		PropagationPolicy: &foreground,
-	})
+	return f.KubeCli.AppsV1beta1().Deployments(f.KubeNS).Delete("etcd-operator", metav1.NewDeleteOptions(0))
```
hongchaodeng: Is `metav1.NewDeleteOptions(0)` going to work? I think this is not going to delete its children, i.e. RS and pods. Can you verify that?

hongchaodeng:
Update:
- we have experimented with deployment deletion. When we call `Deployments().Delete()`, Deployment was deleted first and Pod was deleted later from API.

Conclusion: We need to wait for Pod to be deleted from API.

hasbro17:
Now waiting for all pods with the label `name=etcd-operator` to be gone before we successfully return from `DeleteOperator()`.

test/e2e/upgradetest/upgrade_test.go
```diff
@@ -72,3 +72,51 @@ func TestResize(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestFailureRecovery(t *testing.T) {
```
hongchaodeng: TestHealOneMemberForOldCluster

hongchaodeng: No sub test

test/e2e/e2eutil/util.go
```diff
@@ -27,9 +27,9 @@ import (
 	"k8s.io/client-go/pkg/api/v1"
 )
 
-func KillMembers(kubecli kubernetes.Interface, ns string, names ...string) error {
+func KillMembers(kubeClient kubernetes.Interface, namespace string, names ...string) error {
```
hongchaodeng: Let's use `kubecli` for consistency :)

test/e2e/upgradetest/framework/framework.go
```diff
@@ -98,9 +102,24 @@ func (f *Framework) CreateOperator() error {
 
 func (f *Framework) DeleteOperator() error {
 	foreground := metav1.DeletePropagationForeground
-	return f.KubeCli.AppsV1beta1().Deployments(f.KubeNS).Delete("etcd-operator", &metav1.DeleteOptions{
-		GracePeriodSeconds: func(t int64) *int64 { return &t }(0),
+	if err := f.KubeCli.AppsV1beta1().Deployments(f.KubeNS).Delete("etcd-operator", &metav1.DeleteOptions{
```
hongchaodeng:
```
err := f.KubeCli.AppsV1beta1().Deployment ...
if err != nil {...
```

test/e2e/upgradetest/framework/framework.go
```diff
@@ -98,9 +102,24 @@ func (f *Framework) CreateOperator() error {
...
+		GracePeriodSeconds: func(t int64) *int64 { return &t }(1),
 		PropagationPolicy:  &foreground,
+	}); err != nil {
+		return err
+	}
+
+	// Wait until the etcd-operator pod is actually gone and not just terminating
+	ls := labels.SelectorFromSet(map[string]string{"name": "etcd-operator"})
+	return retryutil.Retry(5*time.Second, 10, func() (bool, error) {
```
hongchaodeng: It usually took 4-5s for a pod to be gone. Let's just wait for 5s x 2 .

test/e2e/upgradetest/framework/framework.go
```diff
@@ -98,9 +102,24 @@ func (f *Framework) CreateOperator() error {
 
 func (f *Framework) DeleteOperator() error {
 	foreground := metav1.DeletePropagationForeground
-	return f.KubeCli.AppsV1beta1().Deployments(f.KubeNS).Delete("etcd-operator", &metav1.DeleteOptions{
-		GracePeriodSeconds: func(t int64) *int64 { return &t }(0),
+	if err := f.KubeCli.AppsV1beta1().Deployments(f.KubeNS).Delete("etcd-operator", &metav1.DeleteOptions{
+		GracePeriodSeconds: func(t int64) *int64 { return &t }(1),
```
hongchaodeng: Don't change this?

test/e2e/e2eutil/util.go
hasbro17: `e2eutil/wait_util.go` and `e2eutil/tpr_util.go` all use the name `kubeClient`. Should I change those to kubeCli too?

hongchaodeng: Not worry about those now.
It is  this PR that doesn't need to change this.

test/e2e/upgradetest/framework/framework.go
hasbro17: It's taking longer than that. Lets keep it at 5s x 5 for now.

test/e2e/e2eutil/util.go
```diff
@@ -27,9 +27,9 @@ import (
 	"k8s.io/client-go/pkg/api/v1"
 )
 
-func KillMembers(kubecli kubernetes.Interface, ns string, names ...string) error {
+func KillMembers(kubeCli kubernetes.Interface, namespace string, names ...string) error {
```
hongchaodeng: kubeCli -> kubecli ...
Please don't change it.

test/e2e/upgradetest/upgrade_test.go
```diff
@@ -72,3 +72,45 @@ func TestResize(t *testing.T) {
...
+
+	err = testF.UpgradeOperator()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if err := e2eutil.KillMembers(testF.KubeCli, testF.KubeNS, names[2]); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := e2eutil.WaitUntilSizeReached(t, testF.KubeCli, 3, 60*time.Second, testEtcd); err != nil {
```
hongchaodeng: It's healing just one member, why still wait 60s? I feel like 30s is enough.

test/e2e/upgradetest/upgrade_test.go
```diff
@@ -72,3 +72,45 @@ func TestResize(t *testing.T) {
...
+		t.Fatalf("failed to resize to 3 members etcd cluster: %v", err)
```
hongchaodeng: failed to heal one member

test/e2e/upgradetest/upgrade_test.go
hasbro17: It fails with 30 seconds because just before killing a member we upgrade the operator first. And it seems like it takes some time for the new operator to startup, detect failed member and heal the cluster.

hongchaodeng: OK. Thanks for looking into that :)

hongchaodeng:
LGTM

PR #1074, opened 2017-05-08
spec: add s3 configuration into s3 source

xiang90:
/cc @hongchaodeng @hasbro17 

[skip ci]

### DIFF
--- a/pkg/spec/backup.go
+++ b/pkg/spec/backup.go
@@ -73,7 +73,20 @@ type PVSource struct {
 	VolumeSizeInMB int `json:"volumeSizeInMB"`
 }
 
+// TODO: support per cluster S3 Source configuration.
 type S3Source struct {
+	// The name of the AWS S3 bucket to store backups in.
+	//
+	// S3Bucket overwrites the default etcd operator wide bucket.
+	S3Bucket string `json:"s3Bucket,omitempty"`
+
+	// The name of the secret object that stores the AWS credential and config files.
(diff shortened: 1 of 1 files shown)


pkg/spec/backup.go
```diff
@@ -73,7 +73,24 @@ type PVSource struct {
...
+	// The name of the AWS S3 bucket to store backups in.
+	//
+	// S3Bucket overwrites the default etcd operator wide bucket.
+	S3Bucket string `json:"s3Bucket,omitempty"`
+
+	// The name of the secret object that stores the AWS credential file.
+	// The file name must be 'credentials'.
+	//
+	// AWSCredentialSecret overwrites the default etcd operator wide AWS credential.
+	AWSCredentialSecret string `json:"awsCredentialSecret,omitempty"`
```
hongchaodeng: Based on past experience, it would be better to combine both credential and config into one secret object.

xiang90: sounds good to me. I also feel so. 

are we going to deprecate the old behavior in etcd operator?

hongchaodeng: Let's put the spec in first. Then we will think about upgrade (maybe not in next release 0.2.7). Then we think about deprecation.

xiang90:
@hongchaodeng updated.

hasbro17:
LGTM

hongchaodeng:
lgtm

PR #1080, opened 2017-05-09
upgradetest: Add test to restore from backup of old cluster

hasbro17:
- Added `TestRestoreFromBackup ` to test if the upgraded operator can restore a cluster from the backup of an older cluster.
- Moved utils for reading and writing data to the etcd cluster into `e2eutil/etcd_util.go`.
- Added a test selector in `upgrade_pass` to make it easy to selectively run tests. And a timeout of 30m since all upgrade tests take at least 10 minutes currently.
- Made the `genName` passed to `CreateCluster()` unique for each upgrade test for ease of manual testing.

### DIFF
--- a/hack/test
+++ b/hack/test
@@ -90,10 +90,11 @@ function e2e_pass {
 }
 
 function upgrade_pass {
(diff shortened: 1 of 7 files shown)


hasbro17:
@hongchaodeng @xiang90 all upgrade tests passing. PTAL.

PR #1082, opened 2017-05-09
upgradetest: Add disaster recovery test

hasbro17:
Added `TestDisasterRecovery()` to test if the upgraded operator can do disaster recovery from the backup of the old cluster once all its pods are deleted.

### DIFF
--- a/test/e2e/upgradetest/upgrade_test.go
+++ b/test/e2e/upgradetest/upgrade_test.go
@@ -312,6 +312,93 @@ func testBackupForOldClusterWithBackupPolicy(t *testing.T, bp *spec.BackupPolicy
 	}
 }
 
+// TestDisasterRecovery tests if the new operator could do disaster recovery from backup of the old cluster.
+func TestDisasterRecovery(t *testing.T) {
+	t.Run("Recover from PV backup", func(t *testing.T) {
+		testDisasterRecoveryWithBackupPolicy(t, e2eutil.NewPVBackupPolicy())
+	})
+	t.Run("Recover from S3 backup", func(t *testing.T) {
+		testDisasterRecoveryWithBackupPolicy(t, e2eutil.NewS3BackupPolicy())
+	})
+}
+
+func testDisasterRecoveryWithBackupPolicy(t *testing.T, bp *spec.BackupPolicy) {
+	err := testF.CreateOperator()
+	if err != nil {
+		t.Fatalf("failed to create operator: %v", err)
+	}
+	defer func() {
+		err := testF.DeleteOperator()
+		if err != nil {
+			t.Fatalf("failed to delete operator: %v", err)
+		}
+	}()
+	err = k8sutil.WaitEtcdTPRReady(testF.KubeCli.CoreV1().RESTClient(), 3*time.Second, 30*time.Second, testF.KubeNS)
+	if err != nil {
+		t.Fatalf("failed to see cluster TPR get created in time: %v", err)
+	}
+
+	bp.CleanupBackupsOnClusterDelete = true
+	testClus := e2eutil.NewCluster("upgrade-test-recovery-", 3)
+	testClus = e2eutil.ClusterWithBackup(testClus, bp)
+	testClus, err = e2eutil.CreateCluster(t, testF.KubeCli, testF.KubeNS, testClus)
+	if err != nil {
+		t.Fatalf("failed to create cluster: %v", err)
+	}
+	defer func() {
+		checker := e2eutil.StorageCheckerOptions{
+			S3Cli:    testF.S3Cli,
+			S3Bucket: testF.S3Bucket,
+		}
+		err := e2eutil.DeleteClusterAndBackup(t, testF.KubeCli, testClus, checker)
+		if err != nil {
+			t.Fatalf("failed to delete cluster and its backup: %v", err)
+		}
+	}()
+
+	names, err := e2eutil.WaitUntilSizeReached(t, testF.KubeCli, 3, 60*time.Second, testClus)
+	if err != nil {
+		t.Fatalf("failed to reach desired cluster size: %v", err)
+	}
+	err = e2eutil.WaitBackupPodUp(t, testF.KubeCli, testF.KubeNS, testClus.Metadata.Name, 60*time.Second)
+	if err != nil {
+		t.Fatalf("failed to create backup pod: %v", err)
+	}
+
+	// write data to etcd and make a backup
+	pod, err := testF.KubeCli.CoreV1().Pods(testF.KubeNS).Get(names[0], metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("failed to get backup pod: %v", err)
+	}
+	err = e2eutil.PutDataToEtcd(fmt.Sprintf("http://%s:2379", pod.Status.PodIP))
+	if err != nil {
+		t.Fatalf("failed to put data to etcd: %v", err)
+	}
+	err = e2eutil.MakeBackup(testF.KubeCli, testClus.Metadata.Namespace, testClus.Metadata.Name)
+	if err != nil {
+		t.Fatalf("failed to make backup: %v", err)
+	}
+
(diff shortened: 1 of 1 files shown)


hasbro17:
@xiang90 @hongchaodeng PTAL. Upgrade tests are passing.

test/e2e/e2eutil/util.go
```diff
@@ -91,6 +91,15 @@ func LogfWithTimestamp(t *testing.T, format string, args ...interface{}) {
 	t.Log(time.Now(), fmt.Sprintf(format, args...))
 }
 
+// get the image of the first container of the pod in the deployment.
+func GetContainerImageNameFromDeployment(kubecli kubernetes.Interface, name, namespace string) (string, error) {
```
hongchaodeng: Don't take this into e2eutil unless it's used somewhere else.

test/e2e/upgradetest/upgrade_test.go
```diff
@@ -312,11 +312,92 @@ func testBackupForOldClusterWithBackupPolicy(t *testing.T, bp *spec.BackupPolicy
 	}
 }
 
-// get the image of the first container of the pod in the deployment.
-func getContainerImageNameFromDeployment(name string) (string, error) {
-	d, err := testF.KubeCli.AppsV1beta1().Deployments(testF.KubeNS).Get(name, metav1.GetOptions{})
+// TestDisasterRecovery tests if the new operator could do disaster recovery from backup of the old cluster.
+func TestDisasterRecovery(t *testing.T) {
+	t.Run("Recovery from PV backup", func(t *testing.T) {
```
hongchaodeng: Recover from PV backup

test/e2e/upgradetest/upgrade_test.go
```diff
@@ -312,11 +312,92 @@ func testBackupForOldClusterWithBackupPolicy(t *testing.T, bp *spec.BackupPolicy
...
+		testDisasterRecoveryWithBackupPolicy(t, e2eutil.NewPVBackupPolicy())
+	})
+	t.Run("Recovery from S3 backup", func(t *testing.T) {
```
hongchaodeng: Recover from ...

test/e2e/upgradetest/upgrade_test.go
```diff
@@ -312,11 +312,92 @@ func testBackupForOldClusterWithBackupPolicy(t *testing.T, bp *spec.BackupPolicy
...
+		testDisasterRecoveryWithBackupPolicy(t, e2eutil.NewS3BackupPolicy())
+	})
+}
+
+func testDisasterRecoveryWithBackupPolicy(t *testing.T, bp *spec.BackupPolicy) {
+	// create operator
```
hongchaodeng: delete this unnecessary comment

test/e2e/upgradetest/upgrade_test.go
```diff
@@ -312,11 +312,92 @@ func testBackupForOldClusterWithBackupPolicy(t *testing.T, bp *spec.BackupPolicy
...
+		if err != nil {
+			t.Fatalf("failed to delete operator:%v", err)
+		}
+	}()
+	err = k8sutil.WaitEtcdTPRReady(testF.KubeCli.CoreV1().RESTClient(), 3*time.Second, 30*time.Second, testF.KubeNS)
+	if err != nil {
+		t.Fatalf("failed to see cluster TPR get created in time: %v", err)
+	}
+
+	// create cluster
```
hongchaodeng: delete it

test/e2e/upgradetest/upgrade_test.go
```diff
@@ -312,11 +312,92 @@ func testBackupForOldClusterWithBackupPolicy(t *testing.T, bp *spec.BackupPolicy
...
+	bp.BackupIntervalInSecond = 60 * 60 * 24 // long enough that no backup was made automatically
```
hongchaodeng: Why do we care about it in this test?

test/e2e/upgradetest/upgrade_test.go
```diff
@@ -312,11 +312,92 @@ func testBackupForOldClusterWithBackupPolicy(t *testing.T, bp *spec.BackupPolicy
...
+	bp.CleanupBackupsOnClusterDelete = true
+	origClus := e2eutil.NewCluster("upgrade-test-recovery-", 3)
```
hongchaodeng: We aren't using origClus anywhere else. Why not just use one variable testClus .

test/e2e/upgradetest/upgrade_test.go
```diff
@@ -312,11 +312,92 @@ func testBackupForOldClusterWithBackupPolicy(t *testing.T, bp *spec.BackupPolicy
...
+		t.Fatalf("failed to kill all members:%v", err)
+	}
+	if names, err = e2eutil.WaitUntilSizeReached(t, testF.KubeCli, 3, 180*time.Second, testClus); err != nil {
+		t.Fatalf("failed to recover all members: %v", err)
+	}
+
+	// ensure the data from the previous cluster is present
+	pod, err = testF.KubeCli.CoreV1().Pods(testF.KubeNS).Get(names[0], metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("failed to get backup pod:%v", err)
```
hongchaodeng: pod: %v

test/e2e/upgradetest/upgrade_test.go
hasbro17: Well we only really care about creating the backup after writing data to etcd. 
So we can't rely on a periodic backup by setting the `BackupIntervalInSecond ` since there's no guarantee if that will make a backup immediately after `PutDataToEtcd`.

That's why we avoid the periodic backup with a long timeout and just request the backup explicitly after `PutDataToEtcd`.

hongchaodeng: Why do we want to avoid periodic backup?

hasbro17: Well we don't have to avoid it. All we require for the test is that a backup is created after `PutDataToEtcd`.

We can do this by either explicitly calling `MakeBackup`, or we can rely on the periodic backup by setting a small timeout and polling backup status until we see that the backup has been created. Currently I'm just doing the former since that's simpler.

I did not see the point of doing both.

hongchaodeng: Yep. I agree with that.

test/e2e/upgradetest/upgrade_test.go
```diff
@@ -312,6 +312,94 @@ func testBackupForOldClusterWithBackupPolicy(t *testing.T, bp *spec.BackupPolicy
...
+		if err != nil {
+			t.Fatalf("failed to delete operator: %v", err)
+		}
+	}()
+	err = k8sutil.WaitEtcdTPRReady(testF.KubeCli.CoreV1().RESTClient(), 3*time.Second, 30*time.Second, testF.KubeNS)
+	if err != nil {
+		t.Fatalf("failed to see cluster TPR get created in time: %v", err)
+	}
+
+	bp.BackupIntervalInSecond = 60 * 60 * 24 // long enough that no backup was made automatically
```
hongchaodeng: Based on previous discussion, I think we don't need this, right?

hongchaodeng:
The code LGTM.
But I have one nit point.

test/e2e/upgradetest/upgrade_test.go
hasbro17: Yes, I just checked that the default `BackupIntervalInSecond` is set to 1800. Which is long enough. I'll remove this.

hasbro17: Edit: It's actually 3600s when we create the backup policy which is still fine for our purpose.

hongchaodeng: It's not about how long it is.
It's that no matter periodic backup is made or not, it doesn't change the result.

hongchaodeng:
lgtm

Issue #1046, opened 2017-05-03
Upgrade from 0.2.6 to 0.3.0


hongchaodeng:
Upgrade test:
- [x] umbrella: https://github.com/coreos/etcd-operator/issues/989
  - [X] Resize test
  - [X] One member healing test. (https://github.com/coreos/etcd-operator/pull/1063)
  - [x] Make backup from old cluster test. (https://github.com/coreos/etcd-operator/pull/1067)
  - [x] recovery from old backup (https://github.com/coreos/etcd-operator/pull/1082)
  - [x] Restore from old backup (https://github.com/coreos/etcd-operator/pull/1080)

We are providing backward compatibility for some features during 0.2.5 -> 0.2.6. We are going to drop this in 0.2.7 . From 0.2.5 or less, users must upgrade to 0.2.6 first, and make sure new backups are made, then upgrade to 0.2.7 .
- [x] S3 dir layout are changed from ${clusterName}/ to ${namespace}/${clusterName}/ (https://github.com/coreos/etcd-operator/pull/1086)
- [X] backup sidecar RS was deleted and Deployment created. (https://github.com/coreos/etcd-operator/pull/1072)

Other issues:
- [x] In #1069, we fixed a bug of backup sidecar upgrade. For 0.2.6 it has the bug, we must update the strategy during upgrade in 0.2.7 . (https://github.com/coreos/etcd-operator/pull/1072)
- [x] Upgrade path doc. Emphasizing that all <0.2.6 versions should upgrade to 0.2.6 first before upgrading to 0.2.7 (https://github.com/coreos/etcd-operator/issues/1091)

hongchaodeng:
In https://github.com/coreos/etcd-operator/pull/1069, we fixed a bug of backup sidecar upgrade. For 0.2.6 it has the bug, we must update the strategy during upgrade in 0.2.7 .

hongchaodeng:
Updated this issue with tasks we need to finish before 0.2.7 release

hongchaodeng:
Our next release is changed to 0.3.0 because we drop compatiblity.
We need provide upgrade path docs.

hongchaodeng:
All done.

PR #1086, opened 2017-05-11
*: drop compatability for non-namespaced s3 backup paths

hasbro17:
Removed `CopyPrefix(clusterName)` so the backup manager will no longer be backwards compatible with non-namespaced s3 backups created by operators <= v0.2.5

### DIFF
--- a/pkg/backup/backup.go
+++ b/pkg/backup/backup.go
@@ -81,12 +81,6 @@ func New(kclient kubernetes.Interface, clusterName, ns string, sp spec.ClusterSp
 			return nil, err
 		}
 
-		// for backward compatibility.
-		err = s3cli.CopyPrefix(clusterName)
-		if err != nil {
-			return nil, err
-		}
-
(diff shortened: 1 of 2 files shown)


hongchaodeng:
ref: https://github.com/coreos/etcd-operator/issues/1046

We need write explicitly the upgrade path to 0.2.7 in both release and upgrade docs.

hongchaodeng:
LGTM

hongchaodeng:
@xiang90 Also give a comment?

xiang90:
lgtm

hasbro17:
Manually testing that the upgrade should fail for `v0.2.5`=>`v0.2.7`/this PR, once the backward compatibility support is dropped.

Currently the `TestBackupForOldCluster` test fails due to another incompatibility because all versions below `v0.2.6` use replica set instead of deployment for backup-sidecar.

hongchaodeng:
lgtm

PR #1087, opened 2017-05-11
upgradetest: remove backup interval

hasbro17:
[skip ci]
No need to set the backup interval since our test is not reliant on the periodic backup.


hongchaodeng:
LGTM

PR #1089, opened 2017-05-11
upgradetest: remove other backupinterval

hasbro17:
[skip ci]

Only keep periodic backup interval long for `TestBackupForOldCluster` and not for `TestRestoreFromBackup` since it doesn't matter in that case if periodic backup is made or not.
Revert #1087


hasbro17:
@hongchaodeng PTAL

hongchaodeng:
lgtm

PR #1101, opened 2017-05-16
e2eutil/wait_util: wait until etcd member is ready

### LINKED ISSUE #1060: e2e/wait_util: should wait for ready members and not number of pods
Currently the function  `waitSizeReachedWithAccept()` is used in the e2e tests to check if a cluster successfully reaches the desired size after some action. Its current implementation only checks the number of pods in the cluster. Rather it should check the number of ready members in the etcd cluster to ensure that the cluster is healthy and of the desired size.

hasbro17:
`waitSizeReachedWithAccept()` now waits until the cluster TPR shows the desired number of ready members.

`WaitUntilPodSizeReached` is needed for the tests `TestPauseControl` and `TestRestartOperator` to directly check the number of pods, since the TPR won't be updated while the operator is paused/deleted. 

fixes #1060

### DIFF
--- a/test/e2e/basic_test.go
+++ b/test/e2e/basic_test.go
@@ -82,10 +82,10 @@ func TestPauseControl(t *testing.T) {
 	if err := e2eutil.KillMembers(f.KubeClient, f.Namespace, names[0]); err != nil {
 		t.Fatal(err)
 	}
-	if _, err := e2eutil.WaitUntilSizeReached(t, f.KubeClient, 2, 10*time.Second, testEtcd); err != nil {
+	if _, err := e2eutil.WaitUntilPodSizeReached(t, f.KubeClient, 2, 10*time.Second, testEtcd); err != nil {
 		t.Fatalf("failed to wait for killed member to die: %v", err)
 	}
-	if _, err := e2eutil.WaitUntilSizeReached(t, f.KubeClient, 3, 10*time.Second, testEtcd); err == nil {
+	if _, err := e2eutil.WaitUntilPodSizeReached(t, f.KubeClient, 3, 10*time.Second, testEtcd); err == nil {
 		t.Fatalf("cluster should not be recovered: control is paused")
 	}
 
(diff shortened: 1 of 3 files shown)


test/e2e/e2eutil/wait_util.go
```diff
@@ -33,7 +33,7 @@ import (
 	"k8s.io/client-go/pkg/api/v1"
 )
 
-type acceptFunc func(*v1.Pod) bool
+type conditionFunc func(*spec.Cluster) bool
```
hongchaodeng: Still use "acceptFunc" but only change parameter type?

hasbro17:
@hongchaodeng PTAL.

test/e2e/basic_test.go
```diff
@@ -82,10 +82,10 @@ func TestPauseControl(t *testing.T) {
 	if err := e2eutil.KillMembers(f.KubeClient, f.Namespace, names[0]); err != nil {
 		t.Fatal(err)
 	}
-	if _, err := e2eutil.WaitUntilSizeReached(t, f.KubeClient, 2, 10*time.Second, testEtcd); err != nil {
+	if _, err := e2eutil.WaitUntilPodSizeReached(t, f.KubeClient, 2, 10*time.Second, testEtcd); err != nil {
```
hongchaodeng: Please don't change the name in this PR

hasbro17: @hongchaodeng The name has not changed. All the tests are still using `WaitUntilSizeReached` for which only the implementation has changed.
Over here `WaitUntilSizeReached()` with the new implementation cannot be used because the operator is paused and the TPR will never be updated to show the size reduce to 2. So we use `WaitUntilPodSizeReached` which has the original implementation of checking the number of pods directly.

The tests for `TestPauseControl ` and `TestRestartOperator ` will not pass if this is kept to `WaitUntilSizeReached()`.

hongchaodeng:
lgtm

PR #1114, opened 2017-05-18
doc: correct diagrams; replace tabs with sapces.

ElijahCaine:
~[The cluster_tls page](https://coreos.com/operators/etcd/docs/latest/design/cluster_tls.html#trust-delegation-diagram) has a diagram which includes mixed spaces and tabs, making it look mangled. This fixes that malformatting.~

~cc @colhom  to make sure the diagram is correct.~

UDPATE: Actually we're deleting that doc instead. @hongchaodeng does this look good to you? I can remove that other commit.

[skip ci]


hongchaodeng:
LGTM

PR #1135, opened 2017-05-27
spec: add support for host networking

clusc:
We have a usage where we need to deploy metric collection agent(collectd) as daemonset on all nodes running etcd(one pod per node). The problem is the agent needs a stable way to talk to the etcd instance on the same node. By running both etcd and collectd under host networking mode, the metric agent can simply talk to `localhost:2379` to collect the metric.


hongchaodeng:
Please report issue first and add your use case and requirement.
We need to discuss each feature and its usage. After approval, we will prioritize into release schedule.

hongchaodeng:
We also need tests.

hongchaodeng:
Closing this since parent issue is closed too.

Issue #953, opened 2017-04-07
Enable StorageClass as option


scheeles:
We want to have more control over the storage. On AWS e.g. we want to use `encrypted` and control the `type`. Additional we use the etcdOperator on Glusterfs and have backup in place.

I think we need can have both option "pv-provisioner" and "storageclass" options.

"pv-provisioner"  is used for new user who wants to start fast, without thinking about the storage in detail.

"storageclass" is for more advanced use case, where you want complete control over the storage. Additional with storageclass we can support more storage provider like Glusterfs. 

Perhaps can make it even more flexible, that you can define per cluster a class e.g.:
```
apiVersion: "etcd.coreos.com/v1beta1"
kind: "Cluster"
metadata:
  name: "etcd-cluster"
spec:
  size: 3
  version: "3.1.4"
  backup:
    # short snapshot interval for testing, do not use this in production!
    backupIntervalInSecond: 30
    maxBackups: 50
    storageClass: "Etcdssd"
    pv:
      volumeSizeInMB: 2048
```

scheeles:
It is related with PR https://github.com/coreos/etcd-operator/pull/949

xiang90:
the storageclass at the operator level was designed for convenience, not for flexibility. so we only want to include a fixed number of storage classes.

the plan was to add a storage class option into the cluster spec, so user can overwrite the global default storage call set at the operator level. this provides the best flexibility.

@scheeles what do you think?

scheeles:
Partly agree. I think we need the storage class also on the operator level (S3 is also available on the operator level). 
You could have a team which is responsible for the etcd operator and the backup (location) of etcd. The developer then should only deploy the etcd cluster. 
I think all backup related thing should be possible to control on the operator level and it should be possible to overwrite on cluster level.

xiang90:
> I think all backup related thing should be possible to control on the operator level 

if we want to support customized storage class (not derived from provisioner), we need to:

1. write a documentation to explain why we have two different pv related flags (storage class vs provisioner) and when to use which. and how are we going to verify the operator level storage class to fail fast.

2. write code to reflect the documented feature

3. write tests to cover the use case

i am not against it, i just want to make everything clear before we move forward. also i want to see storage class overwriting at cluster level to happen.

hongchaodeng:
@scheeles 
Thanks for writing out the issue.

> S3 is also available on the operator level

[...]

PR #1143, opened 2017-05-30
Use a pre-defined StorageClass

narayanan:
- Added support to use a pre-defined StorageClass
- Added Openstack cinder backup support

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,7 +344,7 @@ Check the other two pods and you should see the same result.
 
 - Backup works only for data in etcd3 storage, not for data in etcd2 storage.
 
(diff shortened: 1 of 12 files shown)


xiang90:
@narayanan Thanks for the pull request.

We actually thought about this a lot. So the operate wide pre-defined storage class is for convenience purpose. If we want to allow the max flexible, a better way is to expose storage class in the PV backup policy at cluster wide. So user can use whatever storage class for their own clusters.

What do you think?

hongchaodeng:
Converge the discussion at https://github.com/coreos/etcd-operator/issues/953#issuecomment-305031796 . Let's discuss there if there is any question.

PR #1165, opened 2017-06-05
*: add readiness probe for operator pod

hasbro17:
The operator now creates an http endpoint to indicate its readiness to the kubelet and other components. The endpoint is set to ready right after the controller starts watching for cluster events.

TODO: Set the endpoint to unready if the controller exits or stops?

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"net/http"
 	"os"
 	"os/signal"
 	"runtime"
@@ -32,6 +33,7 @@ import (
 	"github.com/coreos/etcd-operator/pkg/util/k8sutil"
 	"github.com/coreos/etcd-operator/pkg/util/k8sutil/election"
 	"github.com/coreos/etcd-operator/pkg/util/k8sutil/election/resourcelock"
+	"github.com/coreos/etcd-operator/pkg/util/probe"
 	"github.com/coreos/etcd-operator/pkg/util/retryutil"
 	"github.com/coreos/etcd-operator/version"
 
@@ -54,6 +56,7 @@ var (
 	awsSecret        string
 	awsConfig        string
 	s3Bucket         string
+	listenAddr       string
 	gcInterval       time.Duration
 
 	chaosLevel int
@@ -76,6 +79,7 @@ func init() {
 	flag.StringVar(&awsConfig, "backup-aws-config", "",
 		"The name of the kube configmap object that stores the AWS config file. The file name must be 'config'.")
 	flag.StringVar(&s3Bucket, "backup-s3-bucket", "", "The name of the AWS S3 bucket to store backups in.")
+	flag.StringVar(&listenAddr, "listen-addr", "0.0.0.0:8080", "The address on which the HTTP server will listen to")
 	// chaos level will be removed once we have a formal tool to inject failures.
 	flag.IntVar(&chaosLevel, "chaos-level", -1, "DO NOT USE IN PRODUCTION - level of chaos injected into the etcd clusters created by the operator.")
 	flag.BoolVar(&printVersion, "version", false, "Show version and quit")
@@ -150,6 +154,9 @@ func main() {
 			EventRecorder: createRecorder(kubecli, name, namespace),
 		},
 	}
+
+	http.HandleFunc(probe.HTTPReadyzEndpoint, probe.ReadyzHandler)
+	go http.ListenAndServe(listenAddr, nil)
 
 	election.RunOrDie(election.LeaderElectionConfig{
 		Lock:          rl,
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -29,6 +29,7 @@ import (
 	"github.com/coreos/etcd-operator/pkg/spec"
 	"github.com/coreos/etcd-operator/pkg/util/constants"
 	"github.com/coreos/etcd-operator/pkg/util/k8sutil"
+	"github.com/coreos/etcd-operator/pkg/util/probe"
 
 	"github.com/Sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -313,6 +314,8 @@ func (c *Controller) watch(watchVersion string) (<-chan *Event, <-chan error) {
 			}
 
 			c.logger.Infof("start watching at %v", watchVersion)
+
+			probe.SetReady()
 
 			decoder := json.NewDecoder(resp.Body)
 			for {
--- a/pkg/util/probe/readyz.go
+++ b/pkg/util/probe/readyz.go
@@ -1,0 +1,47 @@
+// Copyright 2017 The etcd-operator Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package probe
+
+import (
+	"net/http"
+	"sync"
+)
+
+const (
+	HTTPReadyzEndpoint = "/readyz"
+)
+
+var (
+	mu    sync.Mutex
+	ready = false
+)
+
+func SetReady() {
+	mu.Lock()
+	ready = true
+	mu.Unlock()
+}
+
+// ReadyzHandler writes back the HTTP status code 200 if the operator is ready, and 500 otherwise
+func ReadyzHandler(w http.ResponseWriter, r *http.Request) {
+	mu.Lock()
+	isReady := ready
+	mu.Unlock()
+	if isReady {
+		w.WriteHeader(http.StatusOK)
+	} else {
+		w.WriteHeader(http.StatusInternalServerError)
+	}
+}


pkg/util/constants/constants.go
```diff
@@ -28,4 +28,6 @@ const (
 	PVProvisionerGCEPD  = "kubernetes.io/gce-pd"
 	PVProvisionerAWSEBS = "kubernetes.io/aws-ebs"
 	PVProvisionerNone   = "none"
+
+	ReadyProbeFilePath = "/ready"
```
hongchaodeng: Put this var into k8sutil ?
This is more our k8s related logic.

test/e2e/e2eutil/wait_util.go
```diff
@@ -240,3 +240,23 @@ func waitPodsDeleted(kubecli kubernetes.Interface, namespace string, timeout tim
 	})
 	return pods, err
 }
+
+func WaitUntilPodsReady(kubecli kubernetes.Interface, size int, namespace string, timeout time.Duration, lo metav1.ListOptions) error {
```
hongchaodeng: `size` is not needed, right?
Also switch `timeout` and `lo` position? Group all k8s stuff together is better.

test/e2e/e2eutil/wait_util.go
```diff
@@ -240,3 +240,23 @@ func waitPodsDeleted(kubecli kubernetes.Interface, namespace string, timeout tim
...
+	return retryutil.Retry(5*time.Second, int(timeout/(5*time.Second)), func() (bool, error) {
+		var pods []*v1.Pod
```
hongchaodeng: What is this used for?

test/e2e/e2eutil/wait_util.go
```diff
@@ -240,3 +240,23 @@ func waitPodsDeleted(kubecli kubernetes.Interface, namespace string, timeout tim
...
+		podList, err := kubecli.CoreV1().Pods(namespace).List(lo)
```
hongchaodeng: Why not use watch()?

pkg/util/constants/constants.go
hongchaodeng: "/ready" is in root dir.
Please choose another dir.

test/e2e/upgradetest/framework/framework.go
```diff
@@ -98,13 +101,29 @@ func (f *Framework) CreateOperator() error {
 								ValueFrom: &v1.EnvVarSource{FieldRef: &v1.ObjectFieldSelector{FieldPath: "metadata.name"}},
 							},
 						},
+						ReadinessProbe: &v1.Probe{
+							Handler: v1.Handler{
+								Exec: &v1.ExecAction{
+									Command: readyCmd,
+								},
+							},
+							InitialDelaySeconds: 5,
```
hongchaodeng: Why 5s? It seems too long.

pkg/controller/controller.go
```diff
@@ -141,6 +142,11 @@ func (c *Controller) Run() error {
 		c.waitCluster.Wait()
 	}()
 
+	_, err = os.Create(constants.ReadyProbeFilePath)
```
hongchaodeng: Abstract a method like SetReady() or TellReady()?

test/e2e/upgradetest/framework/framework.go
```diff
@@ -98,13 +101,29 @@ func (f *Framework) CreateOperator() error {
...
 	_, err := f.KubeCli.AppsV1beta1().Deployments(f.KubeNS).Create(d)
-	return err
+	if err != nil {
+		return err
+	}
+
+	lo := metav1.ListOptions{
+		LabelSelector: labels.SelectorFromSet(map[string]string{"name": "etcd-operator"}).String(),
+	}
+	return e2eutil.WaitUntilPodsReady(f.KubeCli, 1, f.KubeNS, 30*time.Second, lo)
```
hongchaodeng: Let's have the method first.
Change the test in another PR.

pkg/util/constants/constants.go
hasbro17: How about `/status/ready`?

hasbro17:
@xiang90 I'm using the name `/readyz` for the endpoint which seems to be the convention for Kubernetes since the liveliness probe is named `/healthz`.
PTAL to see if this is what you had in mind for the readiness probe.

cmd/operator/main.go
```diff
@@ -151,6 +153,9 @@ func main() {
 		},
 	}
 
+	http.HandleFunc("/readyz", probe.ReadyzHandler)
+	go http.ListenAndServe(constants.HTTPServerListenAddr, nil)
```
xiang90: probably we should make the listen port to be configurable.

pkg/controller/controller.go
```diff
@@ -31,6 +31,7 @@ import (
 	"github.com/coreos/etcd-operator/pkg/util/k8sutil"
 
 	"github.com/Sirupsen/logrus"
+	"github.com/coreos/etcd-operator/pkg/util/probe"
```
xiang90: move this closer to line31

pkg/util/probe/readyz.go
```diff
@@ -0,0 +1,38 @@
...
+
+func SetReady() {
+	ready = true
+}
+
+func SetUnready() {
+	ready = false
+}
+
+// ReadyzHandler write back the appropriate HTTP response code based on the OperatorReady variable
```
xiang90: change appropriate to something more specific?

cmd/operator/main.go
```diff
@@ -151,6 +153,9 @@ func main() {
 		},
 	}
 
+	http.HandleFunc("/readyz", probe.ReadyzHandler)
```
xiang90: define readyz as a const?

xiang90:
lgtm

pkg/util/constants/constants.go
```diff
@@ -28,4 +28,6 @@ const (
 	PVProvisionerGCEPD  = "kubernetes.io/gce-pd"
 	PVProvisionerAWSEBS = "kubernetes.io/aws-ebs"
 	PVProvisionerNone   = "none"
+
+	HTTPServerListenAddr = "0.0.0.0:8080"
```
hongchaodeng: Just put this to where the listen happens.

Also listen to `127.0.0.1`? It is easy to open it up later, but hard to limit it back.

hasbro17: I tried listening to `127.0.0.1` and it seems the pod never becomes ready. 
When the kubelet sends a probe to the HTTP endpoint it will send it on the pod IP. Listening to `0.0.0.0` means being reachable from the outside on the IP addresses of all network interfaces of the pod(including the pod IP).
Whereas `127.0.0.1` means listening to requests only directed to localhost from within the pod.
So I've kept it at `0.0.0.0`.

pkg/util/probe/readyz.go
```diff
@@ -0,0 +1,42 @@
...
+
+var (
+	ready = false
+)
+
+func SetReady() {
+	ready = true
+}
+
+func SetUnready() {
```
hongchaodeng: Remove unused method?

pkg/util/probe/readyz.go
```diff
@@ -0,0 +1,42 @@
...
+const (
+	ReadyzEndpoint = "/readyz"
+)
+
+var (
+	ready = false
+)
+
+func SetReady() {
+	ready = true
```
hongchaodeng: Locking is needed.

hasbro17: Do we need locking?
Currently SetReady() is called when the controller's go routine starts watching the cluster.etcd resource to create cluster events on the event channel.
Will there ever be two controller go routines trying to watch the cluster resource?
Also even if there were two controller go-routines, the first of them to start watching the cluster.etcd resource can safely set the state to `ready=true` since the operator would be ready by that point. We don't go back to `ready=false` ever again so there shouldn't be an issue right?

Although if you want to be safe we can just add locking.

hasbro17:
@hongchaodeng I've added locking for SetReady() although I'm not sure if we really need it right now. PTAL.

cmd/operator/main.go
```diff
@@ -76,6 +79,7 @@ func init() {
 	flag.StringVar(&awsConfig, "backup-aws-config", "",
 		"The name of the kube configmap object that stores the AWS config file. The file name must be 'config'.")
 	flag.StringVar(&s3Bucket, "backup-s3-bucket", "", "The name of the AWS S3 bucket to store backups in.")
+	flag.StringVar(&listenPort, "listen-port", "8080", "The port number on which the HTTP server will listen to")
```
xiang90: probably change this to listen addr, which includes both ip and port.

pkg/util/probe/readyz.go
```diff
@@ -0,0 +1,44 @@
...
+
+package probe
+
+import (
+	"net/http"
+	"sync"
+)
+
+const (
+	ReadyzEndpoint = "/readyz"
```
xiang90: HTTPReadyzEndpoint

pkg/util/probe/readyz.go
```diff
@@ -0,0 +1,44 @@
...
+
+func SetReady() {
+	mu.Lock()
+	defer mu.Unlock()
+	ready = true
+}
+
+// ReadyzHandler writes back the HTTP status code 200 if the operator is ready, and 500 otherwise
+func ReadyzHandler(w http.ResponseWriter, r *http.Request) {
+	if ready {
```
hongchaodeng: > I've added locking for SetReady() although I'm not sure if we really need it right now.

You need to lock it because this is reading. Please add a locking for the read. And don't lock the whole method.

pkg/util/probe/readyz.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+var (
+	mu    sync.Mutex
+	ready = false
+)
+
+func SetReady() {
+	mu.Lock()
+	defer mu.Unlock()
```
hongchaodeng: Don't use defer.
Just put it after.

hongchaodeng:
Also change the commit message and title

pkg/util/probe/readyz.go
```diff
@@ -0,0 +1,47 @@
...
+	ready = true
+	mu.Unlock()
+}
+
+// ReadyzHandler writes back the HTTP status code 200 if the operator is ready, and 500 otherwise
+func ReadyzHandler(w http.ResponseWriter, r *http.Request) {
+	mu.Lock()
+	isReady := ready
+	mu.Unlock()
+	if isReady {
```
hasbro17: @hongchaodeng is this what you meant for locking on read?

hongchaodeng: Right

hongchaodeng:
LGTM

Issue #1153, opened 2017-06-01
upgrade test is flaky


hongchaodeng:
Upgrade tests has been flaky. One third of runs failed.

Previous issue:
- [x] https://github.com/coreos/etcd-operator/issues/1076
- [x] https://github.com/coreos/etcd-operator/issues/1151
- [ ] https://github.com/coreos/etcd-operator/issues/1152
- [ ] https://github.com/coreos/etcd-operator/issues/1186
- [x] https://github.com/coreos/etcd-operator/issues/1204

Create this issue to sum up, track all progress, discussions.

hongchaodeng:
I'm stress running each test to see if I could reproduce.

hongchaodeng:
We still have high volumes of upgrade flakes: https://jenkins-etcd.prod.coreos.systems/view/operator/job/etcd-operator-upgrade-stress/

hongchaodeng:
Closing. We have done a lot of work before and it is more stable now

Issue #1076, opened 2017-05-08
upgrade flake: timeout waiting old image pod deleted


hasbro17:
Currently we use a timeout of 30s in `WaitPodsWithImageDeleted()` in etcd operator upgrade:

```go
f.KubeCli.AppsV1beta1().Deployments(f.KubeNS).Patch(d.Name, types.StrategicMergePatchType, patchData)
...
e2eutil.WaitPodsWithImageDeleted(f.KubeCli, f.KubeNS, f.OldImage, 30*time.Second, lo)
```

There has been at least 1 instance of this causing a timeout failure in the upgrade test:
```
cluster: "jenkins"
OPERATOR_IMAGE: gcr.io/coreos-k8s-scale-testing/etcd-operator:21355118f9527430d35249b8234b2aeb76ae3bd9
TEST_NAMESPACE: e2e-etcd-operator-upgrade-19

--- FAIL: TestResize (90.58s)
	util.go:91: 2017-05-08 16:10:44.859024118 +0000 UTC created etcd cluster: upgrade-test-1dkfg
	util.go:91: 2017-05-08 16:10:44.89568783 +0000 UTC waiting size (3), etcd pods: names ([]), nodes ([])
	util.go:91: 2017-05-08 16:10:54.896758238 +0000 UTC waiting size (3), etcd pods: names ([upgrade-test-1dkfg-0000]), nodes ([gke-jenkins-default-pool-8bb205c4-rjb4])
	util.go:91: 2017-05-08 16:11:04.896635429 +0000 UTC waiting size (3), etcd pods: names ([upgrade-test-1dkfg-0000 upgrade-test-1dkfg-0001]), nodes ([gke-jenkins-default-pool-8bb205c4-rjb4 gke-jenkins-default-pool-8bb205c4-k5tk])
	util.go:91: 2017-05-08 16:11:14.897175996 +0000 UTC waiting size (3), etcd pods: names ([upgrade-test-1dkfg-0000 upgrade-test-1dkfg-0001 upgrade-test-1dkfg-0002]), nodes ([gke-jenkins-default-pool-8bb205c4-rjb4 gke-jenkins-default-pool-8bb205c4-k5tk gke-jenkins-default-pool-8bb205c4-xmgs])
	upgrade_test.go:59: still failing after 6 retries
	util.go:91: 2017-05-08 16:12:05.088677594 +0000 UTC waiting pod (upgrade-test-1dkfg-0000) to be deleted.
[...]

PR #1174, opened 2017-06-06
upgradetest: add readiness probe to wait for operator pod

hasbro17:
The upgradetest framework will now check and wait until the operator pod is ready via the readyz probe, for both `CreateOperator()` and `UpgradeOperator()`.
This way each test would not have to wait longer than necessary for the operator to first become ready while checking for cluster size via `WaitUntilSizeReached()`.

Stress testing for notably flaky tests in #1153

### DIFF
--- a/test/e2e/upgradetest/framework/framework.go
+++ b/test/e2e/upgradetest/framework/framework.go
@@ -15,16 +15,19 @@
 package framework
 
 import (
+	"fmt"
 	"os"
 	"time"
 
 	"github.com/coreos/etcd-operator/pkg/util/k8sutil"
+	"github.com/coreos/etcd-operator/pkg/util/probe"
 	"github.com/coreos/etcd-operator/test/e2e/e2eutil"
 
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/pkg/api/v1"
 	appsv1beta1 "k8s.io/client-go/pkg/apis/apps/v1beta1"
@@ -77,6 +80,13 @@ func (f *Framework) CreateOperator() error {
 			Namespace: f.KubeNS,
 		},
 		Spec: appsv1beta1.DeploymentSpec{
+			Strategy: appsv1beta1.DeploymentStrategy{
+				Type: appsv1beta1.RollingUpdateDeploymentStrategyType,
+				RollingUpdate: &appsv1beta1.RollingUpdateDeployment{
+					MaxUnavailable: &intstr.IntOrString{Type: intstr.Int, IntVal: 1},
+					MaxSurge:       &intstr.IntOrString{Type: intstr.Int, IntVal: 1},
+				},
+			},
 			Selector: &metav1.LabelSelector{MatchLabels: selector},
 			Template: v1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
@@ -98,12 +108,30 @@ func (f *Framework) CreateOperator() error {
 								ValueFrom: &v1.EnvVarSource{FieldRef: &v1.ObjectFieldSelector{FieldPath: "metadata.name"}},
 							},
 						},
+						ReadinessProbe: &v1.Probe{
+							Handler: v1.Handler{
+								HTTPGet: &v1.HTTPGetAction{
+									Path: probe.HTTPReadyzEndpoint,
+									Port: intstr.IntOrString{Type: intstr.Int, IntVal: 8080},
+								},
+							},
+							InitialDelaySeconds: 3,
+							PeriodSeconds:       3,
+						},
 					}},
 				},
 			},
 		},
 	}
 	_, err := f.KubeCli.AppsV1beta1().Deployments(f.KubeNS).Create(d)
+	if err != nil {
+		return fmt.Errorf("failed to create deployment: %v", err)
+	}
+
+	lo := metav1.ListOptions{
+		LabelSelector: labels.SelectorFromSet(map[string]string{"name": "etcd-operator"}).String(),
+	}
+	err = e2eutil.WaitUntilPodReady(f.KubeCli, f.KubeNS, lo, 30*time.Second)
 	return err
 }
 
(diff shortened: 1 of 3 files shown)


hasbro17:
The deployment strategy has been changed to `RecreateDeploymentStrategyType` since the previous default strategy:
```
strategy:
      rollingUpdate:
        maxSurge: 25%
        maxUnavailable: 25%
      type: RollingUpdate
```
means `maxUnavailable=0` which will not allow for the older image pod to be terminated/unavailable until the new image pod becomes available. This blocks the newer pod from becoming ready since the older one is still holding the lock for leader election since it is not being terminated.

hongchaodeng:
Why not change Maxunavailable to 1? What's the justification to use recreate strategy instead of that ?

test/e2e/upgradetest/framework/framework.go
```diff
@@ -135,7 +162,14 @@ func (f *Framework) UpgradeOperator() error {
 		LabelSelector: labels.SelectorFromSet(map[string]string{"name": "etcd-operator"}).String(),
 	}
 	_, err = e2eutil.WaitPodsWithImageDeleted(f.KubeCli, f.KubeNS, f.OldImage, 30*time.Second, lo)
-	return err
+	if err != nil {
+		return fmt.Errorf("failed to wait for pod with old image to get deleted: %v", err)
+	}
+	podName, err := e2eutil.WaitUntilPodReady(f.KubeCli, f.KubeNS, lo, 30*time.Second)
+	if err != nil {
```
hongchaodeng: If pod name is not used outside but just part of the error message, why not encapsulate inside the method

hongchaodeng:
Why does the test fail ? is this some flake in the e2e test? If so we should create an issue for it

hongchaodeng:
Seems like an e2e flake due to DNS problem. Filed issue: https://github.com/coreos/etcd-operator/issues/1175

test/e2e/upgradetest/framework/framework.go
```diff
@@ -77,6 +80,9 @@ func (f *Framework) CreateOperator() error {
 			Namespace: f.KubeNS,
 		},
 		Spec: appsv1beta1.DeploymentSpec{
+			Strategy: appsv1beta1.DeploymentStrategy{
+				Type: appsv1beta1.RecreateDeploymentStrategyType,
```
hasbro17: @hongchaodeng When the strategy type is rolling update, with `maxUnavailable=1` and number of replicas is 1, it almost means the same thing as RecreateDeploymentStrategy. Since the old pod will necessarily get terminated before the new one can become ready(with our implementation of readyz). So I just thought it was cleaner to use Recreate.

However since it works for `maxUnavailable=1` as well I'm going to change it to that and stress test again.

hongchaodeng:
LGTM

I modified top comment it doesn't fix https://github.com/coreos/etcd-operator/issues/1153 yet. 
It fixes https://github.com/coreos/etcd-operator/issues/1076.

hongchaodeng:
We need to stress test to see if other tests will still fail. They are probably caused by the same issue.

test/e2e/upgradetest/framework/framework.go
```diff
@@ -98,12 +108,30 @@ func (f *Framework) CreateOperator() error {
...
 	}
 	_, err := f.KubeCli.AppsV1beta1().Deployments(f.KubeNS).Create(d)
+	if err != nil {
+		return fmt.Errorf("failed to create deployment: %v", err)
+	}
+
+	lo := metav1.ListOptions{
+		LabelSelector: labels.SelectorFromSet(map[string]string{"name": "etcd-operator"}).String(),
+	}
+	err = e2eutil.WaitUntilPodReady(f.KubeCli, f.KubeNS, lo, 30*time.Second)
```
hongchaodeng: This is wrong. We shouldn't Wait() inside CreateDeployment(). They are different methods.

test/e2e/upgradetest/framework/framework.go
```diff
@@ -98,12 +108,30 @@ func (f *Framework) CreateOperator() error {
...
 							},
 						},
+						ReadinessProbe: &v1.Probe{
+							Handler: v1.Handler{
+								HTTPGet: &v1.HTTPGetAction{
+									Path: probe.HTTPReadyzEndpoint,
+									Port: intstr.IntOrString{Type: intstr.Int, IntVal: 8080},
+								},
+							},
+							InitialDelaySeconds: 3,
```
hongchaodeng: Why use (3, 3) ?

test/e2e/upgradetest/framework/framework.go
hasbro17: Well if we don't wait inside `CreateDeployment()` then we'll have to call `WaitUntilPodReady ()` in each test case right after `CreateDeployment()`.
This was just to reduce the code duplication for each test. I don't think having it here will give us the wrong behavior but I can move it outside in each test.

Then `WaitUntilPodReady()` would just replace `WaitEtcdTPRReady()` to address #1176 as well. Is that what you had in mind?

test/e2e/upgradetest/framework/framework.go
hasbro17: No particular reason. Anywhere between 3-5s seemed reasonable since my stress tests were passing earlier with this. Did you want to make it longer?

hongchaodeng: I think the period to check readiness could be larger, like 10s?

test/e2e/upgradetest/framework/framework.go
hongchaodeng: >  I don't think having it here will give us the wrong behavior

It is wrong behavior. what if this wait fail? Who will cleanup the deployment?

hongchaodeng: > Then WaitUntilPodReady() would just replace WaitEtcdTPRReady() to address #1176 as well. Is that what you had in mind?

Yes.

test/e2e/upgradetest/framework/framework.go
```diff
@@ -98,12 +108,30 @@ func (f *Framework) CreateOperator() error {
...
 			},
 		},
 	}
 	_, err := f.KubeCli.AppsV1beta1().Deployments(f.KubeNS).Create(d)
+	if err != nil {
+		return fmt.Errorf("failed to create deployment: %v", err)
+	}
+
+	lo := metav1.ListOptions{
+		LabelSelector: labels.SelectorFromSet(map[string]string{"name": "etcd-operator"}).String(),
```
hongchaodeng: We should also refactor the code to unify all this label selectors in upgrade test

Issue #1176, opened 2017-06-07
upgrade test: Use readiness to replace WatiEtcdTPRReady()


hongchaodeng:

hongchaodeng:
Somehow related: https://github.com/coreos/etcd-operator/pull/1174/files#r120730523

hasbro17:
addressed via #1180

Issue #1179, opened 2017-06-08
Add readiness probe to example deployment and e2e test


hongchaodeng:

akauppi:
I am looking for a readiness endpoint, for running `etcd-operator` (0.3.2). Is there something I can already place in `deployment.yaml`? I don't know how important a readiness probe for the operator would be, but it feels like the "right thing" to have one.

PR #1180, opened 2017-06-08
upgradetest: replace WaitEtcdTPR with readiness

hasbro17:
Fix for #1176 

Also moved `WaitUntilPodReady()` outside from `CreateOperator()` so that the operator can be safely cleaned up incase a test fails due `WaitUntilPodReady()`.

### DIFF
--- a/test/e2e/e2eutil/wait_util.go
+++ b/test/e2e/e2eutil/wait_util.go
@@ -240,25 +240,3 @@ func waitPodsDeleted(kubecli kubernetes.Interface, namespace string, timeout tim
 	})
 	return pods, err
 }
-
-// WaitUntilPodReady will wait until the first pod selected by the label 'lo' is ready. So this should be called for lablels that only select one pod.
-func WaitUntilPodReady(kubecli kubernetes.Interface, namespace string, lo metav1.ListOptions, timeout time.Duration) error {
-	var podName string
-	err := retryutil.Retry(5*time.Second, int(timeout/(5*time.Second)), func() (bool, error) {
-		podList, err := kubecli.CoreV1().Pods(namespace).List(lo)
-		if err != nil {
-			return false, err
-		}
-		if len(podList.Items) > 0 {
-			podName = podList.Items[0].Name
-			if k8sutil.IsPodReady(&podList.Items[0]) {
-				return true, nil
-			}
-		}
-		return false, nil
-	})
-	if err != nil {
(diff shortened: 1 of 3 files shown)


test/e2e/upgradetest/upgrade_test.go
```diff
@@ -38,8 +39,12 @@ func TestResize(t *testing.T) {
 			t.Fatal(err)
 		}
 	}()
-	if err = k8sutil.WaitEtcdTPRReady(testF.KubeCli.CoreV1().RESTClient(), 3*time.Second, 30*time.Second, testF.KubeNS); err != nil {
-		t.Fatalf("failed to see cluster TPR get created in time: %v", err)
+
+	lo := metav1.ListOptions{
```
hongchaodeng: Is it better to abstract this to a comprehensive method?
Like WaitOperatorReady() ?

hasbro17: Yeah that's a good idea now that it's duplicated for every test case.

test/e2e/e2eutil/wait_util.go
```diff
@@ -241,9 +241,12 @@ func waitPodsDeleted(kubecli kubernetes.Interface, namespace string, timeout tim
 	return pods, err
 }
 
-// WaitUntilPodReady will wait until the first pod selected by the label 'lo' is ready. So this should be called for lablels that only select one pod.
-func WaitUntilPodReady(kubecli kubernetes.Interface, namespace string, lo metav1.ListOptions, timeout time.Duration) error {
+// WaitUntilOperatorReady will wait until the first pod selected for the label name=etcd-operator is ready.
+func WaitUntilOperatorReady(kubecli kubernetes.Interface, namespace string, timeout time.Duration) error {
```
hongchaodeng: Let's just make it a method of upgrade test framework. It's only used there right now

hasbro17: Sure, although later when we address #1179  and use readiness probe to replace `WaitEtcdTPRReady()` inside `SetupEtcdOperator()` for the e2e framework, then we will have to make `WaitUntilOperatorReady()` common to both frameworks.

hongchaodeng: Let's deal with that later. We need to focus on upgrade test stability right now.

hongchaodeng:
LGTM

PR #1181, opened 2017-06-08
upgradetest: change readiness probe period

hasbro17:
Period increased to 10s to reduce polling frequency of readiness probe.

[skip ci]


hongchaodeng:
lgtm

PR #1202, opened 2017-06-15
Build proxy support

mtau11:
The build script wasn't working for me because the http proxy information wasn't being propagated to the docker build environment (and the Dockerfile was running apk). This pull request will pass the proxy if set as a build-arg


hongchaodeng:
Closing this.
It's possible to expose ENV instead of passing flags: https://docs.docker.com/engine/admin/systemd/#httphttps-proxy

PR #1207, opened 2017-06-20
upgradetest: wait for killed members to be deleted

### LINKED ISSUE #1204: upgrade_test: pod deleted with zero grace period seems to remain for a bit more
So in our test, we kill pod first and then wait for it to recover to given size.
The pod to kill was deleted with zero grace period. Our expectation is that it should be gone right away. However, it still remains for a bit more.
```
    	util.go:91: 2017-06-15 04:20:08.444585639 +0000 UTC backup pod (upgrade-test-v5tgx-backup-sidecar-431216634-vdms5) is running
    	util.go:91: 2017-06-15 04:20:34.067510096 +0000 UTC waiting size (3), healthy etcd members: names ([upgrade-test-v5tgx-0000 upgrade-test-v5tgx-0001 upgrade-test-v5tgx-0002])
    	upgrade_test.go:406: failed to get backup pod: pods "upgrade-test-v5tgx-0000" not found
```
This violates our assumptions.

hasbro17:
Since `WaitUntilSizeReached()` looks at the TPR's `Status.Members.Ready` field to calculate the size of the of the cluster, `WaitUntilMembersWithNamesDeleted()` waits until the pods killed by `KillMembers()` are no longer present in the ready members field.

~fixes #1204~
Note: This can't be used after `KillMembers()` in `TestDisasterRecovery` because disaster recovery recreates members with the same names as the original cluster. So looking at the TPR's ready members there is no way to tell which members are from the original cluster and which ones are from the recreated cluster. So we can't wait for the old pods to be deleted just by looking at ready members.

### DIFF
--- a/test/e2e/e2eutil/wait_util.go
+++ b/test/e2e/e2eutil/wait_util.go
@@ -113,6 +113,43 @@ func waitSizeReachedWithAccept(t *testing.T, kubeClient kubernetes.Interface, si
 	return names, nil
 }
 
+func WaitUntilMembersWithNamesDeleted(t *testing.T, kubeClient kubernetes.Interface, timeout time.Duration, cl *spec.Cluster, targetNames ...string) ([]string, error) {
+	var remaining []string
+	err := retryutil.Retry(10*time.Second, int(timeout/(10*time.Second)), func() (done bool, err error) {
+		currCluster, err := k8sutil.GetClusterTPRObject(kubeClient.CoreV1().RESTClient(), cl.Metadata.Namespace, cl.Metadata.Name)
+		if err != nil {
+			return false, err
+		}
+
+		readyMembers := currCluster.Status.Members.Ready
+		remaining = nil
+		for _, name := range targetNames {
+			if presentIn(name, readyMembers) {
+				remaining = append(remaining, name)
+			}
+		}
+
+		LogfWithTimestamp(t, "waiting on members (%v) to be deleted", remaining)
+		if len(remaining) != 0 {
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	return remaining, nil
+}
+
+func presentIn(a string, list []string) bool {
+	for _, l := range list {
+		if a == l {
+			return true
+		}
+	}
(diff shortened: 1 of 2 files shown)


hongchaodeng:
I remembered there are two tests that waits for the size to reduce in e2e test.
This duplicate functionality might be better replaced by this util?

hasbro17:
Right, I think you mean 
TestRestartOperator: https://github.com/coreos/etcd-operator/blob/master/test/e2e/e2eslow/disruptive_test.go#L50-L55
and
TestPauseControl: https://github.com/coreos/etcd-operator/blob/master/test/e2e/basic_test.go#L85-L90

In those cases, we're using `WaitUntilPodSizeReached()` to directly look at the pods size to first  reduce, and then increase back to 3.

For upgradetests we can't do the same thing by waiting for the etcd-cluster size to reduce, because the operator will be creating new members at the same time. Unlike the above cases where the operator is not running.
So if we're waiting for the size to reduce from 3->2 then that might not happen as a new pod could be created before the killed member's pod is fully terminated. 
So we need to wait specifically for the etcd pods that were killed to be completely deleted.

hongchaodeng:
Yeah.
I mean we should change that as well.

hongchaodeng:
Using ready members field seems better option

hasbro17:
@hongchaodeng For TestRestartOperator and TestPauseControl I don't think we can use the TPR's ready members to check for size. 
In those tests the operator is paused/killed and so it can't update the TPR to indicate that the ready members size has reduced to 2. That is why we check the pod sizes directly in those two cases.

hongchaodeng:
I see.
LGTM for this PR

hasbro17:
On second thought, this will not fix #1204 since that issue is specific to `TestDisasterRecovery`. See note in the first comment. Still thinking of a fix for that.

This PR does make `TestHealOneMemberForOldCluster` more deterministic by properly waiting for the killed member to be deleted.

@hongchaodeng do you want fix #1204 in this PR, or merge this first and address that later in a separate PR.

hongchaodeng:
Let's address that in another RP

PR #1208, opened 2017-06-20
upgradetest: wait for pod size instead of ready member size

hasbro17:
fix for #1204
Using `WaitUntilPodSizeReached ()` to check the pods size instead of  `WaitUntilSizeReached()` to check the ready members, should prevent `TestDisasterRecovery` from mistakenly querying one of the terminating pods.


hongchaodeng:
lgtm

PR #1210, opened 2017-06-21
pkg/cluster: make cluster struct printing more readable

### LINKED ISSUE #1199: print human readable update information for spec
When the spec gets updated, we print out something like

```
time="2017-06-15T14:40:21Z" level=info msg="spec update: from: {3 3.1.8 false 0xc420755f20 <nil> <nil> 0xc420191d10 <nil>} to: {3 3.1.8 false 0xc42095f200 0xc42078e880 <nil> 0xc4202abd00 <nil>}" cluster-name=kube-etcd pkg=cluster
```

this is not readable at all.

hasbro17:
fixes #1199 

The cluster struct's spec/status fields are encoded and logged in json format.
The new output looks like the following:
Cluster creation:
```
time="2017-06-21T19:16:11Z" level=info msg="creating cluster with
Spec:
{
    "size": 3,
    "baseImage": "quay.io/coreos/etcd",
    "version": "3.1.7"
}
Status:
{
    "phase": "Creating",
    "reason": "",
    "controlPaused": false,
    "conditions": null,
    "size": 0,
    "members": {},
    "currentVersion": "",
    "targetVersion": ""
}" cluster-name=example-etcd-cluster pkg=cluster
time="2017-06-21T19:16:11Z" level=info msg="cluster created with seed member (example-etcd-cluster-0000)" cluster-name=example-etcd-cluster pkg=cluster
```

Spec Update:
```
time="2017-06-21T19:17:07Z" level=info msg="Finish reconciling" cluster-name=example-etcd-cluster pkg=cluster
time="2017-06-21T19:17:09Z" level=info msg="spec update:
Old:
{
    "size": 3,
    "baseImage": "quay.io/coreos/etcd",
    "version": "3.1.7"
}
New:
{
    "size": 3,
    "baseImage": "quay.io/coreos/etcd",
    "version": "3.1.8",
    "pod": {
        "labels": {
            "tmp": "example"
        },
        "resources": {}
    }
}" cluster-name=example-etcd-cluster pkg=cluster
time="2017-06-21T19:17:17Z" level=info msg="Start reconciling" cluster-name=example-etcd-cluster pkg=cluster
time="2017-06-21T19:17:17Z" level=info msg="upgrading the etcd member example-etcd-cluster-0000 from 3.1.7 to 3.1.8" cluster-name=example-etcd-cluster pkg=cluster
```

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -16,6 +16,7 @@ package cluster
 
 import (
 	"crypto/tls"
+	"encoding/json"
 	"fmt"
 	"math"
 	"reflect"
@@ -177,7 +178,7 @@ func (c *Cluster) create() error {
 	if err := c.updateTPRStatus(); err != nil {
 		return fmt.Errorf("cluster create: failed to update cluster phase (%v): %v", spec.ClusterPhaseCreating, err)
 	}
-	c.logger.Infof("creating cluster with Spec (%#v), Status (%#v)", c.cluster.Spec, c.cluster.Status)
+	c.logClusterCreation()
 
 	c.gc.CollectCluster(c.cluster.Metadata.Name, c.cluster.Metadata.UID)
 
@@ -269,7 +270,7 @@ func (c *Cluster) run(stopC <-chan struct{}) {
 					break
 				}
 				// TODO: we can't handle another upgrade while an upgrade is in progress
-				c.logger.Infof("spec update: from: %v to: %v", c.cluster.Spec, event.cluster.Spec)
+				c.logSpecUpdate(event.cluster.Spec)
 
 				ob, nb := c.cluster.Spec.Backup, event.cluster.Spec.Backup
 				c.cluster = event.cluster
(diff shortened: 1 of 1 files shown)


hongchaodeng:
Amazing!
LGTM

pkg/cluster/cluster.go
```diff
@@ -177,7 +178,16 @@ func (c *Cluster) create() error {
 	if err := c.updateTPRStatus(); err != nil {
 		return fmt.Errorf("cluster create: failed to update cluster phase (%v): %v", spec.ClusterPhaseCreating, err)
 	}
-	c.logger.Infof("creating cluster with Spec (%#v), Status (%#v)", c.cluster.Spec, c.cluster.Status)
+
+	specBytes, err := json.MarshalIndent(c.cluster.Spec, "", "    ")
```
xiang90: probably move this to a func? like logCreation?

pkg/cluster/cluster.go
```diff
@@ -269,7 +279,15 @@ func (c *Cluster) run(stopC <-chan struct{}) {
 					break
 				}
 				// TODO: we can't handle another upgrade while an upgrade is in progress
-				c.logger.Infof("spec update: from: %v to: %v", c.cluster.Spec, event.cluster.Spec)
+				oldSpec, err := json.MarshalIndent(c.cluster.Spec, "", "    ")
```
xiang90: same here: log spec change?

pkg/cluster/cluster.go
```diff
@@ -269,7 +279,15 @@ func (c *Cluster) run(stopC <-chan struct{}) {
...
+				if err != nil {
+					c.logger.Warnf("failed to marshal cluster spec: %v", err)
```
xiang90: probably just panic here? this is an unexpected error, right?

xiang90:
this approach looks good to me. some minor issues to be fixed. thank you!

pkg/cluster/cluster.go
hasbro17: Yes, but it's not supposed to be fatal since this marshalling is only for logging purposes. So don't want to stop the operator via panic.

xiang90:
for the creation one, status is not needed i think.

pkg/cluster/cluster.go
xiang90: warnf is too weak anyway. probably bump it to errorf.

hongchaodeng:
> for the creation one, status is not needed i think.

This prints spec only

xiang90:
lgtm!

Issue #1211, opened 2017-06-22
etcd-cluster version upgrade fails as kubelet fails to restart containers 


hasbro17:
Upgrading the etcd cluster by changing the TPR's `spec.version` field causes all etcd pods to change their status from `Running` to `Error`. This eventually causes the operator to remove all the members it perceives as being dead and going into disaster recovery mode.

#### Steps to reproduce:
Kubernetes version: 1.6.4
- Create the etcd-operator using the [example/deployment.yaml](https://github.com/coreos/etcd-operator/blob/master/example/deployment.yaml)
- Create a simple cluster using [example/example-etcd-cluster.yaml](https://github.com/coreos/etcd-operator/blob/master/example/example-etcd-cluster.yaml)
- Wait until all 3 members are ready, then edit and update the cluster TPR's `spec.version` field to `3.1.9`
- The pods will eventually change their status to `Error` and get removed by the operator

#### Logs and outputs

Watching the etcd-pods change state from `Running` to `Error`. This output was possible by pausing the operator before it could remove any of the dead members.
```bash
NAME                             READY     STATUS    RESTARTS   AGE
etcd-operator-3920290006-b43n3   1/1       Running   0          1m
example-etcd-cluster-0000   1/1       Running   0         43s
example-etcd-cluster-0001   1/1       Running   0         27s
example-etcd-cluster-0002   1/1       Running   0         11s
example-etcd-cluster-0000   1/1       Running   0         1m
example-etcd-cluster-0001   1/1       Running   0         1m
example-etcd-cluster-0002   1/1       Running   0         1m
example-etcd-cluster-0000   0/1       Error     0         1m
example-etcd-cluster-0000   0/1       Error     0         1m
example-etcd-cluster-0001   0/1       Error     0         1m
example-etcd-cluster-0002   0/1       Error     0         1m
example-etcd-cluster-0002   0/1       Error     0         1m
```

etcd-operator logs after the spec update:
```
time="2017-06-22T18:04:39Z" level=info msg="Finish reconciling" cluster-name=example-etcd-cluster pkg=cluster
time="2017-06-22T18:04:46Z" level=info msg="spec update: from: {3 3.1.8 false <nil> <nil> <nil> <nil> <nil>} to: {3 3.1.9 false <nil> <nil> <nil> <nil> <nil>}" cluster-name=example-etcd-cluster pkg=cluster
[...]

PR #1217, opened 2017-06-23
e2eutil/wait_util: check container status version instead of TPR version

hasbro17:
Fix the e2e test `TestEtcdUpgrade` to detect the version upgrade failure from #1211 by checking that the etcd pods are Running and their container status images are updated to the desired version.

With no fix for #1211 the test `TestEtcdUpgrade` should now fail.


hasbro17:
@etcd-bot retest this please

xiang90:
lgtm

Issue #1112, opened 2017-05-18
Can we able to create AWS IAM rolebased etcd-operator?


ssubramanian123:
Hi,

I am trying to create jenkins pipeline to deploy vault. As part of etcd-operator need below three steps which involves AWS credentials and config. Since storing AWS credentials in stash is not secure (not good practice), looking for alternate option. Is there any best option to create pipeline for vault or can we use AWS IAM role ? 
````
kubectl create secret generic aws --from-file=credentials --namespace=vault
kubectl create configmap aws --from-file=config --namespace=vault
kubectl create -f etcd/etcd-operator.yaml --namespace=vault
````
````
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: etcd-operator
  namespace: vault
spec:
  replicas: 1
  template:
    metadata:
      labels:
        name: etcd-operator
    spec:
      containers:
      - name: etcd-operator
        image: quay.io/coreos/etcd-operator:v0.2.6
        command:
        - /usr/local/bin/etcd-operator
        - --backup-aws-secret=aws
        - --backup-aws-config=aws
        - --backup-s3-bucket=my-test-bucket
        env:
        - name: MY_POD_NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace
        - name: MY_POD_NAME
          valueFrom:
            fieldRef:
              fieldPath: metadata.name
````

hongchaodeng:
IAM roles are per node, right?
We better rely on k8s to do cluster level management.
Maybe ask this question on k8s upstream?

ssubramanian123:
@hongchaodeng IAM role for etcd-operator cluster. We are configuring vault backup as S3 with below AWS credentials.. Is there any way to use IAM role instead of secret/config?

Current
````command:
        - /usr/local/bin/etcd-operator
        - --backup-aws-secret=aws
        - --backup-aws-config=aws
        - --backup-s3-bucket=my-test-bucket````

Expectation.. something like
````command:
        - /usr/local/bin/etcd-operator
        - --backup-aws-role=myimarole
        - --backup-s3-bucket=my-test-bucket````

hongchaodeng:
How to create a aws client via IAM?

ssubramanian123:
@hongchaodeng in the above use case, I am referring only credential section. Apologies, still if it is not clear. 

As part of above step, we are creating S3 bucket and using "aws" secret/config accessing that S3 bucket.

[...]

PR #1221, opened 2017-06-27
dont write aws creds to disk and dont require secret

sstarcher:
This accomplishes 2 main goals.
* No longer write AWS credentials to disk
* Allow for the use of AWS IAM instance profiles as an optional feature.  If the secret is not specified it will fall back to the aws credential chain instead of outright failing.

I kept the format of the secret using the AWS credential file and the config file, but since these are no longer being written to disk it would simplify the configuration for users to not use the AWS format and instead call the values out specifically in the k8s secret.

The unit tests were passing, but I need to verify this in my cluster.

### DIFF
--- a/pkg/backup/s3/s3.go
+++ b/pkg/backup/s3/s3.go
@@ -19,8 +19,10 @@ import (
 	"path"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/go-ini/ini"
 )
 
 const (
@@ -38,11 +40,7 @@ type S3 struct {
 // Please refer to http://docs.aws.amazon.com/sdk-for-go/api/aws/session/
 // for setting up credentials and configuration.
 func New(bucket, prefix string) (*S3, error) {
-	return NewFromSessionOpt(bucket, prefix, session.Options{
-		// Setting this is equal to the AWS_SDK_LOAD_CONFIG environment variable was set.
-		// We want to save the work to set AWS_SDK_LOAD_CONFIG=1 outside.
-		SharedConfigState: session.SharedConfigEnable,
-	})
+	return NewFromSessionOpt(bucket, prefix, session.Options{})
 }
 
 func NewFromSessionOpt(bucket, prefix string, so session.Options) (*S3, error) {
@@ -147,4 +145,50 @@ func (s *S3) CopyPrefix(from string) error {
 		}
 	}
 	return nil
+}
+
+// LoadCredentials reads the default profile section from a aws credentials file
+func LoadCredentials(data []byte) (creds *credentials.Value, err error) {
+	config, err := ini.Load(data)
+	if err != nil {
+		return
+	}
+	iniProfile, err := config.GetSection("default")
+	if err != nil {
+		return
+	}
+
+	id, err := iniProfile.GetKey("aws_access_key_id")
+	if err != nil {
+		return
+	}
+
+	secret, err := iniProfile.GetKey("aws_secret_access_key")
+	if err != nil {
+		return
+	}
+
+	// Default to empty string if not found
+	token := iniProfile.Key("aws_session_token")
+
+	return &credentials.Value{
+		AccessKeyID:     id.String(),
+		SecretAccessKey: secret.String(),
+		SessionToken:    token.String(),
+	}, nil
+}
+
+// LoadConfig reads in an aws configuration file data
+func LoadConfig(data []byte) (string, error) {
+	config, err := ini.Load(data)
+	if err != nil {
+		return "", err
+	}
+	iniProfile, err := config.GetSection("default")
+	if err != nil {
+		return "", err
+	}
+
+	region, err := iniProfile.GetKey("region")
+	return region.String(), err
 }
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: abfa1d8c2be45a5826712ed0ec8cde1bde9566525cec1e36d85783e42726f1dc
-updated: 2017-06-19T09:44:03.664783187-07:00
+hash: db22f36a9471df190a07c8920f9a65613807835a0166cfef641d215129da5987
+updated: 2017-06-27T08:26:51.461920388-04:00
 imports:
 - name: bitbucket.org/ww/goautoneg
   version: 75cd24fc2f2c2a2088577d12123ddee5f54e0675
@@ -85,7 +85,7 @@ imports:
 - name: github.com/ghodss/yaml
   version: 73d445a93680fa1a78ae23a5839bad48f32ba1ee
 - name: github.com/go-ini/ini
-  version: 2e44421e256d82ebbf3d4d4fcabe8930b905eff3
+  version: d3de07a94d22b4a0972deb4b96d790c2c0ce8333
 - name: github.com/go-openapi/jsonpointer
   version: 46af16f9f7b149af66e5d1bd010e3574dc06de98
 - name: github.com/go-openapi/jsonreference
--- a/glide.yaml
+++ b/glide.yaml
@@ -19,3 +19,5 @@ import:
 - package: github.com/jpillora/go-ogle-analytics
 - package: golang.org/x/net
 - package: golang.org/x/time
+- package: github.com/go-ini/ini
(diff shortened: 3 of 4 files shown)


hongchaodeng:
@sstarcher 
Thanks for writing the POC.

Currently, we do not need to understand credential and config files' format. However, with this PR, it tries to understand the internal format, e.g. `aws_access_key_id`, `region`.

Additionally, it assumes to parse only specific keys, e.g. `region`. What if there are more keys to parse? Do we keep updating the LoadConfig func() ourselves?

I think both of above things should be done by the SDK itself. Ideally, the SDK should provide a method that has better abstraction:
```Go
func NewAWS(*Credentials, *Config) *Client ...
func ParseCredential([]byte) *Credential ...
func ParseConfig([]byte) *Config ...
```

We should try to look into the SDK and find if there is any way to achieve above abstractions. If not found, we should try to report to the SDK repo instead.

sstarcher:
The only relevant bit is the region.  The way that etcd-operator is interacting with the aws sdk are methods that are normally only relevant to a user interacting on a local computer.  The way you are using it is highly frowned upon.  These secrets should not be hitting the filesystem.

hongchaodeng:
> These secrets should not be hitting the filesystem.

I totally agree. That's the current limitation of the SDK. We should push the SDK to fix it. 
I don't see any reason why SDK shouldn't support parsing data instead of file?

> supporting the specific options that are relevant

This would incur more support burden on our side, which we couldn't afford at the moment.

sstarcher:
Most of the options are not relevant and are not intended for programmatic use.

sstarcher:
If you are not interested in my security changes I can revert them out, but I would like to get in support for the AWS IAM instance profile when the secret is not set.  If not I will have to fork this project which I would prefer not to do.

hongchaodeng:
>  I would like to get in support for the AWS IAM instance profile when the secret is not set.

So this PR is for another purpose? How are they relevant?

sstarcher:
@hongchaodeng if you read my very first comment I said this PR accomplishes 2 things.

hongchaodeng:
First of all, using IAM has nothing to do with parsing the credentials and config. We should discuss them separately.

Second, I do not understand how it could fall back to use the IAM role? etcd operator is run as a pod.

sstarcher:
It can fall back to an IAM role when the pod is running on a cluster in AWS.

hongchaodeng:
> It can fall back to an IAM role when the pod is running on a cluster in AWS.

That would be awesome! We should definitely have that.

ref: 
- http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html
- https://github.com/coreos/etcd-operator/issues/1112

sstarcher:
if #1112 is just trying to use the credentials of the server the following commit should be sufficent.

https://github.com/sstarcher/etcd-operator/commit/b6513fe60790399650f7847f3d085972b06f9916

hongchaodeng:
Yep.
We are testing it. 
If you can also help test it that would be great.

sstarcher:
I'll be testing it out more fully hopefully tomorrow.  I'm currently upgrading my setup.

hongchaodeng:
So I tested it. Here's my finding:
- It can automatically inherit node's IAM and credentials is managed. Awesome!
- Nonetheless, it still needs additional config, like `region`:
```
MissingRegion: could not find region configuration
```

hongchaodeng:
Anything could be automated, we should go for it.
But we would need more design around how to handle the config file.

sstarcher:
I would highly recommend not using the config file and adding the region to https://github.com/coreos/etcd-operator/blob/master/pkg/backup/s3/s3config/s3_ctx.go#L3

Those config files are to be used by users, not by software using the SDK.

The secrets themselves should not be credential files, but should key: values of 
* AccessKeyId
* SecretAccessKey
* SessionToken

I would be more than happy to make those changes.

mikkeloscar:
> Nonetheless, it still needs additional config, like region:

One way to get the region automatically is to read it from ec2 metadata: https://github.com/aws/aws-sdk-go/issues/1103

sstarcher:
@mikkeloscar That would only get you the region of the server itself.  In this case we actually need the region of the bucket.  The s3api does allow you to request what region the bucket is in though, but this is assuming that the credentials have the proper access level.

mikkeloscar:
@sstarcher ah true, I didn't think of that.

hongchaodeng:
> I would highly recommend not using the config file and adding the region to ...
> The secrets themselves should not be credential files, but should key: values of...
> - AccessKeyId
> - SecretAccessKey
> - SessionToken

This go back to https://github.com/coreos/etcd-operator/pull/1221#issuecomment-311419162 . Cluster spec should not understand the format of the config/credentials. This should be done within the SDK.

sstarcher:
@hongchaodeng you can keep saying that, but it's not going to happen.  The AWS SDKs for different languages are all pretty much the same.  They are not going to add your use case.

The method the SDK prefer is for you to programatically set the values you need.  You should not be writing files to disk.

xiang90:
@sstarcher This PR needs to be separated into two:

1. No longer write AWS credentials to disk
2. Allow for the use of AWS IAM instance profiles as an optional feature. If the secret is not specified it will fall back to the aws credential chain instead of outright failing.

> it would simplify the configuration for users to not use the AWS format

No. We want AWS format. Is there a good reason to use something else?

sstarcher:
@xiang90 the AWS format is designed to be used by an end user and not by software.  You only need 1/10th of the settings that are in the AWS configuration files.  They are designed for multi-tenant use and you are specially making edge cases saying in your docs to use `default`.

A better option for everyone involved would be to have configuration for the 4 options you need and can support.

pkg/backup/s3/s3.go
```diff
@@ -148,3 +146,49 @@ func (s *S3) CopyPrefix(from string) error {
...
 	return nil
 }
+
+// LoadCredentials reads the default profile section from a aws credentials file
+func LoadCredentials(data []byte) (creds *credentials.Value, err error) {
+	config, err := ini.Load(data)
+	if err != nil {
+		return
+	}
+	iniProfile, err := config.GetSection("default")
```
xiang90: This looks pretty fragile. We are trying to understand AWS config internal thing here.

Is this the AWS SDK suggested way? Can you point me to any reference?

sstarcher: This code snippet was pulled directly from the AWS SDK.

xiang90:
> the AWS format is designed to be used by an end user and not by software. You only need 1/10th of the settings that are in the AWS configuration files.

Can you point me to any AWS SDK reference for this? Thank you.

pkg/backup/s3/s3.go
sstarcher: https://github.com/aws/aws-sdk-go/blob/a6b7ec5c121b41aec09583889b84fbdd12d3827b/aws/credentials/shared_credentials_provider.go#L79

xiang90: yea. that is bad. too fragile to use internal pkg.

xiang90:
@sstarcher 

> the AWS format is designed to be used by an end user and not by software.

Right. And remember that the current AWS config is set by end user. The operator should be a dumb pipe. It should not even try to interpret AWS config the end user provide.

sstarcher:
@xiang90 if you intent do that your only option is to write the secrets to disk.  If you feel that is the better solution than no changes are needed.

xiang90:
> if you intend do that your only option is to write the secrets to disk

What the your main concern about the secret putting the secret on dis? security? What I said is the optimal solution for the operator. I am open to discuss. If one the benefit clearly overweights its disadvantage, we still can do it.

sstarcher:
My original goal was to support the AWS instance role at the same time I refactored the credentials to not be on disk.  I should of kept my goal more concise.  Yes, security is a concern with writing the credentials to disk.

Using the AWS file format also makes configuration more difficult vs informing the user that these are the 4 options that need to be set.

mikkeloscar:
@sstarcher 

> My original goal was to support the AWS instance role

Do you intend to follow up with a PR for just this? Would be great to have this!

hongchaodeng:
> My original goal was to support the AWS instance role

This intention is very good. We should have it.
It wouldn't be hard. @sstarcher Do you want to work on it? Or do you want us to take care of it?

sstarcher:
I won't have time this week.  Feel free to take care of it. @hongchaodeng 

Having to create a secret to configure the region is still very painful for the user and will require me to fork this repo for it to be easier.

Issue #1205, opened 2017-06-17
enforce on disk action log for self hosted cluster


xiang90:
The self hosted etcd cluster is critical to the health of a kubernetes cluster, thus to the operator itself.

The highest priority thing for us is to ensure the reliability of such a cluster. We want to debug EVERY failure cases.

Sometimes, users have troubles with their self hosted clusters, and when it happens they cannot even access their k8s cluster. It is usually hard for them to get the previous logs, especially for their testing/dev clusters without external logging system hooked up. But we need etcd operator log to debug such cases.

To make our debugging life easier, we should write another set of log down to local disk for self hosted cluster for every action it takes.. This is very common tricks for early stage systems that care about debug-ability. 

/cc @hongchaodeng

xiang90:
how to log:

use a well known host path
*append* to a log file, never truncate

things to log:

- add new pod (with pod manifest)
- remove pod (with pod name)
- spec change (with old and new spec in human readable format)
- changes in running pod (pod failure)
- any API error or failure

/cc @hongchaodeng

coresolve:
💯 agree!

hongchaodeng:
done

PR #1232, opened 2017-06-28
*: audit operator actions to hostPath volume

hasbro17:
Partially addresses #1205 

The operator will now audit the following actions for a self-hosted cluster to a hostPath volume:
- Create an etcd pod
- Delete an etcd pod
- A spec update to the cluster
- Deleting a failed cluster

To enable the auditing a hostPath volume must be mounted to the mountPath `/var/tmp/etcd-operator` as shown below. The hostPath itself can be anything but should ideally be `/var/tmp` so the operator container can write to the volume even when running as non-root.
```
spec:
  containers:
  - name: etcd-operator
    volumeMounts:
    - mountPath: /var/tmp/etcd-operator
      name: audit-volume
  volumes:
  - name: audit-volume
    hostPath:
      path: /var/tmp/
```
The audit file will be named `<cluster-name>.log` .
If no volume is mounted at the desired mountPath the operator will not audit and function normally as before.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -28,6 +28,7 @@ import (
 	"github.com/coreos/etcd-operator/pkg/backup/s3/s3config"
 	"github.com/coreos/etcd-operator/pkg/chaos"
 	"github.com/coreos/etcd-operator/pkg/controller"
+	"github.com/coreos/etcd-operator/pkg/debug"
 	"github.com/coreos/etcd-operator/pkg/garbagecollection"
 	"github.com/coreos/etcd-operator/pkg/util/constants"
 	"github.com/coreos/etcd-operator/pkg/util/k8sutil"
@@ -72,6 +73,7 @@ var (
 
 func init() {
 	flag.BoolVar(&analyticsEnabled, "analytics", true, "Send analytical event (Cluster Created/Deleted etc.) to Google Analytics")
+	flag.StringVar(&debug.DebugFilePath, "debug-logfile-path", "", "only for a self hosted cluster, the path where the debug logfile will be written, recommended to be under: /var/tmp/etcd-operator/debug/ to avoid any issue with lack of write permissions")
 
 	flag.StringVar(&pvProvisioner, "pv-provisioner", constants.PVProvisionerGCEPD, "persistent volume provisioner type")
 	flag.StringVar(&awsSecret, "backup-aws-secret", "",
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/coreos/etcd-operator/pkg/backup/s3/s3config"
+	"github.com/coreos/etcd-operator/pkg/debug"
 	"github.com/coreos/etcd-operator/pkg/garbagecollection"
 	"github.com/coreos/etcd-operator/pkg/spec"
 	"github.com/coreos/etcd-operator/pkg/util/etcdutil"
@@ -66,6 +67,8 @@ type Config struct {
 
 type Cluster struct {
 	logger *logrus.Entry
+	// debug logger for self hosted cluster
+	debugLogger *debug.DebugLogger
 
 	config Config
 
@@ -93,14 +96,20 @@ type Cluster struct {
 
 func New(config Config, cl *spec.Cluster, stopC <-chan struct{}, wg *sync.WaitGroup) *Cluster {
 	lg := logrus.WithField("pkg", "cluster").WithField("cluster-name", cl.Metadata.Name)
+	var debugLogger *debug.DebugLogger
+	if cl.Spec.SelfHosted != nil {
+		debugLogger = debug.New(cl.Metadata.Name)
+	}
+
 	c := &Cluster{
-		logger:  lg,
-		config:  config,
-		cluster: cl,
-		eventCh: make(chan *clusterEvent, 100),
-		stopCh:  make(chan struct{}),
-		status:  cl.Status.Copy(),
-		gc:      garbagecollection.New(config.KubeCli, cl.Metadata.Namespace),
+		logger:      lg,
+		debugLogger: debugLogger,
+		config:      config,
+		cluster:     cl,
+		eventCh:     make(chan *clusterEvent, 100),
+		stopCh:      make(chan struct{}),
+		status:      cl.Status.Copy(),
+		gc:          garbagecollection.New(config.KubeCli, cl.Metadata.Namespace),
 	}
 
 	wg.Add(1)
@@ -481,6 +490,12 @@ func (c *Cluster) removePod(name string) error {
 		if !k8sutil.IsKubernetesResourceNotFoundError(err) {
 			return err
 		}
+		if c.isDebugLoggerEnabled() {
+			c.debugLogger.LogMessage(fmt.Sprintf("pod (%s) not found while trying to delete it", name))
+		}
+	}
+	if c.isDebugLoggerEnabled() {
+		c.debugLogger.LogPodDeletion(name)
 	}
 	return nil
 }
@@ -631,4 +646,15 @@ func (c *Cluster) logSpecUpdate(newSpec spec.ClusterSpec) {
 	for _, m := range strings.Split(string(newSpecBytes), "\n") {
 		c.logger.Info(m)
 	}
+
+	if c.isDebugLoggerEnabled() {
+		c.debugLogger.LogClusterSpecUpdate(string(oldSpecBytes), string(newSpecBytes))
+	}
+}
+
+func (c *Cluster) isDebugLoggerEnabled() bool {
+	if c.cluster.Spec.SelfHosted != nil && c.debugLogger != nil {
+		return true
+	}
+	return false
 }
--- a/pkg/debug/debug_logger.go
+++ b/pkg/debug/debug_logger.go
@@ -1,0 +1,84 @@
+// Copyright 2017 The etcd-operator Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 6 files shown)


hongchaodeng:
Any examples what the audit log looks like?

pkg/cluster/audit.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/Sirupsen/logrus"
+	"k8s.io/client-go/pkg/api/v1"
+)
+
+func NewAuditLogger(cl *spec.Cluster, lg *logrus.Entry) *logrus.Logger {
+	if cl.Spec.SelfHosted == nil {
+		return nil
+	}
+
+	mountPath := "/var/tmp/etcd-operator/"
```
hongchaodeng: There is a `OperatorRoot`

hasbro17:
[These logs](https://gist.github.com/hasbro17/9d10da479a2c024caf508e4973e8ede0) were generated from a non-self hosted cluster creating the audit logs during testing. It was easier to perform more actions on a non-self hosted cluster to generate the audit.

I'm generating more detailed logs for a self-hosted cluster via bootkube at the moment. Will post them here once I'm done.

pkg/cluster/audit.go
```diff
@@ -0,0 +1,98 @@
...
+
+	"github.com/coreos/etcd-operator/pkg/spec"
+	"github.com/coreos/etcd-operator/pkg/util/k8sutil"
+
+	"github.com/Sirupsen/logrus"
+	"k8s.io/client-go/pkg/api/v1"
+)
+
+func NewAuditLogger(cl *spec.Cluster, lg *logrus.Entry) *logrus.Logger {
+	if cl.Spec.SelfHosted == nil {
```
hongchaodeng: We should make this behavior explicit outside:
```
if selfhosted != nil {
  c.AuditLogger = NewAuditLogger()
}
```

pkg/cluster/audit.go
```diff
@@ -0,0 +1,98 @@
...
+		return nil
+	}
+
+	l := logrus.New()
+	l.Out = logFile
+	l.Infof("Starting audit logs for self-hosted etcd cluster: %v", cl.Metadata.Name)
+	return l
+}
+
+func (c *Cluster) auditPodCreation(pod *v1.Pod, podCreationErr error) {
```
hongchaodeng: We should provide an AuditLogger object instead of coupling it as Cluster methods.

pkg/cluster/audit.go
xiang90: this will always be mounted. so we need a flag to enable audit logging now.

xiang90: actually "audit logging" might be a bad name. probably debug logging is more suitable.

pkg/cluster/audit.go
```diff
@@ -0,0 +1,98 @@
...
+
+	mountPath := "/var/tmp/etcd-operator/"
+	_, err := os.Stat(mountPath)
+	if os.IsNotExist(err) {
+		lg.Infof("mountPath(%v) not detected, no auditing will be performed: %v", mountPath, err)
+		return nil
+	}
+	lg.Infof("detected the mountPath(%v): starting to audit operator actions to the mountPath", mountPath)
+
+	fileName := mountPath + cl.Metadata.Name + ".log"
```
xiang90: use path.join

pkg/util/k8sutil/pod_util.go
```diff
@@ -168,3 +169,11 @@ func getPodReadyCondition(status *v1.PodStatus) *v1.PodCondition {
 	}
 	return nil
 }
+
+func GetReadablePodSpec(pod *v1.Pod) (string, error) {
```
xiang90: PodSpecToPrettyJSON

pkg/cluster/audit.go
hasbro17: @xiang90 Actually the `OperatorRoot` filepath is not mounted/created before the debug logger is setup.
The `OperatorRoot` path is created after that by the backup manager during the [cluster setup](https://github.com/coreos/etcd-operator/blob/master/pkg/cluster/cluster.go#L109) by the backup manager while creating the credentials and config files from the aws secret.
https://github.com/coreos/etcd-operator/blob/master/pkg/cluster/backupstorage/s3.go#L35-L39 

Although I agree an explicit flag to enable debugging is better than silently detecting the presence of the mountPath.

My only concern now is that when you mount a hostPath to `/var/tmp/etcd-operator/` for debugging, and if you set up S3 backup, the credentials and config files will be written to the hostPath as well now. Even though they are [deleted later](https://github.com/coreos/etcd-operator/blob/master/pkg/cluster/backupstorage/s3.go#L91) when the cluster is destroyed successfully, just wondering if this is alright.

xiang90: it is OK to put them under the same operator root dir. but we do not want our debug logging to be deleted by any accident.

hasbro17: Oh yeah the debug logging won't be deleted. Just the aws credential and config files setup by the backup manager. Alright then, I guess we're good to share the `OperatorRoot` for debugging.

xiang90: that is fine.

hasbro17:
@xiang90 @hongchaodeng Made the changes requested. The debug logger is now a separate object that appends to a log file for the cluster.
To turn on the debug logging the following need to be true:
- The cluster must be self hosted
- A volume must be mounted at the path `/var/tmp/etcd-operator/debug`
- The etcd-operator flag `debug` must be set to true

The pod spec is not printed for pods added/deleted because that makes the output too verbose.
An update in the cluster spec is logged since that helps make sense of the other actions(pod creation/deletion) that the etcd-operator logs. However that can be removed if not neeeded.

[Logs](https://gist.github.com/hasbro17/0ec68145ea057a4b49be3726ae8e0b27) for a 4 master node self hosted cluster. The following actions were performed:
- Resize the cluster from 1->4 pods after its creation
- Delete one pod manually
- Resize the cluster from 4->3 pods
The messages prepended with `##########` were added by me to indicate certain events like the operator restarting.

xiang90:
> The pod spec is not printed for pods added/deleted because that makes the output too verbose.

i think it is still worth to add even though it can be verbose.

cmd/operator/main.go
```diff
@@ -72,6 +73,7 @@ var (
 
 func init() {
 	flag.BoolVar(&analyticsEnabled, "analytics", true, "Send analytical event (Cluster Created/Deleted etc.) to Google Analytics")
+	flag.BoolVar(&debug.DebugEnabled, "debug", false, "Enable debug logging of operator actions to a hostPath volume mounted at /var/tmp/etcd-operator/")
```
xiang90: enable-debug-file (also you can make this arg to take a actual file path. default can be `/var/tmp/etcd-operator/debug.log`)

operator probably should not know about hostpath thing. it just writes logs to somewhere.

xiang90: well, actually enable-selfhosted-debug-log is better.

pkg/debug/debug_logger.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func New(clusterName string) *DebugLogger {
+	if !DebugEnabled {
+		return nil
+	}
+
+	logger := logrus.WithField("pkg", "debug")
+
+	mountPath := path.Join(constants.OperatorRoot, "debug")
```
xiang90: we can remove this. operator does not have to know anything about the host path mount. 

it just logs to what is configured by the deployment when selfhosted-debug-log is enabled.

pkg/debug/debug_logger.go
```diff
@@ -0,0 +1,97 @@
...
+	_, err := os.Stat(mountPath)
+	if os.IsNotExist(err) {
+		logger.Errorf("No volumed mounted at mountPath(%v): debug logging will not be performed: %v", mountPath, err)
```
xiang90: call createDirAll instead here.

hasbro17:
More [verbose logs](https://gist.github.com/hasbro17/c17541ded3c1577d5ce3eacaf69ada69) which print pod spec for every pod created.

xiang90:
lgtm after addressing the flag merge issue discussed offline. thanks!

hasbro17:
@hongchaodeng please take a final look. I've tested it manually on a self hosted cluster.
The debug logging will now be enabled if:
- Cluster is self-hosted
- And the flag `debug-logfile-path` is set

cmd/operator/main.go
```diff
@@ -72,6 +73,7 @@ var (
 
 func init() {
 	flag.BoolVar(&analyticsEnabled, "analytics", true, "Send analytical event (Cluster Created/Deleted etc.) to Google Analytics")
+	flag.StringVar(&debug.DebugFilePath, "debug-logfile-path", "", "path where the debug logfile will be written, recommended to be under: /var/tmp/etcd-operator/debug/ to avoid any issue with lack of write permissions")
```
xiang90: mention that it is only for self hosted cluster?

pkg/cluster/cluster.go
```diff
@@ -481,6 +490,12 @@ func (c *Cluster) removePod(name string) error {
 		if !k8sutil.IsKubernetesResourceNotFoundError(err) {
 			return err
 		}
+		if c.cluster.Spec.SelfHosted != nil && c.debugLogger != nil {
+			c.debugLogger.LogMessage(fmt.Sprintf("pod (%s) not found while trying to delete it", name))
+		}
+	}
+	if c.cluster.Spec.SelfHosted != nil && c.debugLogger != nil {
```
xiang90: can you enable/disable the logger during the initialization? then we do not need to check this everywhere.

hongchaodeng: Or just make it a function?
Like isDebugLoggerEnabled() ?

pkg/cluster/cluster.go
```diff
@@ -631,4 +646,15 @@ func (c *Cluster) logSpecUpdate(newSpec spec.ClusterSpec) {
 	for _, m := range strings.Split(string(newSpecBytes), "\n") {
 		c.logger.Info(m)
 	}
+
+	if c.IsDebugLoggerEnabled() {
+		c.debugLogger.LogClusterSpecUpdate(string(oldSpecBytes), string(newSpecBytes))
+	}
+}
+
+func (c *Cluster) IsDebugLoggerEnabled() bool {
```
hongchaodeng: Why is this public?

PR #1242, opened 2017-06-30
e2e/recovery_test: wait for pod size instead of ready member size

hasbro17:
Use `WaitUntilPodSizeReached()` to check the pods size, instead of `WaitUntilSizeReached()` which looks at the TPR's ready members, in the tests `TestOneMemberRecovery` and `TestDisasterRecovery`.

Before this the tests would mistakenly assume that the cluster had been recovered to size 3 after killing one or all members. The issue being that there is a delay between killing a member pod, and the cluster TPR getting updated to reflect that the removal of that member.

We fixed the same problem for other tests that use `KillMember()` in upgradetest and e2e tests in #1208 and #1101


hongchaodeng:
LGTM

PR #1244, opened 2017-07-03
adds support for using aws IAM credentials

sstarcher:
In reference to #1221.  This makes the secret and the 2 configurations under the secret optional.

This makes the secrets for configuration optional.  A few valid configurations would be.

* Create the secret for both credentials and config (Current method)
* Use AWS IAM instance profile with the config for aws region
* Use AWS IAM instance profile with an alternative aws region method such as AWS_REGION or AWS_DEFAULT_REGION environment variables.

@hongchaodeng

### DIFF
--- a/pkg/cluster/backupstorage/s3.go
+++ b/pkg/cluster/backupstorage/s3.go
@@ -1,6 +1,7 @@
 package backupstorage
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path"
@@ -36,14 +37,11 @@ func NewS3Storage(s3Ctx s3config.S3Context, kubecli kubernetes.Interface, cluste
 			if err := os.MkdirAll(dir, 0700); err != nil {
 				return nil, err
 			}
-			credsFile, configFile, err := setupAWSConfig(kubecli, ns, p.S3.AWSSecret, dir)
+			options, err := setupAWSConfig(kubecli, ns, p.S3.AWSSecret, dir)
 			if err != nil {
 				return nil, err
 			}
-			return backups3.NewFromSessionOpt(p.S3.S3Bucket, prefix, session.Options{
-				SharedConfigState: session.SharedConfigEnable,
-				SharedConfigFiles: []string{configFile, credsFile},
-			})
+			return backups3.NewFromSessionOpt(p.S3.S3Bucket, prefix, *options)
 		} else {
 			return backups3.New(s3Ctx.S3Bucket, prefix)
 		}
@@ -93,23 +91,37 @@ func (s *s3) Delete() error {
 	return nil
 }
 
-func setupAWSConfig(kubecli kubernetes.Interface, ns, secret, dir string) (string, string, error) {
+func setupAWSConfig(kubecli kubernetes.Interface, ns, secret, dir string) (*session.Options, error) {
+	options := &session.Options{}
+	options.SharedConfigState = session.SharedConfigEnable
+	options.SharedConfigFiles = make([]string, 0)
+
 	se, err := kubecli.CoreV1().Secrets(ns).Get(secret, metav1.GetOptions{})
 	if err != nil {
-		return "", "", err
+		return nil, fmt.Errorf("setup AWS config failed: get k8s secret failed: %v", err)
 	}
-	creds := se.Data[spec.AWSSecretCredentialsFileName]
-	credsFile := path.Join(dir, "credentials")
-	config := se.Data[spec.AWSSecretConfigFileName]
-	configFile := path.Join(dir, "config")
 
-	err = ioutil.WriteFile(credsFile, creds, 0600)
-	if err != nil {
-		return "", "", err
+	creds := se.Data[spec.AWSSecretCredentialsFileName]
+	if len(creds) != 0 {
+		credsFile := path.Join(dir, "credentials")
+		err = ioutil.WriteFile(credsFile, creds, 0600)
+		if err != nil {
+			return nil, fmt.Errorf("setup AWS config failed: write credentials file failed: %v", err)
+		}
+
+		options.SharedConfigFiles = append(options.SharedConfigFiles, credsFile)
 	}
-	err = ioutil.WriteFile(configFile, config, 0600)
-	if err != nil {
-		return "", "", err
+
+	config := se.Data[spec.AWSSecretConfigFileName]
+	if config != nil {
+		configFile := path.Join(dir, "config")
+		err = ioutil.WriteFile(configFile, config, 0600)
+		if err != nil {
+			return nil, fmt.Errorf("setup AWS config failed: write config file failed: %v", err)
+		}
+
+		options.SharedConfigFiles = append(options.SharedConfigFiles, configFile)
 	}
-	return credsFile, configFile, nil
+
+	return options, nil
 }


xiang90:
@sstarcher 

We need to document this somewhere. or users have no way to even know IAM works and how it works for etcd operator. I like this feature by the way.

pkg/cluster/backupstorage/s3.go
```diff
@@ -93,23 +90,36 @@ func (s *s3) Delete() error {
...
+
 	creds := se.Data[spec.AWSSecretCredentialsFileName]
-	credsFile := path.Join(dir, "credentials")
-	config := se.Data[spec.AWSSecretConfigFileName]
-	configFile := path.Join(dir, "config")
+	if creds != nil {
+		credsFile := path.Join(dir, "credentials")
+		err = ioutil.WriteFile(credsFile, creds, 0600)
+		if err != nil {
+			return
```
xiang90: kill naked return

pkg/cluster/backupstorage/s3.go
```diff
@@ -93,23 +90,36 @@ func (s *s3) Delete() error {
...
-	err = ioutil.WriteFile(configFile, config, 0600)
-	if err != nil {
-		return "", "", err
+
+	config := se.Data[spec.AWSSecretConfigFileName]
+	if config != nil {
+		configFile := path.Join(dir, "config")
+		err = ioutil.WriteFile(configFile, config, 0600)
+		if err != nil {
+			return
```
xiang90: same here

pkg/cluster/backupstorage/s3.go
```diff
@@ -93,23 +90,36 @@ func (s *s3) Delete() error {
...
+		}
+
+		options.SharedConfigFiles = append(options.SharedConfigFiles, configFile)
 	}
-	return credsFile, configFile, nil
+
+	return
```
xiang90: same here.

pkg/cluster/backupstorage/s3.go
sstarcher: not a fan of naked returns?

sstarcher:
@xiang90 anything else you would like me to change?

xiang90:
@sstarcher sorry. fighting against some test flakes and other user issues this week. we will give this closer look early next week. thanks for your patience.

doc/user/backup_config.md
```diff
@@ -28,12 +28,24 @@ If configurations for both levels are specified then the cluster level configura
 
 ### Operator level configuration  
 
-See the [S3 backup deployment](../../example/deployment-s3-backup.yaml.template) template on how to configure the operator to enable S3 backups. The following flags need to be passed to operator:
+The operator supports AWS credential configuration in several ways.
+* Environment variables
```
hongchaodeng: This is not something we need to explain in this doc. AWS docs should explain this.

doc/user/backup_config.md
```diff
@@ -28,12 +28,24 @@ If configurations for both levels are specified then the cluster level configura
...
+* Shared credentials file
+  * `backup-aws-secret` can be used to configure `~/.aws/credentials`
+  * `backup-aws-config` can be used to configure `~/.aws/config`
+* EC2 instance profiles
+
+See the [S3 backup deployment](../../example/deployment-s3-backup.yaml.template) template for an example of using shared credential files for the operator to enable S3 backups. 
+
+The following flags are supported for the operator:
 - `backup-aws-secret`: The name of the kube secret object that stores the AWS credential file. The file name must be 'credentials'.
-Profile must be "default".
```
hongchaodeng: Why remove this?

pkg/cluster/backupstorage/s3.go
```diff
@@ -36,14 +36,11 @@ func NewS3Storage(s3Ctx s3config.S3Context, kubecli kubernetes.Interface, cluste
 			if err := os.MkdirAll(dir, 0700); err != nil {
 				return nil, err
 			}
-			credsFile, configFile, err := setupAWSConfig(kubecli, ns, p.S3.AWSSecret, dir)
+			options, err := setupAWSConfig(kubecli, ns, p.S3.AWSSecret, dir)
```
hongchaodeng: Just to make sure. This is only changing the per cluster option, right? This is fine. But please mention it in "backup_config.md" per cluster option section.

doc/user/backup_config.md
sstarcher: Just to verify you just want me to delete these lines?

doc/user/backup_config.md
sstarcher: Because a user can use other values besides default?  As the AWS docs state if the user specifies `AWS_DEFAULT_PROFILE` environment variable they can use whatever profile they want.  I can add these lines back, but I don't believe it's a valid statement.

sstarcher: reverted

doc/user/backup_config.md
sstarcher: deleted

pkg/cluster/backupstorage/s3.go
sstarcher: added docs

pkg/cluster/backupstorage/s3.go
```diff
@@ -93,23 +90,36 @@ func (s *s3) Delete() error {
...
 	if err != nil {
-		return "", "", err
+		return options, nil
 	}
+
 	creds := se.Data[spec.AWSSecretCredentialsFileName]
-	credsFile := path.Join(dir, "credentials")
-	config := se.Data[spec.AWSSecretConfigFileName]
-	configFile := path.Join(dir, "config")
+	if creds != nil {
```
hongchaodeng:
```
if len(creds) != 0 {
```

pkg/cluster/backupstorage/s3.go
```diff
@@ -93,23 +90,36 @@ func (s *s3) Delete() error {
 	return nil
 }
 
-func setupAWSConfig(kubecli kubernetes.Interface, ns, secret, dir string) (string, string, error) {
+func setupAWSConfig(kubecli kubernetes.Interface, ns, secret, dir string) (options session.Options, err error) {
```
hongchaodeng: 1. Don't use named returns
2. return `(*session.Options, error)`

pkg/cluster/backupstorage/s3.go
```diff
@@ -93,23 +90,36 @@ func (s *s3) Delete() error {
...
+	options.SharedConfigState = session.SharedConfigEnable
+	options.SharedConfigFiles = make([]string, 0)
+
 	se, err := kubecli.CoreV1().Secrets(ns).Get(secret, metav1.GetOptions{})
 	if err != nil {
-		return "", "", err
+		return options, nil
```
hongchaodeng: fmt.Errorf("setup AWS config failed: get k8s secret failed: %v", err)

pkg/cluster/backupstorage/s3.go
```diff
@@ -93,23 +90,36 @@ func (s *s3) Delete() error {
...
+
 	creds := se.Data[spec.AWSSecretCredentialsFileName]
-	credsFile := path.Join(dir, "credentials")
-	config := se.Data[spec.AWSSecretConfigFileName]
-	configFile := path.Join(dir, "config")
+	if creds != nil {
+		credsFile := path.Join(dir, "credentials")
+		err = ioutil.WriteFile(credsFile, creds, 0600)
+		if err != nil {
+			return options, err
```
hongchaodeng: fmt.Errorf("setup AWS config failed: write credentials file failed: %v", err)

pkg/cluster/backupstorage/s3.go
```diff
@@ -93,23 +90,36 @@ func (s *s3) Delete() error {
...
-	err = ioutil.WriteFile(configFile, config, 0600)
-	if err != nil {
-		return "", "", err
+
+	config := se.Data[spec.AWSSecretConfigFileName]
+	if config != nil {
+		configFile := path.Join(dir, "config")
+		err = ioutil.WriteFile(configFile, config, 0600)
+		if err != nil {
+			return options, err
```
hongchaodeng: fmt.Errorf("setup AWS config failed: write config file failed: %v", err)

doc/user/backup_config.md
```diff
@@ -76,7 +82,7 @@ Then we could start using S3 storage for backups. See [spec examples](spec_examp
 
 See the [S3 backup with cluster specific configuration](https://github.com/coreos/etcd-operator/blob/master/doc/user/spec_examples.md#s3-backup-and-cluster-specific-s3-configuration) spec to see what the cluster's `spec.backup` field should be configured as to set a cluster specific S3 backup configuration. The following additional fields need to be set under the cluster spec's `spec.backup.s3` field:
 - `s3Bucket`: The name of the S3 bucket to store backups in.
-- `awsSecret`: The secret object name which should contain two files named `credentials` and `config` .
+- `awsSecret`: The secret object name which may contain two files named `credentials` and `config` .
```
hongchaodeng: There should be more detailed explanation for how this works.
But let's move this PR forward quickly.
Maybe you can undo the changes in this file. And then submit another PR for docs change (we don't need to wait CI). Or just leave it to us to change the docs.

pkg/cluster/backupstorage/s3.go
```diff
@@ -93,23 +90,36 @@ func (s *s3) Delete() error {
...
+		}
+
+		options.SharedConfigFiles = append(options.SharedConfigFiles, configFile)
 	}
-	return credsFile, configFile, nil
+
+	return options, err
```
hongchaodeng:
```
return options, nil
```

sstarcher:
@hongchaodeng let me know if I missed anything you requested.

pkg/cluster/backupstorage/s3.go
```diff
@@ -36,14 +37,11 @@ func NewS3Storage(s3Ctx s3config.S3Context, kubecli kubernetes.Interface, cluste
...
-			credsFile, configFile, err := setupAWSConfig(kubecli, ns, p.S3.AWSSecret, dir)
+			options, err := setupAWSConfig(kubecli, ns, p.S3.AWSSecret, dir)
 			if err != nil {
 				return nil, err
 			}
-			return backups3.NewFromSessionOpt(p.S3.S3Bucket, prefix, session.Options{
-				SharedConfigState: session.SharedConfigEnable,
-				SharedConfigFiles: []string{configFile, credsFile},
-			})
+			return backups3.NewFromSessionOpt(p.S3.S3Bucket, prefix, options)
```
hongchaodeng: need to pass `*options`

pkg/cluster/backupstorage/s3.go
```diff
@@ -93,23 +91,37 @@ func (s *s3) Delete() error {
...
-func setupAWSConfig(kubecli kubernetes.Interface, ns, secret, dir string) (string, string, error) {
+func setupAWSConfig(kubecli kubernetes.Interface, ns, secret, dir string) (*session.Options, error) {
+	options := &session.Options{}
+	options.SharedConfigState = session.SharedConfigEnable
+	options.SharedConfigFiles = make([]string, 0)
+
 	se, err := kubecli.CoreV1().Secrets(ns).Get(secret, metav1.GetOptions{})
 	if err != nil {
-		return "", "", err
+		return options, fmt.Errorf("setup AWS config failed: get k8s secret failed: %v", err)
```
hongchaodeng: return nil, ...

pkg/cluster/backupstorage/s3.go
```diff
@@ -93,23 +91,37 @@ func (s *s3) Delete() error {
...
-	err = ioutil.WriteFile(configFile, config, 0600)
-	if err != nil {
-		return "", "", err
+
+	config := se.Data[spec.AWSSecretConfigFileName]
+	if config != nil {
+		configFile := path.Join(dir, "config")
+		err = ioutil.WriteFile(configFile, config, 0600)
+		if err != nil {
+			return options, fmt.Errorf("setup AWS config failed: write config file failed: %v", err)
```
hongchaodeng: return nil, ...

PR #1250, opened 2017-07-06
Add SkipBootMemberRemoval to self-hosted spec

### LINKED ISSUE #1246: Decoupling boot member removal
I've been speaking with @luxas about how we can incorporate the etcd-operator into kubeadm, and one thing we've run into is how the etcd-operator handles removing the boot etcd member after the new etcd cluster is created. Right now it waits for a fixed time before beginning removal (60 seconds), but this is suboptimal for a few reasons:

- If there's 5TB of data and a slow network it will take longer than 60s, but the operator will just cut things in the middle

- Tightly coupling the creation of the new resource and deletion of old introduces some unnecessary constraints. Specifically, if we take the intended workflow of kubeadm:

1. Static pods (boot etcd, apiserver, controller-manager and scheduler) starts
2. etcd-operator deployment is created
3. TPR data is sent to operator specifying the boot etcd IP
4. operator creates self-hosted etcd cluster
5. operator adds new members to existing cluster and migrates data to new members
6. self-hosted control panel is deployed (apiserver uses service IP of self-hosted etcd)
7. static pods (apiserver, controller-manager and scheduler) are deleted directly after the self-hosted version is up
8. operator removes boot etcd from etcd cluster
9. boot etcd pod is deleted

But right now steps 5 and 8 are coupled together with a 60s buffer. Would it be possible to add an option to the self-hosted TPR spec which allows a user to disable the auto deletion of the boot member? If the bool is true, they are expected to handle the removal themselves.

jamiehannaford:
Fixes #1246

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### Added
 
+- New self-hosted field `SkipBootMemberRemoval` allows users to skip the
+  auto-deletion of the boot etcd member.
+
 ### Changed
 
 ### Removed
@@ -291,4 +294,3 @@ etcd container with this new field.
 ### Deprecated
 
 ### Security
-
--- a/pkg/cluster/self_hosted.go
+++ b/pkg/cluster/self_hosted.go
@@ -196,20 +196,25 @@ func (c *Cluster) migrateBootMember() error {
 		return err
 	}
 
-	go func() {
-		// TODO: a shorter timeout?
-		// Waiting here for cluster to get stable:
-		// - etcd data are replicated;
-		// - cluster TPR state has switched to "Running"
-		delay := 60 * time.Second
-		c.logger.Infof("wait %v before removing the boot member", delay)
-		time.Sleep(delay)
+	if c.cluster.Spec.SelfHosted.SkipBootMemberRemoval {
+		c.logger.Infof("skipping boot member (%s) removal; you will need to remove it yourself", endpoint)
+	} else {
+		c.logger.Infof("beginning the process of removing boot member (%s) from the cluster", endpoint)
+		go func() {
+			// TODO: a shorter timeout?
+			// Waiting here for cluster to get stable:
+			// - etcd data are replicated;
+			// - cluster TPR state has switched to "Running"
+			delay := 60 * time.Second
+			c.logger.Infof("waiting %v before removing the boot member", delay)
+			time.Sleep(delay)
 
-		err = etcdutil.RemoveMember([]string{newMember.ClientAddr()}, c.tlsConfig, bootMember.ID)
-		if err != nil {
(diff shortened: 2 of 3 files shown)


jamiehannaford:
cc/ @luxas

xiang90:
lgtm

luxas:
Perfect, thanks @jamiehannaford!

test/e2e/e2eslow/self_hosted_test.go
```diff
@@ -92,6 +93,40 @@ func testCreateSelfHostedClusterWithBootMember(t *testing.T) {
 	}
 }
 
+func testCreateSelfHostedClusterWithBootMemberNoAutoRemoval(t *testing.T) {
```
hongchaodeng: This test seems to fail:
```
FAIL: TestSelfHosted/migrate_boot_member_to_self_hosted_cluster_but_do_not_auto-remove_boot_member 
```

hongchaodeng: What's the point of this test? What is the scenario it tests for?

test/e2e/e2eslow/self_hosted_test.go
```diff
@@ -92,6 +93,40 @@ func testCreateSelfHostedClusterWithBootMember(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	defer func() {
+		if err := e2eutil.DeleteCluster(t, f.KubeClient, testEtcd); err != nil {
+			t.Fatal(err)
+		}
+	}()
+
+	if _, err := e2eutil.WaitUntilSizeReached(t, f.KubeClient, 2, 120*time.Second, testEtcd); err != nil {
```
hongchaodeng: This won't succeed.
There is an assumption that the boot member needs to be removed before scaling.

hongchaodeng:
@jamiehannaford 
I would like to merge this PR. But the test doesn't make much sense and is failing.
Can you separate the test out and also put it to CHANGELOG:
```
### Added

- `SkipBootMemberRemoval` does xxx
```

I think this makes more sense. Eventually, I want to make the same change in bootkube and only rely on installer tool to remove the member.

hongchaodeng:
Later on, I will change `testCreateSelfHostedClusterWithBootMember` to cover your scenario.

hongchaodeng:
LGTM

Failed test is known regression on k8s 1.7 and we are fixing it:
```
FAIL: TestClusterRestoreSameName 
```

Low risk, merging right now.

hongchaodeng:
@jamiehannaford 
FYI, we have nightly job that builds master into `quay.io/coreos/etcd-operator:dev`

jamiehannaford:
Awesome, thanks @hongchaodeng

PR #1259, opened 2017-07-07
upgrade_e2e: fix restore same name test

hongchaodeng:
This is the only failing test in upgrade e2e now.
Same issue as https://github.com/coreos/etcd-operator/pull/1251 .


xiang90:
lgtm

PR #1262, opened 2017-07-07
upgrade_e2e: delete endpoint after delete operator

hongchaodeng:
Upgrade test flakes caused by that even though two tests now uses different names for deployment , their endpoint locks still share the same name.

We could simply delete the endpoint after deleting the operator in each test.

### DIFF
--- a/test/e2e/upgradetest/framework/framework.go
+++ b/test/e2e/upgradetest/framework/framework.go
@@ -142,7 +142,12 @@ func (f *Framework) DeleteOperator(name string) error {
 		LabelSelector: labels.SelectorFromSet(operatorLabelSelector(name)).String(),
 	}
 	_, err = e2eutil.WaitPodsDeletedCompletely(f.KubeCli, f.KubeNS, 30*time.Second, lo)
-	return err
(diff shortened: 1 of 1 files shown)


test/e2e/upgradetest/framework/framework.go
```diff
@@ -142,7 +142,10 @@ func (f *Framework) DeleteOperator(name string) error {
 		LabelSelector: labels.SelectorFromSet(operatorLabelSelector(name)).String(),
 	}
 	_, err = e2eutil.WaitPodsDeletedCompletely(f.KubeCli, f.KubeNS, 30*time.Second, lo)
-	return err
+	if err != nil {
+		return err
+	}
+	return f.KubeCli.CoreV1().Endpoints(f.KubeNS).Delete("etcd-operator", metav1.NewDeleteOptions(0))
```
xiang90: add a comment here? leader election might change to not use endpoint in the future. this deletion might fails with resource not found.

hongchaodeng:
fixed

xiang90:
lgtm

Issue #1117, opened 2017-05-20
Use Taint Based Evictions


chancez:
TaintBasedEvictions is an alpha feature in 1.6 which let's you specify tolerations that effect when a pod is evicted as a result of certain node conditions. This can be used to prevent the pod from being evicted for a certain amount of time (via `tolerationSeconds`), or indefinitely when a node becomes NotReady, or unknown, etc.

Docs:

https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#per-pod-configurable-eviction-behavior-when-there-are-node-problems-alpha-feature

Notes: The etcd-operator could probably allow users to enable this today, so long as they can set tolerations, and not have to worry if the feature flag is even enabled, since it would just have no effect when the alpha feature is not enabled.

philips:
Yes! We should do this; glad this feature exists.

Then we need some way to reconcile actual node death with cloud providers, which we should probably do anyways. Create a NodeDeletionController or something.

chancez:
I'm pretty certain that an actual node death (ie: termination) is already reconciled correctly in Kubernetes. If you mean something else, like "down and not coming back, but not terminated", then I think that would ideally be handled by an ASG health check where the node is unhealthy and terminated by the ASG after the configured grace period.

As that relates to the TaintBasedEviction tolerations, if the node object is deleted (as a result of the instainced being terminated), then then the pod will be GCed/deleted in the API rather than evicted, and then re-created to be scheduled.

Of course, this all assumes the cloud, baremetal likely would need some kinda node deletion handling still.

diegs:
I would assume baremetal folks would want to know that one of their nodes is non-functional and go fix it.

hongchaodeng:
ref: https://kubernetes.io/docs/concepts/architecture/nodes/

> If the Status of the Ready condition is “Unknown” or “False” for longer than the pod-eviction-timeout, an argument passed to the kube-controller-manager, all of the Pods on the node are scheduled for deletion by the Node Controller. The default eviction timeout duration is five minutes. In some cases when the node is unreachable, the apiserver is unable to communicate with the kubelet on it. The decision to delete the pods cannot be communicated to the kubelet until it re-establishes communication with the apiserver. In the meantime, the pods which are scheduled for deletion may continue to run on the partitioned node.

We have hit this issue again (https://github.com/coreos-inc/tectonic-issues/issues/268). We need to figure out ways to prevent etcd pods from being evicted.

[...]

PR #1280, opened 2017-07-13
pkg/cluster: be smart about number of master nodes before remove any …

xiang90:
…member from the cluster

 If we do not have enough master nodes, we should simply wait for the old node to be online again.

Removing the etcd member will not help us to recover the cluster to the desired number of members, since we do not have enough master nodes for self hosted etcd.

Instead, there is a large chance that the old node is taken offline for maintenance temporary experiencing network partition. When it comes back, it will recover itself since we persist data for self hosted case.

/cc @hongchaodeng @hasbro17

### DIFF
--- a/pkg/cluster/reconcile.go
+++ b/pkg/cluster/reconcile.go
@@ -151,6 +151,29 @@ func (c *Cluster) removeOneMember() error {
 }
 
 func (c *Cluster) removeDeadMember(toRemove *etcdutil.Member) error {
+	if c.cluster.Spec.SelfHosted != nil {
+		selectedNodes, err := c.selectSchedulableNodes()
+		if err != nil {
+			return err
+		}
+
+		// If we do not have enough master nodes, we should simply wait for the old node
+		// to be online again.
+		//
+		// Removing the etcd member will not help us to recover the cluster to the desired
+		// number of members, since we do not have enough master nodes for self hosted etcd.
+		//
+		// Instead, there is a large chance that the old node is taken offline for maintenance
+		// or experiencing temporary network partition. When it comes back, it will recover itself
+		// since we persist data for self hosted case.
+
+		if nodeNum := len(selectedNodes); nodeNum < c.cluster.Spec.Size {
+			c.logger.Warningf("ignored removing failed member (%s). Not enough master nodes (%v) to recover, want at least %d", toRemove.Name, selectedNodes, c.cluster.Spec.Size)
+			c.logger.Infof("waiting for the failed master node to recover, or more master nodes")
+			return nil
+		}
+	}
+
 	c.logger.Infof("removing dead member %q", toRemove.Name)
 	c.status.AppendRemovingDeadMember(toRemove.Name)
 
--- a/pkg/cluster/self_hosted.go
+++ b/pkg/cluster/self_hosted.go
@@ -28,7 +28,9 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 )
 
-func (c *Cluster) selectMasterNodes() ([]string, error) {
+// selectSchedulableNodes finds all nodes that the etcd pod can be placed.
+// The selected nodes must satisfy the node selector and are in ready state.
+func (c *Cluster) selectSchedulableNodes() ([]string, error) {
 	var selector string
 	if c.cluster.Spec.Pod != nil && len(c.cluster.Spec.Pod.NodeSelector) != 0 {
 		selector = labels.SelectorFromSet(c.cluster.Spec.Pod.NodeSelector).String()
@@ -37,13 +39,17 @@ func (c *Cluster) selectMasterNodes() ([]string, error) {
 		LabelSelector: selector,
 	})
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to list master nodes: %v", err)
 	}
-	res := make([]string, len(nodes.Items))
-	for i := range nodes.Items {
-		res[i] = nodes.Items[i].Name
+	// TODO: take taints into consideration? k8s might support schedule dryrun for other
+	// components to test schedulability though...
+	ns := make([]string, 0)
+	for _, n := range nodes.Items {
+		if k8sutil.IsNodeReady(n) {
+			ns = append(ns, n.Name)
+		}
 	}
-	return res, nil
+	return ns, nil
 }
 
 func (c *Cluster) waitNewMember(oldN, retries int, name string) error {
@@ -112,7 +118,7 @@ func (c *Cluster) inspectSelfHostedMember(memberName, podName, ns string, oldN i
 }
 
 func (c *Cluster) addOneSelfHostedMember() error {
-	selectedNodes, err := c.selectMasterNodes()
+	selectedNodes, err := c.selectSchedulableNodes()
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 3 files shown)


pkg/cluster/reconcile.go
```diff
@@ -151,6 +151,29 @@ func (c *Cluster) removeOneMember() error {
...
+		// to be online again.
+		//
+		// Removing the etcd member will not help us to recover the cluster to the desired
+		// number of members, since we do not have enough master nodes for self hosted etcd.
+		//
+		// Instead, there is a large chance that the old node is taken offline for maintenance
+		// or experiencing temporary network partition. When it comes back, it will recover itself
+		// since we persist data for self hosted case.
+
+		if nodeNum := len(selectedNodes); nodeNum < c.cluster.Spec.Size {
```
hongchaodeng: this is still removing the etcd, right?

xiang90: what do you mean? i do not quite catch.

hongchaodeng: So if I have 3 nodes, and expected size to be 3. What the code will do:
1. it will detect `nodeNum == spec.Size`
2. it will continue to delete the member

xiang90: correct. when you have 3 nodes, and there is a failed member in a 3 nodes etcd cluster, it means one of the node is not running etcd. we should go ahead to remove the failed one, and add one back.

hongchaodeng: The failed one will restart itself, right?

xiang90: > The failed one will restart itself, right?

i do not understand. if you have 3 master nodes, and only two of them are running etcd. then it means you originally have 4 master nodes, and one of them is either dead or partitioned. it is reasonable to remove the member on that lost node and readd a member to the one without etcd.

can you provide me a full story of the case you have in mind? from the normal operation, to when it fails, and finally to what action the etcd operator is expected to take?

hongchaodeng: > then it means you originally have 4 master nodes, and one of them is either dead or partitioned

selectMasters() will return 4 nodes

xiang90: > selectMasters() will return 4 nodes

OK. Then select masters need to be fixed. it should select out ONLY ALIVE master nodes.

hongchaodeng: Right. We should select only "ready" nodes.

hasbro17: Okay so the master nodes selected by `selectMasterNodes()` are to be considered alive,
 i.e condition `Ready=True` only, and all other conditions false.
And nodes not selected could either be temporarily offline(Condition: `Ready=False/Unkown`), or have some permanent issue(Condition: `OutOfDisk=True`).
Does that sound right?
https://kubernetes.io/docs/concepts/architecture/nodes/#condition

xiang90: @hasbro17 yes. i think we need to find ready=true to start with. i am working on it :P. will ping you when it is done.

hasbro17:
SGTM. I can try to test this PR out on a multi-master self-hosted cluster by manually forcing a node to become unready and see that the operator does not delete the pod of the downed cluster.

xiang90:
@hasbro17 Thanks. Let me know how it goes. I am trying to add node kill/partition into coreos/ktestutil. so that we can write automated test against this :P.

hongchaodeng:
LGTM

hongchaodeng:
Haseeb and I justified a bit why we need this change. I'm writing down the discussions here.

Note that if a node was going down but not deleted, k8s would not delete pods from the node in a short time. In long time (5 mins), eviction will happen. This is not something this PR would address because it is controller manager taking the action. See https://github.com/coreos/etcd-operator/issues/1117 .

This is to take care when a node was reboot. When it was shutdown, it would have terminated all processes. After it reboot up, `docker ps` would show that etcd container `Exit (0)` before. Kubelet would take this information and update APIServer that etcd pod's phase to be `Completed`, and then based on etcd pod's `RestartPolicy: Always` to restart it. There could be a window that operator might delete it. This PR prevents that from happening.

xiang90:
> This is to take care when a node was reboot. When it was shutdown, it would have terminated all processes. After it reboot up, docker ps would show that etcd container Exit (0) before. Kubelet would take this information and update APIServer that etcd pod's phase to be Completed, and then based on etcd pod's RestartPolicy: Always restart it. There could be a window that operator might delete it. This PR prevents that from happening.

it is also possible that etcd operator have deleted the pod. then there will be a new pod in the pending state until this node back online. so the original pod will be deleted by the kubelet on that node, and a new pod will be created. this is another unwanted behavior.

However, on the other side, if this node is actually dead for ever, etcd operator WILL DO NOTHING until there is a new node comes up. This might lead to a bad situation.

For example, we have a 5 master nodes, 5 etcd nodes setup:
(n1, n2 ,n3, n4, n5)
(e1, e2, e3, e4, e5)

After n1 dies

with this PR
(    , n2 ,n3, n4, n5)
(**e1**, e2, e3, e4, e5)

with out this PR
(    , n2 ,n3, n4, n5)
(    , e2, e3, e4, e5)

After n2 dies

with this PR
(    ,    ,n3, n4, n5)
(**e1**, **e2**, e3, e4, e5)

with out this PR
(    ,    ,n3, n4, n5)
(    ,   , e3, e4, e5)

After n2 dies

with this PR
(    ,    ,   , n4, n5)
(**e1**, **e2**, **e3**, e4, e5) -> **Lost quorum**!

with out this PR
(    ,    ,  , n4, n5) -> Still OK!
(    ,   ,   , e4, e5)

But I think the situation I showed is a extreme case, since users should set node alter, and add back enough nodes when there are node failures. So I believe the advantage this PR brings overweights its disadvantage.

hongchaodeng:
After n1 dies

with this PR
( , n2 ,n3, n4, n5)
(e1, e2, e3, e4, e5)

without this PR
( , n2 ,n3, n4, n5)
( , e2, e3, e4, e5)

`======`

Are you assuming n1 is not deleted from API. Even without this PR, e1 won't be deleted, right?

xiang90:
> Are you assuming n1 is not deleted from API. Even without this PR, e1 won't be deleted, right?

I assume n1 is deleted from the api server, but no new node is added.

xiang90:
Again

> But I think the situation I showed is a extreme case

I have this assumption, since if n1 is deleted, either the auto scaling group deleted it or a human deleted it from k8s api. A new node should be expected to add back shortly either by the scaling group and the same human.

hongchaodeng:
I thought maintenance would done via node draining: https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/#use-kubectl-drain-to-remove-a-node-from-service

xiang90:
@hongchaodeng 

Well... Yes, if it is a planned long-term maintenance. If I just want to take the node down for 3 minutes for updating my OS. I probably wont brother to ask all the pods to be rescheduled, especially for master nodes.

hongchaodeng:
Why is the node deleted from API?

xiang90:
What do you mean? I explained that a human might do it or aws auto scaling group might do it. The node might be forever dead, for example the hardware is broken so that the hardware under the vm is taken to replace.

hongchaodeng:
If a node was deleted, then the etcd pod be deleted too. The etcd operator will create a new member on new nodes.

PR #1294, opened 2017-07-14
pollPod add handle inactive pod.ignore inactive pod.

wilhelmguo:
Kubernetes will not force delete pods and change pod's status.phase, if node unready or lost, the 
status.phase still Running, but set deletionTimestamp.
If deletionTimestamp set we think the pods not Running.

[skip ci]


hongchaodeng:
If deletionTimestamp is not nil, the pod will normally be deleted soon. It's by design that we won't remove membership until pod is entirely deleted from API.

The problem you described is that node is lost. If this happens, it needs external manual or automatic process to remove the dead node and add a new one.

xiang90:
i am closing this now. deletion timestamp is designed for graceful shutdown. when it is set, the pod will be deleted soon anyway. this seems like an optimization that we do not need.

Issue #1305, opened 2017-07-17
support etcd bucket prefix


sstarcher:
Current the format for ETCD backups in a self-hosted cluster requires that each cluster has it's own bucket.  From a management perspective it could be helpful to have a single bucket that contains multiple clusters.

Current the format is `BUCKET/v1/NAMESPACE/ETCD_CLUSTER`
Would it be possible to support or interested in supporting a custom location or a prefixed location.

`BUCKET/ID/v1/NAMESPACE/ETCD_CLUSTER` would work in my situation as we don't give each cluster a bucket, but a subfolder under the bucket itself.

hongchaodeng:
fixed in https://github.com/coreos/etcd-operator/pull/1372

PR #1306, opened 2017-07-17
(feat) add support for s3 path prefix configuration

sstarcher:
#1305 I have tested this in the following configuration.  Which results in files under
`mybucket/path/etcd/v1/kube-system/kube-etcd/` for a self-hosted etcd setup

```
  backup:
    backupIntervalInSecond: 300
    maxBackups: 5
    s3:
      awsSecret: aws
      s3Bucket: mybucket
      s3Path: path/etcd
    storageType: S3
```

### DIFF
--- a/pkg/backup/s3/s3.go
+++ b/pkg/backup/s3/s3.go
@@ -38,32 +38,28 @@ type S3 struct {
 // New returns a S3 translator from default shared config.
 // Please refer to http://docs.aws.amazon.com/sdk-for-go/api/aws/session/
 // for setting up credentials and configuration.
-func New(bucket, prefix string) (*S3, error) {
-	return NewFromSessionOpt(bucket, prefix, session.Options{
+func New(bucket, pathPrefix, prefix string) (*S3, error) {
+	return NewFromSessionOpt(bucket, pathPrefix, prefix, session.Options{
 		// Setting this is equal to the AWS_SDK_LOAD_CONFIG environment variable was set.
 		// We want to save the work to set AWS_SDK_LOAD_CONFIG=1 outside.
 		SharedConfigState: session.SharedConfigEnable,
 	})
 }
 
-func NewFromSessionOpt(bucket, prefix string, so session.Options) (*S3, error) {
+func NewFromSessionOpt(bucket, pathPrefix, prefix string, so session.Options) (*S3, error) {
 	sess, err := session.NewSessionWithOptions(so)
 	if err != nil {
 		return nil, fmt.Errorf("new AWS session failed: %v", err)
 	}
 	cli := s3.New(sess)
 
-	return &S3{
-		bucket: bucket,
-		prefix: prefix,
-		client: cli,
-	}, nil
+	return NewFromClient(bucket, pathPrefix, prefix, cli), nil
 }
 
-func NewFromClient(bucket, prefix string, cli *s3.S3) *S3 {
+func NewFromClient(bucket, pathPrefix, prefix string, cli *s3.S3) *S3 {
 	return &S3{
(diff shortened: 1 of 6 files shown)


hongchaodeng:
@sstarcher 
Thanks for the PR. This looks a good feature to have.
Please wait a week or so because we are doing CRD migration work. Thanks for your patience in advance!

sstarcher:
No problem, I'm hoping to have time to finish it up and fix the build issue.

pkg/backup/s3/s3.go
```diff
@@ -38,40 +38,36 @@ type S3 struct {
...
-	return &S3{
-		bucket: bucket,
-		prefix: prefix,
-		client: cli,
-	}, nil
+	return NewFromClient(bucket, path_prefix, prefix, cli), nil
 }
 
-func NewFromClient(bucket, prefix string, cli *s3.S3) *S3 {
+func NewFromClient(bucket, path_prefix, prefix string, cli *s3.S3) *S3 {
```
xiang90: pathPrefix

pkg/spec/backup.go
```diff
@@ -85,6 +85,8 @@ type S3Source struct {
 	// S3Bucket overwrites the default etcd operator wide bucket.
 	S3Bucket string `json:"s3Bucket,omitempty"`
 
+	S3Path string `json:"s3Path,omitempty"`
```
xiang90: doc string

xiang90:
looks good in general. can we write a test for this? also conflicts need to be resolved. thanks!

sstarcher:
@xiang90 I rebased and made the modifications you requested.  As for the test I used an existing integration test and set it up to use the prefix for testing.

xiang90:
lgtm

/cc @hongchaodeng @hasbro17 can you please take another look?

pkg/backup/s3/s3.go
```diff
@@ -38,40 +38,36 @@ type S3 struct {
...
-	}, nil
+	return NewFromClient(bucket, pathPrefix, prefix, cli), nil
 }
 
-func NewFromClient(bucket, prefix string, cli *s3.S3) *S3 {
+func NewFromClient(bucket, pathPrefix, prefix string, cli *s3.S3) *S3 {
 	return &S3{
 		bucket: bucket,
-		prefix: prefix,
+		prefix: path.Join(pathPrefix, v1, prefix),
```
hongchaodeng: It is not good to pass two "prefix" here. That is confusing and complicates calling depth.

It would be preferred to join those prefixes before calling New().


One suggestion is to change v1 to a func:
```
func V1Path(userPrefix, clusterPrefix string) string {
  return path.Join(userPrefix, "v1", clusterPrefix)
}
```

pkg/backup/s3_backend_test.go
```diff
@@ -92,3 +92,53 @@ func TestS3BackendPurge(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestS3BackendWithPrefix(t *testing.T) {
```
hongchaodeng: This doesn't test the actual prefix. A white box testing of S3 path is needed.

xiang90:
@sstarcher kindly ping. would like to get this merged for the next release :)

sstarcher:
@xiang90 I don't know if I'll have time before your next release to make the changes @hongchaodeng  requested

fanminshi:
favor https://github.com/coreos/etcd-operator/pull/1372 over this

sstarcher:
@fanminshi thanks for continuing the work.  When #1372 is merged we can close this.

PR #1327, opened 2017-07-25
*: replace TPR with CRD and update apiVersion and kind in docs and examples

hasbro17:
@hongchaodeng everything under `doc/design` is currently unchanged since I'm not sure if we are planning to keep them up to date.

Still need to test out each of the updated examples with the new spec to make sure that they work correctly.


hongchaodeng:
LGTM

Issue #1236, opened 2017-06-29
migrate to CRD


xiang90:

hongchaodeng:
So in order to migrate to CRD, here are the steps that we need to do:
1. create CRD
2. copy cluster TPR to cluster CR
3. shutdown operator
4. update relevant resources' ownerRefs
5. delete cluster TPR, delete TPR
6. restart operator

hongchaodeng:
We will prioritize this item since TPR will be removed in k8s 1.8 . We need to ship this as early as we could and have it tested as much as possible.

hongchaodeng:
Has 30 passes right now: https://jenkins-etcd-public.prod.coreos.systems/view/operator/job/etcd-operator-flake/

Testing is stable for "crd" branch.

PR #1331, opened 2017-07-27
*: Replace TPR with CRD

hongchaodeng:
ref: https://github.com/coreos/etcd-operator/issues/1236

All requirements have been fulfilled. See above issues.


hongchaodeng:
It has been fully tested. Of course other community tryouts are encouraged!

Issue #1308, opened 2017-07-18
Support encryption of backup


klausenbusk:
The etcd backup can contain some very sensitive data (e.g: when using self-hosted etcd with [bookube](https://github.com/coreos/bootkube), it contains all the k8s secrets).

The access to the backups should be restricted (ex: IAM), but providing a way to encrypt the backup wouldn't hurt.

hongchaodeng:
@Nesze Moving discussion of https://github.com/coreos/etcd-operator/issues/1339 to here.

Encryption is preferable on etcd backup data. However, it shouldn't be coupled with etcd backup sidecar or etcd operator. etcd operator should only take the backup. It would then call another service to encrypt the data and even actually store the data into S3. Because this would make etcd operator have less dependencies and easier to maintain.

Solution:
- Ideally, we will have extensible storage backend design (https://github.com/coreos/etcd-operator/issues/1241) which calls REST APIs. If you can't wait for this, then
- Currently, we have a backend interface: https://github.com/coreos/etcd-operator/blob/2665c3e51963836119a296f37e10b71938c23c81/pkg/backup/backend.go . You can implement it, improve it to fit the use case.

hongchaodeng:
Closing this.
See custom backup plan in https://github.com/coreos/etcd-operator/issues/1626

Issue #1241, opened 2017-06-30
Extensible backup storage interface


hongchaodeng:
Currently our backup sidecar handles the backup logic and controls how to store backup files. We have opinions in how to do it based on our experience.

However, over time we have seen growing user scenarios that are different than our "how to store files":
- User who doesn't like our config format: https://github.com/coreos/etcd-operator/pull/1221#issuecomment-311533483
- User who haves their in house storage solutions (from Ebay) (https://github.com/coreos/etcd-operator/issues/1245)
- User who wants new cloud features: Azure (https://github.com/coreos/etcd-operator/issues/1110, https://github.com/coreos/etcd-operator/pull/1237), GCS (https://github.com/coreos/etcd-operator/pull/1127)
- User who want to encrypt backup: https://github.com/coreos/etcd-operator/issues/1308

It would be reasonable to have a more extensible backup storage interface, which makes backup sidecar to only handle the backup logic, but leaves "how to store files" up to the interface. For example, have another HTTP service that serves APIs for write, read, delete, list operations of snapshot data. This would also decouple config/credential management from backup logic.

hongchaodeng:
We should close this for https://github.com/coreos/etcd-operator/issues/1626.
Please see and follow up conversation in that thread.

PR #1336, opened 2017-07-31
[WIP] Support encryption of backup

klausenbusk:
Not tested

----
Fix https://github.com/coreos/etcd-operator/issues/1308
Just a quick attempt, not tested or anything and it could properly be improved :)

### DIFF
--- a/pkg/backup/backup.go
+++ b/pkg/backup/backup.go
@@ -15,8 +15,12 @@
 package backup
 
 import (
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/rand"
 	"crypto/tls"
 	"fmt"
+	"io"
 	"log"
 	"os"
 	"path"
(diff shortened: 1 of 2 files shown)


hongchaodeng:
We won't go the route of implementing encryption in etcd operator.
Ideally, we will rely on external service like Vault to do encryption.

hongchaodeng:
We will provide simpler ways to extend storage backend in the future: https://github.com/coreos/etcd-operator/issues/1241 .
You can implement a encryption backend for it later.

klausenbusk:
I'm properly just gonna crate my own simple backup solution. I only have a single `etcd` cluster (for k8s).
So something like:
```
f="$(date -Isec)-etcd.backup"
ETCDCTL_API=3 etcdctl snapshot save "${f}"
echo "${SOME_SECRET}" | gpg --batch -c --passphrase-fd 0 s
mc cp "${f}.gpg" s3/some-bucket/
```

PR #1345, opened 2017-08-04
doc/user/op_guide: Fix typo verb tense create vs create

joshix:
as downstream in https://github.com/coreos-inc/coreos-pages/pull/1773

[skip ci]


hongchaodeng:
Thanks!

Issue #1323, opened 2017-07-25
Persistent/Durable etcd cluster


sgotti:
I'd like to be able to use etcd-operator to achieve a persistent/durable etcd cluster. I'd like to avoid as much as possible (only when I know that my members will never come back) the need to restore from a backup only because a majority of member dies at the same time (this could happen a lot of times for a lot of unpredictable reasons).
If I know that they will come back (rescheduled by etcd-operator) I'll prefer losing availability for some time (while waiting for they to come back).

Right now looks like etcd operator cannot achieve this. From the README.md:
```
If the majority of etcd members crash, but at least one backup exists for the cluster, the etcd operator can restore the entire cluster from the backup.
```

I did a little analysis on how etcd-operator works in this post https://sgotti.me/post/kubernetes-persistent-etcd/ 

Now etcd-operator directly schedules pod (acts like a custom controller) and uses emptyDir for member data. Everytime a pod dies (or is manually deleted) a new replacement pod is created and the old member removed from the etcd cluster. When a majority of pods dies at the same time the cluster cannot be restored for two primary reasons:

* etcd member data not available anymore
* replacement pods have different ip addresses and you cannot update the peer member urls in a unquorate cluster. 

To fix these points I can see the below solutions, but, probably, this will require a big change in the current etcd-operator architecture:

### Use "persistent" etcd member data.
* Use persistent volumes to not lose etcd member data after a pod dies (or is deleted).
* When a pod is crashed or deleted etcd-operator should create a new replacement pod reusing the pv previously associated to the previous dead pod.
* As an additional note, using persistent volumes will require an *at most once* scheduling like done by k8s statefulset (see https://github.com/kubernetes/community/pull/124) to avoid, when using a pv provider that doesn't have an inner guarantee of mouting a pv only on one node at a time, possible data corruptions.

### Persistent peer ip addresses.

When creating a replacement pod for the above point you'll end with a pod with a new ip. You could update the etcd cluster peer ip addresses with the new pod address but this will work only when losing a minority of the etcd members. If you lose e majority of the etcd members the etcd cluster will be unquorate and won't accept a member update.

About this last point, before https://github.com/coreos/etcd/issues/6336 you was able to provide a peer url containing a domain name. So instead of using "persistent" peer ip addresses you could use "persistent" network names for example creating an headless service with an endpoint for every member pod that will resolve a fixed member name (like etcd-0, etcd-1 etc...) to the ip of the current pod (like done by a k8s statefulset).

[...]

Issue #1201, opened 2017-06-15
kubernetes: explore the localstorage work


philips:
Localstorage would make it possible for us to move off of hostPath. See instructions here: https://github.com/kubernetes-incubator/external-storage/tree/master/local-volume

cc @lpabon

hongchaodeng:
Ref https://github.com/coreos/etcd-operator/issues/957

Shedding some thoughts.
Using local pv will empower us to use cloud SSD disk, and tolerate failures and restarts, and reduce the overhead of replicating a fresh new member.

philips:
Based on the local volume prototype above someone (maybe an intern?) can start exploring this and providing feedback. A goal of the upstream team is to get this to stable and usable by etcd Operator.

lpabon:
We will add it to our plans. There is also work to add local or network
block based storage. Maybe necessary for databases.

- Luis

prune998:
Any news here ?

PR #1349, opened 2017-08-07
[WIP] *: Add initial PersistentVolume feature

sgotti:
this patch adds initial persistent volume feature for etcd data.

When the usePV option is enabled in the etcd cluster spec every
created pod will get a related persistent volume claim and the pod
restart policy will be set to RestartAlways.

This initial patch only covers the case where a pod crashes (or a node
restart before the related pod is evicted by the k8s node
controller). When using a persistent volume, instead of deleting the pod
and creating a new one, the pod will be restarted by k8s.

A future patch will change the reconcile logic to also handle cases
where one or more pods are deleted (manually or by k8s) recreating the
same pod using the same pvc.

The etcd data pvc will be garbage collected like the other cluster
resources.

Questions:
* I'm using the same backup storage class (the name isn't great since it starts with `etcd-backup-`. Should we define different storage classes using the same provisioner or permit different provisioner for backup and data? 

* I'll add some questions inline in the code.

TODO:
- [ ] Check all the tests on your infrastructure (I'm testing on a local k8s cluster disabling tests that needs S3 and updating the generated storage class with another pv provisioner).

### DIFF
--- a/example/example-etcd-cluster-with-pv.yaml
+++ b/example/example-etcd-cluster-with-pv.yaml
@@ -1,0 +1,11 @@
+apiVersion: "etcd.database.coreos.com/v1beta2"
+kind: "EtcdCluster"
+metadata:
+  name: "example-etcd-cluster"
+spec:
+  size: 3
+  version: "3.1.8"
+  pod:
+    usePV: true
+    pvPolicy:
+      volumeSizeInMB: 1024
--- a/hack/test
+++ b/hack/test
@@ -88,7 +88,10 @@ function e2e_pass {
 
 	# Run all the tests by default
 	E2E_TEST_SELECTOR=${E2E_TEST_SELECTOR:-.*}
-	go test "./test/e2e/" -run "$E2E_TEST_SELECTOR" -timeout 30m --race --kubeconfig $KUBECONFIG --operator-image $OPERATOR_IMAGE --namespace ${TEST_NAMESPACE}
+	# Run tests with PV support disabled
+	go test -v "./test/e2e/" -run "$E2E_TEST_SELECTOR" -timeout 30m --race --kubeconfig $KUBECONFIG --operator-image $OPERATOR_IMAGE --namespace ${TEST_NAMESPACE}
+	# Run tests with PV support enabled
+	PV_TEST=true go test -v "./test/e2e/" -run "$E2E_TEST_SELECTOR" -timeout 30m --race --kubeconfig $KUBECONFIG --operator-image $OPERATOR_IMAGE --namespace ${TEST_NAMESPACE}
 }
 
 function e2eslow_pass {
--- a/pkg/spec/cluster.go
+++ b/pkg/spec/cluster.go
@@ -28,6 +28,8 @@ import (
 const (
 	defaultBaseImage = "quay.io/coreos/etcd"
 	defaultVersion   = "3.1.8"
+
+	minPVSize = 512 // 512MiB
 )
 
 var (
@@ -154,6 +156,14 @@ type PodPolicy struct {
 	// bootstrap the cluster (for example `--initial-cluster` flag).
 	// This field cannot be updated.
 	EtcdEnv []v1.EnvVar `json:"etcdEnv,omitempty"`
+
+	UsePV    bool      `json:"usePV,omitempty"`
+	PVPolicy *PVPolicy `json:"pvPolicy,omitempty"`
+}
+
+type PVPolicy struct {
+	// VolumeSizeInMB specifies the required volume size for storing etcd data.
+	VolumeSizeInMB int `json:"volumeSizeInMB"`
 }
 
 func (c *ClusterSpec) Validate() error {
@@ -182,6 +192,11 @@ func (c *ClusterSpec) Validate() error {
 				return errors.New("spec: pod labels contains reserved label")
 			}
 		}
+		if c.Pod.UsePV && c.Pod.PVPolicy != nil {
+			if c.Pod.PVPolicy.VolumeSizeInMB < minPVSize {
+				return fmt.Errorf("spec: pod pv volume size lesser than min size (%dMiB)", minPVSize)
+			}
+		}
 	}
 	return nil
 }
@@ -198,6 +213,10 @@ func (c *ClusterSpec) Cleanup() {
 	}
 
 	c.Version = strings.TrimLeft(c.Version, "v")
+
+	if c.Pod.UsePV && c.Pod.PVPolicy == nil {
+		c.Pod.PVPolicy = &PVPolicy{VolumeSizeInMB: minPVSize}
+	}
 }
 
 type ClusterPhase string
--- a/pkg/util/etcdutil/member.go
+++ b/pkg/util/etcdutil/member.go
@@ -70,6 +70,10 @@ func (m *Member) PeerURL() string {
 	return fmt.Sprintf("%s://%s:2380", m.peerScheme(), m.FQDN())
 }
 
+func (m *Member) PVCName() string {
+	return fmt.Sprintf("%s-pvc", m.Name)
+}
+
 type MemberSet map[string]*Member
 
 func NewMemberSet(ms ...*Member) MemberSet {
@@ -187,4 +191,12 @@ func clusterNameFromMemberName(mn string) string {
 		panic(fmt.Sprintf("unexpected member name: %s", mn))
 	}
 	return mn[:i]
+}
+
+func MemberNameFromPVCName(pn string) string {
+	i := strings.LastIndex(pn, "-")
+	if i == -1 {
+		panic(fmt.Sprintf("unexpected pvc name: %s", pn))
+	}
+	return pn[:i]
 }
--- a/pkg/util/k8sutil/k8sutil.go
+++ b/pkg/util/k8sutil/k8sutil.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"path"
 	"strings"
 	"time"
 
@@ -30,6 +31,7 @@ import (
 	appsv1beta1 "k8s.io/api/apps/v1beta1"
 	"k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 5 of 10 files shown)


pkg/spec/cluster.go
```diff
@@ -28,6 +28,8 @@ import (
 const (
 	defaultBaseImage = "quay.io/coreos/etcd"
 	defaultVersion   = "3.1.8"
+
+	minPVSize = 512 // 512MiB
```
sgotti: Is this a sane min size default?

pkg/util/k8sutil/k8sutil.go
```diff
@@ -220,7 +222,38 @@ func addOwnerRefToObject(o metav1.Object, r metav1.OwnerReference) {
...
+			Name:      name,
+			Namespace: namespace,
+			Labels: map[string]string{
+				"etcd_node":    m.Name,
+				"etcd_cluster": clusterName,
+				"app":          "etcd",
+			},
+		},
+		Spec: v1.PersistentVolumeClaimSpec{
+			StorageClassName: &storageClassName,
```
sgotti: Here I'm using Spec.StorageClassName instead of the old annotation `volume.beta.kubernetes.io/storage-class` like it's used in the backup pvc: https://github.com/coreos/etcd-operator/blob/master/pkg/util/k8sutil/backup.go#L74 . Should this be updated?

hack/test
```diff
@@ -88,7 +88,10 @@ function e2e_pass {
 
 	# Run all the tests by default
 	E2E_TEST_SELECTOR=${E2E_TEST_SELECTOR:-.*}
-	go test "./test/e2e/" -run "$E2E_TEST_SELECTOR" -timeout 30m --race --kubeconfig $KUBECONFIG --operator-image $OPERATOR_IMAGE --namespace ${TEST_NAMESPACE}
+	# Run tests with PV support disabled
+	go test -v "./test/e2e/" -run "$E2E_TEST_SELECTOR" -timeout 30m --race --kubeconfig $KUBECONFIG --operator-image $OPERATOR_IMAGE --namespace ${TEST_NAMESPACE}
+	# Run tests with PV support enabled
+	PV_TEST=true go test -v "./test/e2e/" -run "$E2E_TEST_SELECTOR" -timeout 30m --race --kubeconfig $KUBECONFIG --operator-image $OPERATOR_IMAGE --namespace ${TEST_NAMESPACE}
```
sgotti: Instead of changing all the test functions to use a common function that accept a `usePV bool` and create a parent function calling it with it to true and to false, I just (for speed) added an env variable that changes the behavior of `NewCluster`. Let me know your preference.

example/example-etcd-cluster-with-pv.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: "EtcdCluster"
+metadata:
+  name: "example-etcd-cluster"
+spec:
+  size: 3
+  version: "3.1.8"
+  pod:
+    usePV: true
+    pvPolicy:
+      volumeSizeInMB: 1024
```
sgotti: This is just a proposed spec. I put it under the PodPolicy but I'm not sure about its location. I'm also not sure about the `usePV` bool name and the related pvPolicy. I'm open for any better idea.

hongchaodeng:
Please hold on.
I'm still not sure if we are going to use PVC for etcd storage.
Instead, we would like to go the path of local PV: https://github.com/coreos/etcd-operator/issues/1201

xiang90:
> Please hold on.
I'm still not sure if we are going to use PVC for etcd storage.
Instead, we would like to go the path of local PV: #1201

Local PV and PVC are different. I am fine with adding PV as long as it does not interleave too much with the current logic. Or we have to refactor the code before adding it.

hongchaodeng:
> Local PV and PVC are different.

Well. They are somehow the same. They are both PV abstractions. They can share the same code path for tolerating etcd downtime, GC, and management code in etcd operator. At least, we should have decided down what local PV looks like before adding option for non-local PV. Local PV should take higher priority.

pkg/spec/cluster.go
```diff
@@ -154,6 +156,14 @@ type PodPolicy struct {
 	// bootstrap the cluster (for example `--initial-cluster` flag).
 	// This field cannot be updated.
 	EtcdEnv []v1.EnvVar `json:"etcdEnv,omitempty"`
+
+	UsePV    bool      `json:"usePV,omitempty"`
```
xiang90: kill this bool? we can check if PVPolicy is empty.

hongchaodeng:
Actually it is fine to have PVC support. But it is very important to make sure the abstractions are right so that we can share the code path with local PV as well.

pkg/util/etcdutil/member.go
```diff
@@ -188,3 +192,11 @@ func clusterNameFromMemberName(mn string) string {
 	}
 	return mn[:i]
 }
+
+func MemberNameFromPVCName(pn string) string {
```
xiang90: doc string.

pkg/util/k8sutil/k8sutil.go
```diff
@@ -220,7 +222,38 @@ func addOwnerRefToObject(o metav1.Object, r metav1.OwnerReference) {
...
+			},
+		},
+	}
+
+	addOwnerRefToObject(pvc.GetObjectMeta(), owner)
+
+	return pvc
+}
+
+func NewEtcdPod(m *etcdutil.Member, initialCluster []string, clusterName, state, token string, cs spec.ClusterSpec, usePVC bool, owner metav1.OwnerReference) *v1.Pod {
```
xiang90: can we infer PVC from cluster spec?

pkg/spec/cluster.go
xiang90: seems ok

sgotti:
> Well. They are somehow the same. They are both PV abstractions. They can share the same code path for tolerating etcd downtime, GC, and management code in etcd operator. At least, we should have decided down what local PV looks like before adding option for non-local PV. Local PV should take higher priority.

@hongchaodeng @xiang90 My opinion is that they are both PV and both can be abstracted using a PVC. I'll always use a PVC instead of directly creating a PV since a PVC adds some goodies for PV management (the storageclass => provisioning and volume node affinity).
The main difference between them is the node affinity. One is "local" (available only on one node) and others (EBS, GCE PD etc...) are "global" (available on more than one node (one node at a time)), though they can be available only on some subset of the total k8s cluster nodes like EBS that are per availability zone).

This difference in node affinity can bring to some differences in the "reconcile" logic. Taking from https://github.com/coreos/etcd-operator/issues/1323#issuecomment-318684441: 

> 1. pod restart - still use previous PV
> 2. pod movement - detach the PV on the pod to move (or dead), and reattach that PV to the new pod

(NOTE: this PR implements 1).

Point 2, if intended as pod movement between different nodes, can be done only with "global" PV.

Another feature (that is one of the main reason I'm trying to implement PV support) is that, instead of doing "disaster recovery" when there're less than N/2+1 RUNNING etcd pods due to some k8s nodes going down, one could just wait for them to be come back (node reentered the k8s cluster that will restart the pod since the policy is RestartAlways) or wait for pods to be deleted from the API (by the node controller + manual intervention or a fence controller) and recreate them using the previous PV that will be scheduled on another node.

This is the second part of this saga. I already implemented a part of it (but with a lot of assumption since with PV you have multiple choices) so If you want I'll be happy to open another RFC PR if it can be helpful to better explain the overall idea.

pkg/util/k8sutil/k8sutil.go
sgotti: In this PR it could be removed because not needed.

It's here because in the second PR where I implemented POD recreation using an existing PVC I wanted to handle the case where the user disabled PVC in the cluster spec but I'd like to continue using existing PVC for pod recreation. But now that I think about it another implementation could just ignore existing PVC when PV support is disabled. So I'll remove it.

pkg/spec/cluster.go
sgotti: Yeah, that was just one idea: use a bool to enable PV usage and the user could leave an nil PVPolicy (that will use the default VolumeSizeInMB value). Removing it will require users to always set PVPolicy.VolumeSizeInMB but I think it's not so complicated from the user perspective.

hongchaodeng:
@sgotti 
So I think the purpose of this PR is to **add a PVC for each etcd pod and change restartPolicy to Always**, right?  Let's not handle too many stuff at once and take it step by step:
- Assume the etcd pod will always restart on the same node for now
- For next PR, it will handle the toleration, re-attach PVC to another pod, disaster recovery, etc. Needs discussion in https://github.com/coreos/etcd-operator/issues/1323 .

So this PR will handle:
- create PVC when scaling up etcd
- delete PVC when scaling down etcd
- wait etcd to restart instead of removing it

For this to move forward, this is what we should do:
- A PR for only spec change. How would user tell us it expects each etcd pod to uses a PVC, and what kind of PV it would use?
- A PR for documentation, how could users use it.
- A PR for internal changes, some heavy-lifting refactoring of the code to make sure good design
- A PR for testing.

xiang90:
@sgotti 

I think the direction is good in general. As @hongchaodeng mentions, can we split this effort into multiple PRs? Thank you!

xiang90:
@sgotti kindly ping on this one. is there anything we could help to move this forward?

sgotti:
@xiang90 @hongchaodeng sorry for being late, I'm in vacation until next week. OK for the plan, I'm not sure if you already have in mind which refactoring will be needed, and I'll keep implementation and tests in the same PR for obvious reasons.

As a first step I'd like to propose some possible changes on the way volume management is handled in the etcd-operator. Is it ok if I open a new dedicated issue (since I have in mind different option with different pros and cons)?

xiang90:
> Is it ok if I open a new dedicated issue (since I have in mind different option with different pros and cons)?

sure.

georgekuruvillak:
@sgotti I had a query regarding keeping restartPolicy to Always. I would like to know if a data-corruption in etcd data volume will result in the pod getting restarted unnecessarily. Would it not be better if we can handle it using the reconciliation logic? We could know the memberId thats down and in turn the pvc associated. Could we not create the pod with the same memberId and pvc attached? 

While [replacing a failed member](https://kubernetes.io/docs/tasks/administer-cluster/configure-upgrade-etcd/#replacing-a-failed-etcd-member), its mentioned that we remove the instance and add a new instance. Just restarting the pod does not conform to what is mentioned in this documentation.

sgotti:
> Is it ok if I open a new dedicated issue (since I have in mind different option with different pros and cons)?
> sure.

I wanted to propose to directly use StorageClass in the spec instead of the `--pv-provisioner` option but I noticed this has been already accepted during my vacations in #1361. That's really good. So I just went ahead opening PR #1373 to discuss just spec changes.

sgotti:
> @sgotti I had a query regarding keeping restartPolicy to Always. I would like to know if a data-corruption in etcd data volume will result in the pod getting restarted unnecessarily. 

I agree that adding etcd data to a persistent volume adds the requirement to also handle a data corruptions case that previously wasn't needed.

> Would it not be better if we can handle it using the reconciliation logic? We could know the memberId thats down and in turn the pvc associated. Could we not create the pod with the same memberId and pvc attached?

I think that restartPolicy Always is a fast way (as proposed by @xiang90 ) to use k8s features to restart a pod if it has crashed or the node reboots without adding this logic to the reconcile loop. Another way will be as you proposed to do it inside the operator reconcile loop.

But I don't see how this could help detecting corrupted data vs pod/etcd not starting for other reasons. I'm not sure which is the best way to detect a corruption.

My intial idea was to just remove and add a new member (with a new empty PV) if it's not healthy for some time. This will work also with restartPolicy to true and could be done only if the etcd cluster is healthy so won't create problem if all the etcd pods (or nodes) goes down at the same time achieving the goal of this PR.

Quite probably @hongchaodeng @xiang90 have better ideas.

georgekuruvillak:
@sgotti, @xiang90 Is the first part of the PR without PVC reuse going to be incorporated upstream? When can it be roughly expected to be merged?

sgotti:
closing this in favor of multiple PRs (first is #1373)

PR #1351, opened 2017-08-08
Add Cinder as a backup option

jamiehannaford:
Adds support for Cinder as a PV backup type. 

Like the other PV types, this already assumes the user has a working OpenStack cluster and they've configured kubelet, etc. to use it.

### DIFF
--- a/doc/user/backup_config.md
+++ b/doc/user/backup_config.md
@@ -1,7 +1,7 @@
 # Backup Options Config Guide
 
 In etcd operator, we provide the following options to save cluster backups to:
-- Persistent Volume (PV) on GCE or AWS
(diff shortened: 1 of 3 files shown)


hongchaodeng:
`pv-provisioner` flag would be deprecated in the future. Instead, we should go the path of storage class per cluster spec. See https://github.com/coreos/etcd-operator/issues/953

xiang90:
@jamiehannaford do we still need this as the storage class is supported?

everett-toews:
@xiang90 Jamie is out for a couple of weeks on vacation. I chatted with him yesterday and he mentioned that #949 supersedes this. I'm pretty sure it can be closed and Jamie can always reopen it later on if necessary.

PR #1352, opened 2017-08-09
Pass EtcdEnv vars through to the backup pod spec

jswoods:
This allows passing through env vars to the backup pod by using the existing pod policy.

The reason for this is that I run our clusters with a non-standard `KUBERNETES_SERVICE_HOST`. I understand I may also be able to override the `-master` flag, but this seems more flexible.

### DIFF
--- a/pkg/util/k8sutil/pod_util.go
+++ b/pkg/util/k8sutil/pod_util.go
@@ -153,6 +153,14 @@ func applyPodPolicyToPodTemplateSpec(clusterName string, pod *v1.PodTemplateSpec
 	}
 
 	mergeLabels(pod.Labels, policy.Labels)
+
+	if len(policy.EtcdEnv) != 0 {
+		for i := range pod.Spec.Containers {
+			if pod.Spec.Containers[i].Name == "backup" {
+				pod.Spec.Containers[i].Env = append(pod.Spec.Containers[i].Env, policy.EtcdEnv...)
+			}
+		}
+	}
 }
 
 // IsPodReady returns false if the Pod Status is nil


xiang90:
@jswoods EtcdEnv is for the etcd pod as documented. I do not really want to add a backupPod env just for this use case. But I have not came up with a better suggestion yet. Do you have any other potential solutions?

jswoods:
@xiang90 While I see what you're saying, PodPolicy is already being used to override the backup pod here https://github.com/coreos/etcd-operator/blob/master/pkg/util/k8sutil/pod_util.go#L148-L153. Possibly we could put overrides under the BackupPolicy instead? E.g. BackupPolicy.EtcdEnv? Though that would lead to some duplicate code, and it may be better to keep the overrides under a generic pod policy.

I do see how it will be necessary to override various parts of the backup pod (the override of node selectors and tolerations are already allowed). Environment variable overrides will be important anytime anyone needs more granular control the backup pod such as in my case.

I think the core issue here is that only certain attributes of the pod spec are exposed. Ideally, we could easily override any part of the pod spec for either the etcd pods or the back up pods. Otherwise you could be fighting this every time anyone wants to set anything custom on the pods. Resources, mounts, sidecar containers, etc. Perhaps we could have a default pod spec but then have the ability set overrides by merging the entire v1.Pod with the default one (either for etcd or the backup pod). e.g.

```
backup:
  podSpec:
    affinity: ..
    tolerations: [...]
    containers:
    - name: backup
       image: ...
       env:
       - name: ...
          value: ...
```

And then whatever is there is merged with the default pod spec for the backup pod. The sort of odd thing would be specifying the container name but that may be a small price to pay to avoid managing individual overrides of the pod spec.

Either way I'm hoping you will consider this PR for a least a short term approach to exposing environment variables such as how node selectors and tolerations are currently done.

xiang90:
backup pod concept is removed from etcd operator. so this is not needed.

PR #1354, opened 2017-08-10
Add global storageclass flag

jamiehannaford:
This PR supersedes https://github.com/coreos/etcd-operator/pull/949 and introduces a new global `--storageclass` flag. If the user wants to use `--storageclass`, they must set `--pv-provisioner=none`.

It's also worth noting that this PR doesn't implement cluster-specific storage classes, just globally. This can be done at a later date.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Finally, it is safe to upgrade operator. It's highly recommended to save a backu
 
 ### Added
 
+- A new `--storageclass` flag exists, allowing more granular control over how etcd clusters are backed up to PVs.
+
 ### Changed
 
 - Default timeout for snapshots done by backup sidecar increased from 5 seconds to 1 minute
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -53,6 +53,7 @@ import (
 var (
 	analyticsEnabled bool
 	pvProvisioner    string
+	storageClass     string
 	namespace        string
 	name             string
 	awsSecret        string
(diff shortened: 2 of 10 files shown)


hongchaodeng:
Hi @jamiehannaford 

We shouldn't add a new global operator flag. See my comment at https://github.com/coreos/etcd-operator/issues/953#issuecomment-305031796 . Instead, it is preferred to have per cluster spec option.

For the PR, I would recommend a discussion on API design first before implementing it. Either submit a API change PR or create a new issue to discuss it.

jamiehannaford:
@hongchaodeng Ah okay, sorry, I misunderstood your [comment](https://github.com/coreos/etcd-operator/pull/949#issuecomment-321012124). I'll try to get another PR submitted this week for per cluster storage classes.

In terms of this PR, maybe per cluster spec would be better and this can be closed. The only use case I can imagine a global flag is useful for is as a default storage class when users forget to (or don't want to) specify them in a cluster spec.

jamiehannaford:
Also, is it worth deprecating `--pv-provisioner` in the same PR?

hongchaodeng:
> Also, is it worth deprecating --pv-provisioner in the same PR?

Let's deal with deprecation later and focus on enabling storable class in cluster spec.

hongchaodeng:
> The only use case I can imagine a global flag is useful for is as a default storage class when users forget to (or don't want to) specify them in a cluster spec.

We provide both options for S3. But over time, we figured that we can't really align both of them for the same user demand and feature set. And most of time, users would just use the cluster spec thing. I assume the duplication isn't that trouble.

jamiehannaford:
👍

jamiehannaford:
@hongchaodeng Now that I've submitted a new PR, shall I close this one?

hongchaodeng:
@jamiehannaford 
Closing. Thanks!

PR #1358, opened 2017-08-11
Implement StorageClass spec field

### LINKED ISSUE #953: Enable StorageClass as option

jamiehannaford:
Adds a `StorageClass` field to the `BackupPolicy` spec. Fixes #953.

Although I've verified this works, I didn't add any e2e tests because I wasn't sure whether your CI cluster supports any custom StorageClasses.

### DIFF
--- a/pkg/cluster/backup_manager.go
+++ b/pkg/cluster/backup_manager.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"path"
 	"time"
 
 	"github.com/coreos/etcd-operator/client/experimentalclient"
@@ -77,10 +78,14 @@ func (bm *backupManager) setupStorage() (s backupstorage.Storage, err error) {
 	b := cl.Spec.Backup
 	switch b.StorageType {
 	case spec.BackupStorageTypePersistentVolume, spec.BackupStorageTypeDefault:
-		if c.PVProvisioner == constants.PVProvisionerNone {
-			return nil, errNoPVForBackup
+		storageClass := b.PV.StorageClass
+		if len(storageClass) == 0 {
+			if c.PVProvisioner == constants.PVProvisionerNone {
+				return nil, errNoPVForBackup
+			}
+			storageClass = k8sutil.StorageClassPrefix + "-" + path.Base(c.PVProvisioner)
 		}
-		s, err = backupstorage.NewPVStorage(c.KubeCli, cl.Name, cl.Namespace, c.PVProvisioner, *b)
+		s, err = backupstorage.NewPVStorage(c.KubeCli, cl.Name, cl.Namespace, storageClass, *b)
 	case spec.BackupStorageTypeS3:
 		if len(c.S3Context.AWSConfig) == 0 && b.S3 == nil {
 			return nil, errNoS3ConfigForBackup
--- a/pkg/util/k8sutil/backup.go
+++ b/pkg/util/k8sutil/backup.go
@@ -36,7 +36,9 @@ import (
 )
 
 const (
-	storageClassPrefix        = "etcd-backup"
+	// StorageClassPrefix is the prefix used when creating custom storage classes
+	// for backups through a given provisioner.
+	StorageClassPrefix        = "etcd-backup"
 	BackupPodSelectorAppField = "etcd_backup_tool"
 	backupPVVolName           = "etcd-backup-storage"
 	awsCredentialDir          = "/root/.aws/"
@@ -50,7 +52,7 @@ const (
 
 func CreateStorageClass(kubecli kubernetes.Interface, pvProvisioner string) error {
 	// We need to get rid of prefix because naming doesn't support "/".
-	name := storageClassPrefix + "-" + path.Base(pvProvisioner)
+	name := StorageClassPrefix + "-" + path.Base(pvProvisioner)
 	class := &v1beta1storage.StorageClass{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: name,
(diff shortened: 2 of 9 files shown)


pkg/spec/backup.go
```diff
@@ -39,6 +39,12 @@ type BackupPolicy struct {
 
 	StorageSource `json:",inline"`
 
+	// StorageClass indicates what Kubernetes storage class will be used to
+	// snapshot etcd cluster state to a persistent volume. This enables the user
+	// to have fine-grained control over how backups work, since it uses the
+	// existing StorageClass mechanism in Kubernetes.
+	StorageClass string `json:",inline"`
```
xiang90: change the json tag to storageClass?

pkg/cluster/backup_manager.go
```diff
@@ -77,10 +77,10 @@ func (bm *backupManager) setupStorage() (s backupstorage.Storage, err error) {
 	b := cl.Spec.Backup
 	switch b.StorageType {
 	case spec.BackupStorageTypePersistentVolume, spec.BackupStorageTypeDefault:
-		if c.PVProvisioner == constants.PVProvisionerNone {
+		if c.PVProvisioner == constants.PVProvisionerNone && b.StorageClass == "" {
 			return nil, errNoPVForBackup
 		}
-		s, err = backupstorage.NewPVStorage(c.KubeCli, cl.Name, cl.Namespace, c.PVProvisioner, *b)
+		s, err = backupstorage.NewPVStorage(c.KubeCli, cl.Name, cl.Namespace, c.PVProvisioner, b.StorageClass, *b)
```
xiang90: probably convert PVProvisioner to storage class here before passing into the NewPVStorage func.

NewPVStorage can only take the storage class as a parameter.

pkg/spec/backup.go
xiang90: this should be put inside PVSource

pkg/cluster/backup_manager.go
hongchaodeng: So probably when getting a spec, check if StorageClass exists -- if not, default it to global PVProvisioner. In the future, we can also return an error.

hongchaodeng:
@jamiehannaford 
The change generally looks good. There are some comments on implementation details. I'm wondering if you mind separating out changes of pkg/cluster/backup_manager.go, pkg/cluster/backupstorage/pv.go , pkg/util/k8sutil/backup.go into another PR, and only keep docs, spec changes in this PR? We can then merge this PR first.

pkg/cluster/backup_manager.go
jamiehannaford: SGTM

jamiehannaford:
@hongchaodeng No, that's fine with me. I'll open a new PR for the spec and docs, then when that's merged, I'll rebase with all the other needed changes.

xiang90:
@jamiehannaford can you rebase this over https://github.com/coreos/etcd-operator/pull/1361? Thank you!

jamiehannaford:
@xiang90 @hongchaodeng Rebased and RFR

xiang90:
@jamiehannaford can you add a test with storageClass specified? thanks!

pkg/util/k8sutil/backup.go
```diff
@@ -36,7 +36,7 @@ import (
 )
 
 const (
-	storageClassPrefix        = "etcd-backup"
+	StorageClassPrefix        = "etcd-backup"
```
xiang90: doc string on public field (i know other public fields are undocumented, but we need to improve this over time as the operator stabilizes)

thank you!

xiang90:
@jamiehannaford The implementation looks great. thanks for getting this done!

hongchaodeng:
@jamiehannaford 
You can assume our e2e has a storageclass called `standard`.
I'm fine to add the test in another PR (maybe better)? Take a look at `test/e2e/recovery_test.go` at `TestDisasterRecoveryMaj()`.

pkg/cluster/backup_manager.go
```diff
@@ -77,10 +78,14 @@ func (bm *backupManager) setupStorage() (s backupstorage.Storage, err error) {
 	b := cl.Spec.Backup
 	switch b.StorageType {
 	case spec.BackupStorageTypePersistentVolume, spec.BackupStorageTypeDefault:
-		if c.PVProvisioner == constants.PVProvisionerNone {
-			return nil, errNoPVForBackup
+		storageClass := b.PV.StorageClass
+		if storageClass == "" {
```
hongchaodeng: nit:
```
if len(storageClass) == 0 {
```

pkg/util/k8sutil/backup.go
```diff
@@ -36,7 +36,9 @@ import (
 )
 
 const (
-	storageClassPrefix        = "etcd-backup"
+	// StorageClassPrefix is the prefix used when creating custom storage classes
+	// used in backup operations
```
xiang90: StorageClassPrefix is the prefix used when creating custom storage classes for backups through a given provisioner.

xiang90:
lgtm. defer to @hongchaodeng

hongchaodeng:
@jamiehannaford 
lgtm after the nit too.

One suggestion: just pick the test commit (https://github.com/coreos/etcd-operator/pull/1362) here. It looks good. I will merge the two at once.

jamiehannaford:
@hongchaodeng This all ready to go

xiang90:
@jamiehannaford Thanks for the work and contribution! Merging as @hongchaodeng already approved.

PR #1360, opened 2017-08-12
Add support for etcd 3.2 to backup version matrix

bai:
Currently when I try to restore etc 3.2 backup (e.g. during TPR->CRD upgrade) I get an error that "version 3.2 is not compatible with version 3.2".

This PR adds 3.2 to backup/restore compatibility matrix.

### DIFF
--- a/pkg/backup/version.go
+++ b/pkg/backup/version.go
@@ -24,6 +24,7 @@ import (
 var compatibilityMap = map[string]map[string]struct{}{
 	"3.0": {"2.3": struct{}{}, "3.0": struct{}{}},
 	"3.1": {"3.0": struct{}{}, "3.1": struct{}{}},
+	"3.2": {"3.1": struct{}{}, "3.2": struct{}{}},
 }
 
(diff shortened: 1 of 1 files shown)


pkg/backup/version.go
```diff
@@ -24,6 +24,7 @@ import (
 var compatibilityMap = map[string]map[string]struct{}{
 	"3.0": {"2.3": struct{}{}, "3.0": struct{}{}},
 	"3.1": {"3.0": struct{}{}, "3.1": struct{}{}},
+	"3.2": {"3.2": struct{}{}},
```
hongchaodeng: "3.2": {"3.2": struct{}{}, "3.1": struct{}{}},

hongchaodeng: etcd has one minor version backwards compatibility

bai: Thanks! Updated the commit.

hongchaodeng:
lgtm.
Thanks @bai

PR #1237, opened 2017-06-29
backend: add Azure Blob Storage (ABS) backup support

vdice:
This PR adds Azure Blob Storage (ABS) backup support.

Accomplishes part of https://github.com/coreos/etcd-operator/issues/1110

[skip ci]

PR #1369, opened 2017-08-25
feat(abs): add operator logic

vdice:
Ref https://github.com/coreos/etcd-operator/pull/1237

[skip ci]

### DIFF
--- a/pkg/cluster/backupstorage/abs.go
+++ b/pkg/cluster/backupstorage/abs.go
@@ -1,0 +1,104 @@
+// Copyright 2017 The etcd-operator Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package backupstorage
+
+import (
+	"os"
+	"path"
+
+	backupabs "github.com/coreos/etcd-operator/pkg/backup/abs"
+	"github.com/coreos/etcd-operator/pkg/backup/abs/absconfig"
+	"github.com/coreos/etcd-operator/pkg/spec"
+
+	"github.com/coreos/etcd-operator/pkg/backup/env"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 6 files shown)


pkg/cluster/backupstorage/abs.go
```diff
@@ -0,0 +1,90 @@
...
+	return nil
+}
+
+func setupABSCreds(kubecli kubernetes.Interface, ns, secret string) error {
+	se, err := kubecli.CoreV1().Secrets(ns).Get(secret, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	os.Setenv(env.ABSStorageAccount, string(se.Data[spec.ABSStorageAccount]))
```
vdice: @hongchaodeng migrating your concern to this PR.  Looking for the advised practice to get these values to  the function that consumes them in `pkg/backup/abs/abs.go` rather than setting the env here.  Perhaps create an `ABSSecret` struct to hold this data and pass along?

hongchaodeng: Right.

hongchaodeng: Right

pkg/cluster/backupstorage/abs.go
```diff
@@ -0,0 +1,90 @@
+package backupstorage
```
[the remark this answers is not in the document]
vdice: thanks; added.

pkg/util/k8sutil/backup.go
```diff
@@ -183,6 +184,52 @@ func AttachOperatorS3ToPodSpec(ps *v1.PodSpec, s3Ctx s3config.S3Context) {
...
+		Name:      backupenv.ABSStorageKey,
+		ValueFrom: &v1.EnvVarSource{SecretKeyRef: &storageKeySelector},
+	}, v1.EnvVar{
+		Name:  backupenv.ABSContainer,
+		Value: ws.ABSContainer,
+	})
+}
+
+// AttachOperatorABSToPodSpec attaches an operator's ABS credentails to a Pod
+func AttachOperatorABSToPodSpec(ps *v1.PodSpec, absCtx absconfig.ABSContext) {
```
xiang90: i thought we removed the operator wide ABS credential?

vdice: Oops good call; I thought we had as well 🤔  
Now fixed up; thanks!

xiang90:
LGTM.

hongchaodeng:
This is not reviewed fully and should not be merged.

- There is a Validate() func which is not used.
- Second, there are a lot of funcs using `os.Getenv` which isn't thread safe. [This comment](https://github.com/coreos/etcd-operator/pull/1369#discussion_r135367058) means we can jump back to this once two dep PRs are merged (docs & spec, backup sidecar).

PR #1370, opened 2017-08-25
feat(abs): add tests

vdice:
Ref https://github.com/coreos/etcd-operator/pull/1237

### DIFF
--- a/pkg/backup/abs_backend_test.go
+++ b/pkg/backup/abs_backend_test.go
@@ -1,0 +1,219 @@
+// Copyright 2017 The etcd-operator Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package backup
+
+import (
+	"bytes"
+	"crypto/rand"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"reflect"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/storage"
+	"github.com/coreos/etcd-operator/pkg/backup/abs"
+)
+
+const integrationTestEnvVar = "RUN_INTEGRATION_TEST"
+
+var (
+	accountName    = storage.StorageEmulatorAccountName
+	accountKey     = storage.StorageEmulatorAccountKey
+	DefaultBaseURL = "http://127.0.0.1:10000"
+	prefix         = "testprefix"
+	blobContents   = "ignore"
+
+	absIntegrationTestNotSet = fmt.Sprintf("skipping ABS integration test due to %s not set", integrationTestEnvVar)
+)
+
+func generateRandomContainerName() (string, error) {
+	n := 5
+	b := make([]byte, n)
+	if _, err := rand.Read(b); err != nil {
+		return "", err
+	}
+	randContainerName := fmt.Sprintf("testcontainer-%X", b)
+
+	return randContainerName, nil
+}
+
+func TestABSBackendContainerDoesNotExist(t *testing.T) {
+	if os.Getenv(integrationTestEnvVar) != "true" {
+		t.Skip(absIntegrationTestNotSet)
+	}
+
+	container, err := generateRandomContainerName()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	storageClient, err := storage.NewClient(accountName, accountKey, DefaultBaseURL, "", false)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, err = abs.NewFromClient(container, prefix, &storageClient)
+	if err == nil {
+		t.Fatal(err)
+	}
+	if err.Error() != fmt.Sprintf("container %s does not exist", container) {
+		t.Fatal(err)
+	}
+}
+
+func TestABSBackendGetLatest(t *testing.T) {
+	if os.Getenv(integrationTestEnvVar) != "true" {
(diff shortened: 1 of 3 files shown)


hongchaodeng:
@vdice 
Can you give instructions for what to set up in test infra?

vdice:
@hongchaodeng we just need to run the azurite container as a prerequisite to executing the tests w/ `RUN_ABS_INTEGRATION_TEST` turned on:

```
 $ docker run -d -t -p 10000:10000 quay.io/vdice/azurite
 $ RUN_ABS_INTEGRATION_TEST=true go test -v ./pkg/backup
```

This is how I've been testing locally.  Let me know if this works out!

vdice:
PS the official azure image can be run as well via https://github.com/arafato/azurite#pulling-from-docker-hub

For some reason, at the time of development and testing, I couldn't pull the image... but appears to work now.

pkg/backup/abs_backend_test.go
```diff
@@ -0,0 +1,197 @@
...
+	"os"
+	"reflect"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/storage"
+	"github.com/coreos/etcd-operator/pkg/backup/abs"
+)
+
+const (
+	absIntegrationTestEnvVar = "RUN_ABS_INTEGRATION_TEST"
```
fanminshi: `RUN_ABS_INTEGRATION_TEST` -> `RUN_INTEGRATION_TEST`
I don't think we should have each ENV flag for each type of integration test.

pkg/backup/abs_backend_test.go
```diff
@@ -0,0 +1,197 @@
...
+	blobServiceClient := storageClient.GetBlobService()
+
+	// Create container
+	cnt := blobServiceClient.GetContainerReference(container)
+	options := storage.CreateContainerOptions{
+		Access: storage.ContainerAccessTypePrivate,
+	}
+	_, err = cnt.CreateIfNotExists(&options)
+	if err != nil {
+		if accountName == storage.StorageEmulatorAccountName {
```
fanminshi: isn't `if accountName == storage.StorageEmulatorAccountName` always true?
what's this check for?

pkg/backup/abs_backend_test.go
```diff
@@ -0,0 +1,197 @@
...
+		t.Fatal(err)
+	}
+	defer rc.Close()
+
+	if string(b) != blobContents {
+		t.Errorf("content = %s, want %s", string(b), blobContents)
+	}
+
+	// Delete container
+	opts := storage.DeleteContainerOptions{}
```
fanminshi: probably should have the container delete logic under a defer function just below the container creation logic. Therefore, if any checking condition fails, container is deleted as well.

```go
_, err = cnt.CreateIfNotExists(&options)
if err != nil {...}
defer func() {
  // Delete container
  opts := storage.DeleteContainerOptions{}
  if err := cnt.Delete(&opts); err != nil {
    t.Fatal(err)
  }
}()
```

pkg/backup/abs_backend_test.go
```diff
@@ -0,0 +1,197 @@
...
+	blobServiceClient := storageClient.GetBlobService()
+
+	// Create container
+	cnt := blobServiceClient.GetContainerReference(container)
+	options := storage.CreateContainerOptions{
+		Access: storage.ContainerAccessTypePrivate,
+	}
+	_, err = cnt.CreateIfNotExists(&options)
+	if err != nil {
+		if accountName == storage.StorageEmulatorAccountName {
```
fanminshi: same concern about `accountName == storage.StorageEmulatorAccountName ` as above.

pkg/backup/abs_backend_test.go
```diff
@@ -0,0 +1,197 @@
...
+	leftFiles := []string{makeBackupName("3.1.0", 2)}
+	if !reflect.DeepEqual(leftFiles, names) {
+		t.Errorf("left files after purge, want=%v, get=%v", leftFiles, names)
+	}
+	if err := abs.Delete(makeBackupName("3.1.0", 2)); err != nil {
+		t.Fatal(err)
+	}
+
+	// Delete container
+	opts := storage.DeleteContainerOptions{}
```
fanminshi: wrap delete in a `defer func` just below container creation?

fanminshi:
@vdice I am thinking about creation test container, ` docker run -d -t -p 10000:10000 quay.io/vdice/azurite`, in the testing script itself; azure test container initializes before test run and deletes after test finishes. e.g:

Before test run:
 - check if test azure container exists, if yes, then delete it.
 - start the azure container.
 - exit with error if any of above fails.

After test run:
 - delete azure container.
 - exit with error if deletion fails.

pkg/backup/abs_backend_test.go
vdice: Good call; not needed.

pkg/backup/abs_backend_test.go
vdice: Thank you; amended!

pkg/backup/abs_backend_test.go
```diff
@@ -0,0 +1,193 @@
...
+	"github.com/coreos/etcd-operator/pkg/backup/abs"
+)
+
+const integrationTestEnvVar = "RUN_INTEGRATION_TEST"
+
+var (
+	accountName    = storage.StorageEmulatorAccountName
+	accountKey     = storage.StorageEmulatorAccountKey
+	DefaultBaseURL = "http://127.0.0.1:10000"
+	container      = "testcontainer"
```
fanminshi: I am going to make the Azuite test container be shared among all test builds; Therefore, we need to randomized `container` to avoid test collision.

fanminshi:
one small nit, the pr then looks good to me.

vdice:
@fanminshi logic to randomize test container name added.

fanminshi:
lgtm.

fanminshi:
@vdice could you rebase the pr? the azuite test container pr is merged.

vdice:
@fanminshi rebased; ci looks good?

hongchaodeng:
lgtm

fanminshi:
test passed. @vdice thanks for the pr!

PR #1373, opened 2017-08-29
spec: add PVSource to PodPolicy

sgotti:
As requested by @hongchaodeng here (https://github.com/coreos/etcd-operator/pull/1349#issuecomment-321013872) I opened this PR to discuss only spec changes to add pv support for etcd data.

I wanted to propose to directly use StorageClass in the spec instead of the `--pv-provisioner` option but I noticed this has been already accepted during my vacations in #1361. So I used it, moving it from spec/backup.go to spec/cluster.go and added PVSource also to PodPolicy.

This patch adds the initial spec changes (without implementation) to
support storing etcd data inside a persistent volume.

The PVSource type used by the backups policy is now shared between
backups and pod policies.

Since this only a spec change the required spec validation isn't part of this patch.

### DIFF
--- a/pkg/spec/backup.go
+++ b/pkg/spec/backup.go
@@ -67,24 +67,14 @@ func (bp *BackupPolicy) Validate() error {
 }
 
 type StorageSource struct {
+	// PV represents a Persistent Volume resource, operator will claim the
+	// required size before creating the etcd cluster for backup purpose.
+	// If the snapshot size is larger than the size specified operator would
+	// kill the cluster and report failure condition in status.
 	PV *PVSource `json:"pv,omitempty"`
 	S3 *S3Source `json:"s3,omitempty"`
 	// ABS represents an Azure Blob Storage resource for storing etcd backups
 	ABS *ABSSource `json:"abs,omitempty"`
-}
-
-type PVSource struct {
-	// VolumeSizeInMB specifies the required volume size to perform backups.
-	// Operator will claim the required size before creating the etcd cluster for backup
-	// purpose.
-	// If the snapshot size is larger than the size specified, backup fails.
-	VolumeSizeInMB int `json:"volumeSizeInMB"`
-
-	// StorageClass indicates what Kubernetes storage class will be used to
-	// snapshot etcd cluster state to a persistent volume. This enables the user
-	// to have fine-grained control over how backups work, since it uses the
-	// existing StorageClass mechanism in Kubernetes.
-	StorageClass string `json:"storageClass"`
 }
 
 // TODO: support per cluster S3 Source configuration.
--- a/pkg/spec/cluster.go
+++ b/pkg/spec/cluster.go
@@ -62,6 +62,17 @@ func (c *EtcdCluster) AsOwner() metav1.OwnerReference {
 	}
 }
 
+type PVSource struct {
+	// VolumeSizeInMB specifies the required volume size.
+	VolumeSizeInMB int `json:"volumeSizeInMB"`
+
+	// StorageClass indicates what Kubernetes storage class will be used.
+	// This enables the user to have fine-grained control over how persistent
+	// volumes are created since it uses the existing StorageClass mechanism in
+	// Kubernetes.
+	StorageClass string `json:"storageClass"`
+}
+
 type ClusterSpec struct {
 	// Size is the expected size of the etcd cluster.
 	// The etcd-operator will eventually make the size of the running
@@ -154,6 +165,11 @@ type PodPolicy struct {
 	// bootstrap the cluster (for example `--initial-cluster` flag).
 	// This field cannot be updated.
 	EtcdEnv []v1.EnvVar `json:"etcdEnv,omitempty"`
+
+	// PV represents a Persistent Volume resource.
+	// If defined new pods will use a persistent volume to store etcd data.
+	// TODO(sgotti) unimplemented
+	PV *PVSource `json:"pv,omitempty"`
 }
 
 func (c *ClusterSpec) Validate() error {


pkg/spec/backup.go
```diff
@@ -66,26 +66,15 @@ func (bp *BackupPolicy) Validate() error {
 }
 
 type StorageSource struct {
+	// PV represents a Persistent Volume resource, operator will claim the
+	// required size before creating the etcd cluster for backup purpose.
```
hongchaodeng: ~~for backup purpose~~ .

This is not true after we have PV support for etcd pod.

pkg/spec/backup.go
```diff
@@ -66,26 +66,15 @@ func (bp *BackupPolicy) Validate() error {
...
+	// If the snapshot size is larger than the size specified, backup fails.
```
hongchaodeng: backup fails -> operator would kill the cluster and report failure condition in status.

I guess the right direction is that we would change our ClusterStatus to be more k8s native and add a failure condition, reason for such.

hongchaodeng:
@sgotti Thanks for doing the work!!!

I'm sorry to forget earlier and mention this late. But we need a design doc for this feature. I want to mention explicitly that it is not meant to obstruct you :)

Basically, we follow k8s upstream process to have design written down: https://github.com/kubernetes/community/blob/master/contributors/design-proposals/api-chunking.md

We have much simpler design docs in this repo:
https://github.com/coreos/etcd-operator/blob/master/doc/design/reconciliation.md

We also do design process in multiple internal projects. So it's both open source process and our standard.

## What's in the design doc?

Please at least answer these two questions in design docs:
- How to create and manage (persistent) volumes?
- How to correlate volumes to etcd pods?

You can assume:
- No corruption/failure will happen on PV now. Basically to avoid tolerations. We would need to extend the design docs or create another for the problem.

pkg/spec/cluster.go
```diff
@@ -112,6 +123,10 @@ type ClusterSpec struct {
 
 	// etcd cluster TLS configuration
 	TLS *TLSPolicy `json:"TLS,omitempty"`
+
+	// PV represents a Persistent Volume resource.
+	// If defined new pods will use a persistent volume to store etcd data.
```
xiang90: // If defined pods will use a persistent volume to store etcd data.
// When pod/node fails, the replacement pod will try to reuse the existing PV.
// When scale up, ...
// When scale down,

Basically, we probably need to describe the pv claime/cleanup policy a little bit here.

xiang90: Adding a TODO here. this is not supported as it is. We are not very strict about versioning here since etcd operator is not stable yet.

xiang90:
@hongchaodeng @sgotti 

I am fine with merging the spec as it is. But, yes, we need to have a design doc along with the implementation. it is easier for other people to understand how things work, and help to debug issues.

sgotti:
@hongchaodeng @xiang90 Created the design doc: #1374. While writing it I noticed it was difficult to just put the first part in it because you won't see the final goal and what gains persistent volumes will bring with only that part. So I preferred putting down both parts and all the related notes.

pkg/spec/backup.go
sgotti: That comment is for `StorageSource` that is a backupPolicy only option.

pkg/spec/backup.go
sgotti: I just copied and moved the corrent comment from PVSource to StorageSource. Are you asking for some changes in it in this patch?

pkg/spec/cluster.go
sgotti: > Basically, we probably need to describe the pv claime/cleanup policy a little bit here.

Ok. Perhaps lets wait when the proposal (that talks about these logics) is accepted?

> Adding a TODO here. this is not supported as it is. We are not very strict about versioning here since etcd operator is not stable yet.

That probably the problem on splitting spec and implementation changes. Are you asking to a TODO here and removing in the implementation PR?

xiang90: yea. just a simple TODO in the comment. and remove it in the implementation.

pkg/spec/backup.go
hongchaodeng: ok

pkg/spec/backup.go
hongchaodeng: yes.

sgotti:
Updated adding an "unimplemented" TODO. About https://github.com/coreos/etcd-operator/pull/1373#discussion_r135902026 It's unclear to me what comment changes should I do.

pkg/spec/backup.go
hongchaodeng: Yeah.. Please change it to
```
operator would kill the cluster and report failure condition in status.
```

hongchaodeng:
lgtm after nit

pkg/spec/backup.go
sgotti: @hongchaodeng thanks, updated. It wasn't clear to me if you just wanted to update the comment (I wasn't aware of this behaviour) or you were talking about a future change and how it was related to this patch.

pkg/spec/cluster.go
```diff
@@ -154,6 +165,11 @@ type PodPolicy struct {
 	// bootstrap the cluster (for example `--initial-cluster` flag).
 	// This field cannot be updated.
 	EtcdEnv []v1.EnvVar `json:"etcdEnv,omitempty"`
+
+	// PV represents a Persistent Volume resource.
+	// If defined new pods will use a persistent volume to store etcd data.
+	// TODO(sgotti) unimplemented
+	PV *PVSource `json:"pv,omitempty"`
```
xiang90: @sgotti 

one thing to note here, the backup pod also need to respect (or explicitly ignore this) this in the future. right now, it saves intermediate snapshot data onto its container fs.

xiang90:
lgtm

PR #1374, opened 2017-08-30
doc: persistent volumes for etcd data proposal

sgotti:
Since the final scope is to achieve a durable etcd cluster (#1323) I put in this proposal the changes split in two parts (as requested). IMHO putting only the first part is not really useful because it will hide many detail of the overall idea. For example putting RestartPolicy to Always can bring a little gain for part 1 but could be not useful (or add more issues) for part 2.

I also added some notes since the use of persistence volume requires the knowledge of other k8s logic (https://github.com/kubernetes/community/blob/master/contributors/design-proposals/pod-safety.md) because etcd operator is acting as a pod controller.

[skip ci]

### DIFF
--- a/doc/design/persistent_volumes_etcd_data.md
+++ b/doc/design/persistent_volumes_etcd_data.md
@@ -1,0 +1,112 @@
+# Use Persistent Volumes for etcd data
+
+## Proposal and Motivation
+
+Today etcd-operator creates ephemeral etcd members: in their pods the etcd data is stored inside a volume of type emptyDir and the pod restart policy is set to `Never`.
+This has the following failure cases.
+
+### Failure cases
+
+Assume a N members etcd cluster with all members healthy
+
+#### Case A
+
+##### Event
+
+One of these events happens:
+* one etcd member process exits (processes crashes, killed etc...)
+* a k8s node executing one etcd member pod reboots and comes back before node controller pod eviction triggers
+
+##### Effect
+
+The pod will go in failed state, the etcd cluster remains quorate. Etcd-operator will remove the member from the etcd cluster and add a new member to the cluster and creates a new pod with a new name.
+
+#### Case B
+
+##### Event
+
+* a k8s node executing one etcd member pod becomes partitioned (network partitioning, goes down, etc...) and cannot talk to the api server, after some time the node controller will evict its pods.
+
+##### Effect
+
+The pod will go in an unknown/failed state, the etcd cluster remains quorate. Etcd-operator will remove the member from the etcd cluster and add a new member to the cluster and creates a new pod with a new name.
+
+#### Case C
+
+##### Event
+
+One of these events happens:
+* a majority of etcd processes exits (processes crashes, killed etc...) 
+* k8s nodes scheduling a majority of etcd members pods reboots and come back before node controller pod eviction triggers
+
+##### Effect
+
+The pods will go in failed state, the etcd cluster becomes unquorate. Etcd-operator can only restore the cluster from a backup.
+
+#### Case D
+
+##### Event
+
+* k8s nodes executing a majority of etcd members pods becomes partitioned (network partitioning, goes down, etc...) and cannot talk to the api server, after some time the node controller will evict their pods.
+
+##### Effect
+
+The pods will go in an unknown/failed state and eventually be removed from the API server, the etcd cluster becomes unquorate. Etcd-operator can only restore the cluster from a backup.
+
+
+### Current failure handling
+
+Currently etcd operator handles cases A and B but when a cluster becomes unquorate (cases C and D) the current option is to restore the cluster from a backup.
+
+User may desire consistency over availability and prefer to wait for the cluster the return in a quorate state instead of restoring from a backup (that will contain old data).
+
+This proposal aims to fix cases C and D permitting to the cluster the return in a quorate state without the need to restore from a backup.
+
+## Proposed Changes
+
+The design to be effective is split in two parts.
+
+### Part 1
+
+These changes will put the basis for initial persistent volume management
+
+This first part won't fix case C and D.
+
+- Add to the cluster spec PodPolicy an option to enable putting etcd data inside a Persistent Volume
+- If podPolicy has persistent volumes enabled, when creating a new member also create a Persistent Volume Claim and use it as the volume source for etcd-data
+- Handle etcd data persistent volume claims cleanup:
+  - set PVC ownerRef to EtcdCluster object, which enables GC to clean up if cluster is deleted.
+  - PVC lifecycle is bound to etcd member's lifecycle. PVC is added/removed when etcd member is added/removed (in the implementation we could choose to remove PVC only if there's not pod referencing them to avoid strange behaviors)
+
+#### Pod name to Persistent Volume Claim name mapping
+
+The pod and related PVC are mapped 1 to 1 so their names be easily generated and deducted. I.E. if the pod is called `etcd-0000` its related PVC name will be `etcd-0000-pvc`
+
+
+### Part 2
+
+Using persistent volumes to store etcd data will add additional failure cases:
+
+- volume failure (failing to bind PV to pod).
+- corrupted data (filesystem ok but missing/corrupted files) that will cause etcd to exit. Previously an etcd member with corrupted data will be deleted and a new member added, using a persistent volume we have to avoid reusing the same corrupted data volume.
+
+This parts will fix cases C and D and the above corrupted data case.
+
+It's a change to the reconcile logic to handle deleted pods from the k8s API server that had a persistent volume. To handle corrupted data failures, the reconcile logic change defined below will be used only if the cluster is unquorate. Instead, if the cluster is quorate, the current logic of deleting the member and adding a new member will be used.
+
+- If the cluster is unquorate and a pod of an etcd cluster member is deleted from the API server and it had a persistent volume claim associated, instead of removing the etcd member from the etcd cluster and adding a new member just recreate it (keeping its pod name) binding the existing persistent volume claim. The pod recreation will happen only if the pod doesn't exists anymore in the API because we'll keep the same pod name. This also enable a MANDATORY property: at most once pod existence logic to avoid having in the API at the same time two pods with associated the same PVC.
+
+
+### Future enhancements
+
+Handle failures introduced by this design and not covered with the previous parts:
+
+- volume failure (failing to bind PV to pod).
+
+
+
+## Related issues
+
+- [Explore Local PV](https://github.com/coreos/etcd-operator/issues/1201)
+- [Persistent/Durable etcd cluster](https://github.com/coreos/etcd-operator/issues/1323)
+- [Pod Safety, Consistency Guarantees, and Storage Implications](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/pod-safety.md)


doc/design/persistent_volumes_etcd_data.md
```diff
@@ -0,0 +1,79 @@
...
+
+Add an option to enable storing etcd data inside a persistent volume and related reconcile logic changes.
+
+The change to be effective and cover the above failure scenario is split in two parts.
+
+### Part 1
+
+- Add to the cluster spec PodPolicy an option to enable putting etcd data inside a Persistent Volume
+- If podPolicy has persistent volumes enabled, when creating a pod also create a Persistent Volume Claim and use it as the volume source for etcd-data
+- Handle etcd data persistent volume garbage collection like already done with other resources. While backup pv aren't garbage collected, etcd data pv needs to be garbage collected when not needed anymore.
```
hongchaodeng: > Handle etcd data persistent volume garbage collection like already done with other resources.

This is different. Operator only creates PVC. It doesn't create PV. So I assume here PV is managed by controller manager. When PV is released, k8s controller will remove it. Please write more details about this process.
 
Add more details about how to cleanup PVC too. If a member is removed, we should also delete its PVC.

> While backup pv aren't garbage collected

Why is it related to backup PV?

doc/design/persistent_volumes_etcd_data.md
```diff
@@ -0,0 +1,79 @@
...
+
+#### Results
+
+These changes will put the basis for initial persistent volume management BUT won't fix the failure scenario described at the start of the proposal: if a majority of pods is deleted there's no logic to recreate them with their previous data.
+
+#### Proposed enhancements to Part 1 (@xiang90)
+
+- When using a PVC as volume source also set pod restart policy to `Always`
+
+With this change when the etcd container of the pod dies due to different reasons (etcd crashing, container killed, node reboots but its pods aren't evicted by the k8s node controller) it will be restarted (since Restart Policy is set to Always). Previously the pod will go in Failed state and will be removed and etcd operator will remove the etcd member and add a new member and a new pod.
```
hongchaodeng: Previously the pod will go in Failed state ~~and will be removed~~ and etcd operator will remove

doc/design/persistent_volumes_etcd_data.md
```diff
@@ -0,0 +1,79 @@
...
+
+- Add to the cluster spec PodPolicy an option to enable putting etcd data inside a Persistent Volume
+- If podPolicy has persistent volumes enabled, when creating a pod also create a Persistent Volume Claim and use it as the volume source for etcd-data
+- Handle etcd data persistent volume garbage collection like already done with other resources. While backup pv aren't garbage collected, etcd data pv needs to be garbage collected when not needed anymore.
+
+#### Pod name to Persistent Volume Claim name mapping
+
+The pod and related PVC are mapped 1 to 1 so thet can be easily generated and deducted. I.E. if the pod is called `etcd-0000` its related PVC name will be `etcd-0000-pvc`
+
+#### Results
```
hongchaodeng: Shouldn't it be **Problems**?

doc/design/persistent_volumes_etcd_data.md
```diff
@@ -0,0 +1,79 @@
...
+
+These changes will put the basis for initial persistent volume management BUT won't fix the failure scenario described at the start of the proposal: if a majority of pods is deleted there's no logic to recreate them with their previous data.
```
hongchaodeng: > if a majority of pods is deleted there's no logic to recreate them with their previous data.

Why need to recreate the cluster?

I think for the first phase we will just wait the failed etcd pods to restart until they form healthy quorum again.

doc/design/persistent_volumes_etcd_data.md
```diff
@@ -0,0 +1,79 @@
...
+With this change when the etcd container of the pod dies due to different reasons (etcd crashing, container killed, node reboots but its pods aren't evicted by the k8s node controller) it will be restarted (since Restart Policy is set to Always). Previously the pod will go in Failed state and will be removed and etcd operator will remove the etcd member and add a new member and a new pod.
+
+
+### Part 2
+
+#### Changes
+
+- Change reconcile logic to handle deleted pods that had a persistent volume:
+
+If the a pod of an etcd cluster member is deleted and it had a persistent volume claim associated, instead of removing the etcd member from the etcd cluster and adding a new member just recreate it (keeping its pod name) binding the existing persistent volume claim. The pod recreation will happen only if the pod doesn't exists anymore in the API because we'll keep the same pod name. This also enable a MANDATORY property: at most once pod existence logic to avoid having in the API at the same time two pods with associated the same PVC.
```
hongchaodeng: > If the a pod of an etcd cluster member is deleted

Please list what scenarios a member will be deleted. For this phase it will be deleted only in scale-down

hongchaodeng: > instead of removing the etcd member from the etcd cluster and adding a new member just recreate it (keeping its pod name) binding the existing persistent volume claim.

For this phase, we assume that we will wait for any failed etcd pod to restart instead of recreating it.

Handling permanent failure is a more complicated topic. Like how to manage membership and how to define failure/tolerations.

doc/design/persistent_volumes_etcd_data.md
```diff
@@ -0,0 +1,79 @@
...
+
+#### Results
+
+With this change if a majority of pods is deleted the etcd-operator will recreate them keeping their previous data, so the above failure scenario will be fixed.
+
+#### Notes
+
+From https://github.com/kubernetes/community/blob/master/contributors/design-proposals/pod-safety.md to avoid data corruptions, the node controller will only mark the partitioned node pods for termination but will not remove them from the api. A fence controller (not yet existing as a standard k8s component) or a manual intervention have to ensure that the node is down and then remove it from the API or force remove all of its pods from the API.
+
+So only when the pod is removed from the API the above logic can recreate it. If this isn't going to happen for a long time, the new etcd pod won't be recreated. The below proposed enhancements will help avoiding such cases.
```
hongchaodeng: This is not what that proposal talks about.. Actually that proposal is not related to this proposal right now.. Remove this section?

doc/design/persistent_volumes_etcd_data.md
sgotti: but the pods can be deleted (removed from the API) by k8s itself (or by the user) regardless of restart policy (which is a pod property) as explained previously in the doc.

doc/design/persistent_volumes_etcd_data.md
sgotti: it's removed from the API server by the kubelet.

doc/design/persistent_volumes_etcd_data.md
sgotti: > For this phase, we assume that we will wait for any failed etcd pod to restart instead of recreating it.

This is part one, but as said pods can be removed from the API and Part 2 aims to recreate them

doc/design/persistent_volumes_etcd_data.md
sgotti: > For this phase, we assume that we will wait for any failed etcd pod to restart instead of recreating it.

This is handled by part one, but as said pods can be removed from the API and Part 2 aims to recreate them.

doc/design/persistent_volumes_etcd_data.md
sgotti: > For this phase, we assume that we will wait for any failed etcd pod to restart instead of recreating it.

This is handled by part one, but as said pods can be removed from the API and Part 2 aims to recreate them.



> Handling permanent failure is a more complicated topic. Like how to manage membership and how to define failure/tolerations.

part 2 isn't for handling permament failures but pods removed from the API.

doc/design/persistent_volumes_etcd_data.md
sgotti: Given the above answers I think this is important since it explains why pods are evicted from dead nodes and why they aren't directly removed by the API but will wait for the node to come back or being removed from the API by an external fencing component.

Since when a majority of etcd pods are removed, to restore quorum, etcd-operator (part 2) has to recreate them with their existing data (PV), it's foundamental that etcd-operator, as a pod controller, behaves correctly:

* don't force delete pods with PV from the API (i.e don't use grace period = 0)
* obtain at most once pod guarantee reusing the previous pod name.

doc/design/persistent_volumes_etcd_data.md
sgotti: forgot to add `claim` fixed and detailed.


> Why is it related to backup PV?

just to be clear that only etcd data PVCs will be GCed, backups PVC won't be touched by this proposal. Will remove it if it's tacit.

doc/design/persistent_volumes_etcd_data.md
sgotti: These are the result obtained with part one, not the final goal, just an initial part.

sgotti:
@hongchaodeng thanks for the review, updated the PVC GC part with more details.

Probably my fault explaining it clearly (from the comments looked like a secondary issue) but I'd like to stress that the fact that a POD (and a majority of them) will be removed from the API is a basic k8s function and will happen and should be handled and this proposal wants to just handle this case to fix the explained issues of no doing this. So IMHO only part 1 won't be enough for me. A Statefulset controller does exactly this, if the pod is removed from the API for all the reason explained above, they'll recreate a new pod with the same name and previous PVCs.

doc/design/persistent_volumes_etcd_data.md
hongchaodeng: > but the pods can be deleted (removed from the API) by k8s itself (or by the user) regardless of restart policy (which is a pod property) as explained previously in the doc.

Yeah. That's definitely something we will handle, but not in phase/part 1.
I see our divergence here. My reasoning is that we want to do good engineering -- that means solve problems gradually and one by one. It's easier to implement, review, and keep on track. Again, general rule we apply to engineering.

doc/design/persistent_volumes_etcd_data.md
hongchaodeng: Got it.

Then maybe ` will be removed from APIServer`?

doc/design/persistent_volumes_etcd_data.md
hongchaodeng: > This is handled by part one, but as said pods can be removed from the API and Part 2 aims to recreate them.

This is mentioned in [this discussion](https://github.com/coreos/etcd-operator/pull/1374/files#r136131112). We will solve them one by one, not coupled together.

> part 2 isn't for handling permanent failures but pods removed from the API.

What is the problem you are trying to solve? Can you give specific failure scenario and differentiation?

doc/design/persistent_volumes_etcd_data.md
hongchaodeng: In the design doc, we will only talk about the actions and the reasoning behind the actions. I can see that:
- Action: we need to handle failure because a pod could be failing permanently and we need to recover them.
- Reasoning: Pods could be evicted, Volume could fail permanently, node could fail permanently.

Other than that, what's the point of this paragraph? What's the problem it is trying to solve?

> Since when a majority of etcd pods are removed, to restore quorum, etcd-operator

So this is a **second** problem from the eviction problem you mentioned. Right? If so, separate them into two problem sections and discuss separately.

doc/design/persistent_volumes_etcd_data.md
sgotti: I added at the start of the proposal 4 example cases used to better explain what I want to solve and the effect of every change on these cases.

doc/design/persistent_volumes_etcd_data.md
sgotti: I hope the 4 cases I depicted will help. Action/Reasoning should be covered by them.

doc/design/persistent_volumes_etcd_data.md
```diff
@@ -0,0 +1,82 @@
+# Use Persistent Volumes for etcd data
+
+## Proposal and Motivation
+
+Today etcd-operator creates ephemeral etcd members: in their pod the etcd data is stored inside a volume of type emptyDir and the pod restart policy is set to `Never`.
+
+So, if a pod is deleted from the API server (manually or automatically: i.e a node goes down and the node controller, after a timeout, evicts pods marking them for termination), etcd-operator will remove the related etcd member from the etcd cluster and will create a new pod (with a new name and an empty data dir) and add that new etcd member to the etcd cluster.
+
+When a majority of pods is deleted at the same time the etcd cluster will become unquorate and there'll be no way to restore its quorum since the members and their data are gone. So the unique solution (that etcd-operator currently do) is to reinitialize a new cluster restoring its data from a backup (if available).
```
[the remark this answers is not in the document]
sgotti: done

doc/design/persistent_volumes_etcd_data.md
```diff
@@ -0,0 +1,82 @@
...
+#### Changes
+
+Add an option to enable storing etcd data inside a persistent volume and related reconcile logic changes.
+
+The change to be effective and cover the above failure scenario is split in two parts.
+
+### Part 1
+
+- Add to the cluster spec PodPolicy an option to enable putting etcd data inside a Persistent Volume
+- If podPolicy has persistent volumes enabled, when creating a pod also create a Persistent Volume Claim and use it as the volume source for etcd-data
```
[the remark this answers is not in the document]
sgotti: done

doc/design/persistent_volumes_etcd_data.md
```diff
@@ -0,0 +1,82 @@
...
+- Handle etcd data persistent volume claims garbage collection:
+  - define on the PVC some labels (`etcd-node`, `etcd-cluster`, `app`) and set an owner like already done with other resources. See https://github.com/coreos/etcd-operator/blob/master/doc/design/resource_ownership_and_GC.md
```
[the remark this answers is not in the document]
sgotti: done

doc/design/persistent_volumes_etcd_data.md
```diff
@@ -0,0 +1,136 @@
...
+#### Problems
+
+This first part won't fix case C and D.
+
+#### Proposed enhancements to Part 1 (@xiang90)
+
+- When using a PVC as volume source also set pod restart policy to `Always`
+
+This will fix case C and change behavior of case A and B. The pod won't go in a failed state but the kubelet will try to restart it indefinetly.
+
```
sgotti: @hongchaodeng @xiang90 Perhaps it's also important to note that setting restart policy to `Always` will work also without persistent volumes since the emptydir is removed only when the pod is deleted (https://kubernetes.io/docs/concepts/storage).

So changing restart policy to `Always` is making this proposal more long and confused (multiple combinations) and will probably create some undesired effects and increasing.

So I was thinking to remove it entirely from the design. What do you think?

hongchaodeng: What if the member is removed? In that case we don't want restart.

hongchaodeng: Just to give more context, when a pod is removed, it will exit 0. But it's also fine if it restart: it will keep failing until kubelet kills it.

I think moving forward we will have all pods to always restart. Let's keep it a separate topic and open a new issue for it.

sgotti: > I think moving forward we will have all pods to always restart. Let's keep it a separate topic and open a new issue for it.

I'm not sure if setting Restart Policy to Always is a good idea but you'll probably have other ideas to handle some cases masked by using it, so if you're going to open a new issue I'll add my opinions there.

> Let's keep it a separate topic 

Good. Will remove all the parts related to changing the current pod Restart Policy.

sgotti:
@hongchaodeng updated removing the parts related to changing the restart policy and adding more problems and enhancements to part 2 to handle them.

doc/design/persistent_volumes_etcd_data.md
```diff
@@ -0,0 +1,125 @@
...
+##### Event
+
+* k8s nodes executing a majority of etcd members pods becomes partitioned (network partitioning, goes down, etc...) and cannot talk to the api server, after some time the node controller will evict their pods.
+
+##### Effect
+
+The pods will go in an unknown/failed state and eventually be removed from the API server, the etcd cluster becomes unquorate. Etcd-operator can only restore the cluster from a backup.
+
+
+#### Reasons
```
hongchaodeng: Reasons for ?

doc/design/persistent_volumes_etcd_data.md
```diff
@@ -0,0 +1,125 @@
+# Use Persistent Volumes for etcd data
+
+## Proposal and Motivation
+
+Today etcd-operator creates ephemeral etcd members: in their pod the etcd data is stored inside a volume of type emptyDir and the pod restart policy is set to `Never`.
+
+This covers some failure cases but won't cover other failure cases.
+
+### Example cases
```
hongchaodeng: Failure cases

doc/design/persistent_volumes_etcd_data.md
```diff
@@ -0,0 +1,125 @@
...
+
+This proposal aims to fix cases C and D permitting to the cluster the return in a quorate state without the need to restore from a backup.
+
+## Proposed Changes
+
+#### Changes
+
+The change to be effective and cover the above failure scenario is split in two parts.
+
+### Part 1
```
hongchaodeng: Some thoughts:

- Remove the above "Reasons" section. Add individual "Analysis" section to each part. For Part 1:
  ```
  For failure cases A and C, we can tolerate pod restart if we use PV to store etcd data.
  ```
- Remove "Problems" section. Cover problems & limitations in "Analysis".

doc/design/persistent_volumes_etcd_data.md
```diff
@@ -0,0 +1,125 @@
...
+
+- Add to the cluster spec PodPolicy an option to enable putting etcd data inside a Persistent Volume
+- If podPolicy has persistent volumes enabled, when creating a new member also create a Persistent Volume Claim and use it as the volume source for etcd-data
+- Handle etcd data persistent volume claims cleanup:
+  - set on the PVC an owner like already done with other resources.
```
hongchaodeng: `Set PVC ownerRef to EtcdCluster object, which enables GC to clean up if cluster is deleted.`

Also remove the below " garbage collect also PVCs" line.

doc/design/persistent_volumes_etcd_data.md
```diff
@@ -0,0 +1,125 @@
...
+  - in the reconcile loop, remove PVCs that don't have a matching etcd member in the etcd cluster and their related pod doesn't exists in the k8s API
```
hongchaodeng: `PVC lifecycle is bound to etcd member's lifecycle. PVC is added/removed when etcd member is added/removed.`

hongchaodeng: > remove PVCs that don't have a matching etcd member in the etcd cluster and their related pod doesn't exists in the k8s API

 We don't remove PVCs just because pods were gone. PVC lifecycle is bound to member's lifecycle.

Currently if a pod is gone (evicted maybe?), we remove its member and create a new one with a new pod. With PV support, we will keep its member, then:

- tolerate it for a while (e.g. 5 mins), waiting it to restart. Or
- after toleration period ends or pod is gone from APIServer, create a pod with the same identity and using the same PVC.

In short, we don't need to change membership with PV support. To put it more general, PVC lifecycle is bound to member's lifecycle. For example, we will remove PVC only if its member is removed in scale-down.

hongchaodeng: Note that we don't handle failure & toleration in part 1. Just explain why it doesn't correlate to Pods.

doc/design/persistent_volumes_etcd_data.md
```diff
@@ -0,0 +1,125 @@
...
+
+This first part won't fix case C and D.
+
+### Part 2
+
+#### Changes
+
+- Change reconcile logic to handle deleted pods that had a persistent volume:
+
+If a pod of an etcd cluster member is deleted and it had a persistent volume claim associated, instead of removing the etcd member from the etcd cluster and adding a new member just recreate it (keeping its pod name) binding the existing persistent volume claim. The pod recreation will happen only if the pod doesn't exists anymore in the API because we'll keep the same pod name. This also enable a MANDATORY property: at most once pod existence logic to avoid having in the API at the same time two pods with associated the same PVC.
```
hongchaodeng: So this paragraph only assumes one failure scenario -- pod is deleted.

See comment https://github.com/coreos/etcd-operator/pull/1374/files#r136628955 .

Please write down the two failure scenarios in part 2.

doc/design/persistent_volumes_etcd_data.md
```diff
@@ -0,0 +1,125 @@
...
+
+#### Results
+
+With this change if a majority of pods are deleted the etcd-operator will recreate them keeping their previous data, so case C and D will be fixed.
+
+
+#### Problems
+
+- until the pod isn't removed from the api server it won't be recreated (for some more details see the [Pod Safety, Consistency Guarantees, and Storage Implications](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/pod-safety.md) k8s design doc)
```
hongchaodeng: > until the pod isn't removed from the api server

We could not and would not solve this problem here.

Also, the proposal is about some limitations we need to understand and future directions heading to. It shouldn't be the problems in here. Reference is enough. It should be more practical problems to be written down.

doc/design/persistent_volumes_etcd_data.md
```diff
@@ -0,0 +1,125 @@
...
+- if the etcd data gets corrupted etcd will exit and the pod will go in a failed state and then removed from the api server. The change introdudec in this part will make etcd-operator recreate the pod with the previous corrupted data in a loop.
```
hongchaodeng: Data corruption should be seen as volume failure, and thus should be covered in toleration.

doc/design/persistent_volumes_etcd_data.md
```diff
@@ -0,0 +1,125 @@
...
+- there can be other unpredictable cases where etcd will continuously fail to start like in the above point (fs full, os problems etc...).
```
hongchaodeng: This should be covered as pod failure.

hongchaodeng: We also need to cover node failure in part 2.

sgotti:
> @sgotti
> I remembered now: Node restart will lost the data in emptydir.
> That's why we want local PV.

As far as I know and that the doc states emptyDir is tied to pod life, so it's not a tmpfs. So, if you set the pod restart policy to always and you reboot a node and it comes back online before the pod being evicted by the node controller your pod will be restarted with the previous data.

hongchaodeng:
> As far as I know and that the doc states emptyDir is tied to pod life, so it's not a tmpfs. So, if you set the pod restart policy to always and you reboot a node and it comes back online before the pod being evicted by the node controller your pod will be restarted with the previous data.

This issue is k8s issue. Please discuss it in k8s upstream.
For etcd operator, we won't rely on persistence of emptyDir.

sgotti:
> This issue is k8s issue. Please discuss it in k8s upstream.
For etcd operator, we won't rely on persistence of emptyDir.

There's no issue. I was just replying to your comment.

xiang90:
@sgotti 

We used to rely on the persistency of emptyDir based on pod lifecycle, and assumed that node restart would not clear emptyDir. 

However, @hongchaodeng found that in our node restart tests, sometime emptyDir will be cleared. We did not really look into this issue.

I am not sure if it is still the case today. Probably worth digging into it if anyone has time.

doc/design/persistent_volumes_etcd_data.md
georgekuruvillak: We add a member to the etcd-cluster, a new memberId is generated. If we re-use the PVC, the memberId of the failed member and the new member binding to the PV will be differnent. Would this create further problems?

sgotti: it'll be the same member since it has the same data and same startup command, only a new pod since the previous pod has gone (deleted from the api server)

doc/design/persistent_volumes_etcd_data.md
sgotti: > We also need to cover node failure in part 2.

Are you talking about k8s nodes?

doc/design/persistent_volumes_etcd_data.md
sgotti: There are different kind of failures. I was talking about data corruption (missing or corrupted files).

What's a "volume failure"?

doc/design/persistent_volumes_etcd_data.md
sgotti: I should move this out the "problem" section or just remove it since it's just a note on how k8s works. I mentioned it because the below proposed enhancements avoids this case.

doc/design/persistent_volumes_etcd_data.md
sgotti: Sorry but I lost what other failure scenario you're talking about.

doc/design/persistent_volumes_etcd_data.md
sgotti: > PVC lifecycle is bound to etcd member's lifecycle. PVC is added/removed when etcd member is added/removed.

will add.

> We don't remove PVCs just because pods were gone. PVC lifecycle is bound to member's lifecycle.

I tough it was clear since I put an AND. the second part of the AND is just to avoid possible problems if we are going to remove a PVC before the pod it's assigned to is gone.

doc/design/persistent_volumes_etcd_data.md
georgekuruvillak: Oh. You wouldn't need to add a member as was the case in addOneMember() in cluster/reconcile.go
```go
resp, err := etcdcli.MemberAdd(ctx, []string{newMember.PeerURL()})
```

doc/design/persistent_volumes_etcd_data.md
hongchaodeng: OK. Sure.

Just to sync up

doc/design/persistent_volumes_etcd_data.md
hongchaodeng: Sure. Let me clarify what should be done in part 2 more.


### Problems

What problems we have left in part 1? Failures:

- pod failure: pod deleted (for some reason), pod SLA low (better recreate it somewhere else)
- volume failure (failing to bind): In the future, k8s will add a taint to PV and kubelet will evict the pod. Pod unready -> pod deleted.
- node failure (eviction): pod unready -> pod deleted.
- scheduling failure: pod pending

So you are mostly right that pod being deleted is the most possible failures. But there could be other situations that we want to delete the pod in operator.

### Solutions

For pod deleted, we could just recreate the pod.

However, we also need to have tolerations. For example, we should set tolerations on volume taint, node taint. And we should tolerate pod "unready" for whatever reason for maximum time.

After a second thought, I think we could either deal with both recreating the pod and tolerations in part 2, or separate tolerations into part 3, or just leave tolerations as TODO (we can keep on working on it later).

doc/design/persistent_volumes_etcd_data.md
hongchaodeng: Sorry my fault.. Should be taken as pod failure (pod SLA low) case: https://github.com/coreos/etcd-operator/pull/1374#discussion_r137345989 .
But it won't be "removed from APIServer". We needs to evict it from operator side.

doc/design/persistent_volumes_etcd_data.md
hongchaodeng: Yeah. I categorize them in https://github.com/coreos/etcd-operator/pull/1374#discussion_r137345989 .

doc/design/persistent_volumes_etcd_data.md
hongchaodeng: Just remove it.

To sync up: We can't have at-most-once guarantee of pods. But we can have at most once guarantee of etcd pod running because underlying PV (GCE PD, EBS for instance) guarantees only one binding. Probably this should be mentioned.

I'm not really sure about the "proposed enhancements". Isn't it the same as the solutions to part 2?

doc/design/persistent_volumes_etcd_data.md
hongchaodeng: @georgekuruvillak 

> When a member is added to the etcd-cluster, a new memberId is generated.

No new ID for this new change.

> If the PVC is re-used, the memberId of the failed member and the new member binding to the PV will be differnent. 

It would be the same identity, network, state with PV like EBS, GCE PD.

doc/design/persistent_volumes_etcd_data.md
sgotti: > We can't have at-most-once guarantee of pods

I lost you :stuck_out_tongue_closed_eyes:.
I think we can have and must have (to avoid some volumes kind like iscsi being mounted on more than one node/pod at a time) at-most-once guarantee of pods. This is what stateful sets do and it's an enforced behavior of all other custom sf pod controller by the linked design doc. 

Relying on underlying storage that can be mounted only on one node at a time is an automatic fix to avoid data corruptions but won't work for all kind of storage, that's one of the reasons why that proposal has been created.

Note that this is already implemented in the k8s core components. The implementation is quite easy on the controller side: just use the same pod name so you can recreate it only if it's gone from the API. The core is in the node controller and how a pod is being deleted (no default force deletion in kubectl): 
https://github.com/kubernetes/kubernetes/pull/35235
https://github.com/kubernetes/kubernetes/pull/35484

As a side note, that the reason why there's a need, on every env (cloud or baremetal) to manually fence partitioned nodes or the implementation of a fence-controller.

doc/design/persistent_volumes_etcd_data.md
sgotti: Thanks for the explanation, I agree with you. Perhaps a part 3 approach will be a good step but I fear that until it's implemented, the part 2 implementation won't gracefully handle some failures (i.e. corrupted data) that are today handled with the current ephemeral pods. What's your preference?

doc/design/persistent_volumes_etcd_data.md
sgotti: > I'm not really sure about the "proposed enhancements". Isn't it the same as the solutions to part 2?

The subtle difference is in the choices we can have when a member pod having data on a PV is gone:

1. if the cluster is quorate we have 2 choices:
 * recreate the pod of that member
 * delete and add a new member
2. if the cluster is not quorate my only choice is to recreate the pod of that member

Part 2 proposes to always recreate the pod of the member regardless if the cluster is quorate or not.
But this will fail forever in the case of corrupted data or other failures on starting the etcd member process.

So the enhancement says to recreate pods only if the cluster isn't quorate and when quorate use the remove/add member logic.

Perhaps considering these cases under the "tolerations" point of view is a better approach than this "fixed" logic

doc/design/persistent_volumes_etcd_data.md
sgotti: I agree

doc/design/persistent_volumes_etcd_data.md
hongchaodeng: No preference. All is good.

hongchaodeng:
Can you update the proposal? I think we agree on the same approach now. Let's get this move forward and merged

sgotti:
> Can you update the proposal? I think we agree on the same approach now. Let's get this move forward and merged

Updated. I may have lost some required changes so let me know. I tried to define which new failures scenarios needs to be considered after enabling persistent volume for etcd data and kept some of them in a "future enhancements" parts to see how to handle them (like using tolerations). I haven't added some of the scenarios you described (like pod stuck in pending case) since it looks like something independent from the changes added by the design and that could happen also today.

hongchaodeng:
lgtm

doc/design/persistent_volumes_etcd_data.md
```diff
@@ -0,0 +1,112 @@
...
+
+This parts will fix cases C and D and the above corrupted data case.
+
+It's a change to the reconcile logic to handle deleted pods from the k8s API server that had a persistent volume. To handle corrupted data failures, the reconcile logic change defined below will be used only if the cluster is unquorate. Instead, if the cluster is quorate, the current logic of deleting the member and adding a new member will be used.
+
+- If the cluster is unquorate and a pod of an etcd cluster member is deleted from the API server and it had a persistent volume claim associated, instead of removing the etcd member from the etcd cluster and adding a new member just recreate it (keeping its pod name) binding the existing persistent volume claim. The pod recreation will happen only if the pod doesn't exists anymore in the API because we'll keep the same pod name. This also enable a MANDATORY property: at most once pod existence logic to avoid having in the API at the same time two pods with associated the same PVC.
```
georgekuruvillak: If the cluster is unquorate and the volume has been corrupted, wont the etcd pod go into an infinite restart cycle as it re-attaches to the same corrupted PV? Is there any corruption check in place that will delete the PV? Should not the cluster be re-created from the backup?

irfn:
@sgotti 
Since there is no further activity on this thread i wanted some clarifications
- Would implementing this proposal just involve moving the etcd pods from deployment to stateful sets?
- Are there any situations where statefulsets would not recover from scenarios Case C and Case D barring a disk failure.
My understanding of stateful sets is the current one maybe this has to do with shortcomings of previous implementations 
/cc @hongchaodeng @georgekuruvillak

PR #1394, opened 2017-09-12
doc: add a note on TLS to the RBAC docs

brunomcustodio:
In order to create an `etcd` cluster with TLS enabled, the service account being used needs access to the `etcd-client-tls`, `etcd-peer-tls` and `etcd-server-tls` secrets.

---

Since one needs `secrets` for both TLS and S3 I don't know whether it would be preferable to include `secrets` right away here:

https://github.com/coreos/etcd-operator/blame/master/doc/user/rbac.md#L81

Any thoughts?

### DIFF
--- a/doc/user/rbac.md
+++ b/doc/user/rbac.md
@@ -95,7 +95,19 @@ rules:
 EOF
 ```
 
-If you need use s3 backup, add these to above input:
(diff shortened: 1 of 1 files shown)


hongchaodeng:
This looks good enough.
Thanks @brunomcustodio

PR #1395, opened 2017-09-12
*: document resource labels

hasbro17:
@hongchaodeng I've listed out all the resource labels I could see but PTAL and let me know if we need to align them better.

Only the backup sidecar pod uses the `app=etcd_backup_tool` label which I think was required to select it in e2e testing.


hongchaodeng:
lgtm

PR #1400, opened 2017-09-13
Fix DNS discovery.

### LINKED ISSUE #1381: Services can't be used for DNS discovery
I'm using `quay.io/coreos/etcd-operator:v0.5.1` to create the following cluster:

```yaml
kind: EtcdCluster
metadata:
  name: etcd
spec:
  size: 3
  version: "3.2.7"
  pod:
    antiAffinity: true
```

I see two services created with ports named `client` and `peer`:

```
$ kubectl describe services
Name:			etcd
Namespace:		default
Labels:			app=etcd
			etcd_cluster=etcd
Annotations:		service.alpha.kubernetes.io/tolerate-unready-endpoints=true
Selector:		app=etcd,etcd_cluster=etcd
Type:			ClusterIP
IP:			None
Port:			client	2379/TCP
Endpoints:		10.8.0.15:2379,10.8.1.9:2379,10.8.2.10:2379
Port:			peer	2380/TCP
Endpoints:		10.8.0.15:2380,10.8.1.9:2380,10.8.2.10:2380
Session Affinity:	None
Events:			<none>


Name:			etcd-client
Namespace:		default
Labels:			app=etcd
			etcd_cluster=etcd
Annotations:		service.alpha.kubernetes.io/tolerate-unready-endpoints=true
Selector:		app=etcd,etcd_cluster=etcd
Type:			ClusterIP
IP:			10.11.253.157
Port:			client	2379/TCP
Endpoints:		10.8.0.15:2379,10.8.1.9:2379,10.8.2.10:2379
Session Affinity:	None
Events:			<none>
```

I can use the `etcd-client` service as the endpoint and see that the cluster is functional:

```
# etcdctl -v
etcdctl version 2.2.5

# ETCDCTL_ENDPOINT="http://etcd-client:2379" etcdctl cluster-health
member 186504f50442094c is healthy: got healthy result from http://etcd-0001.etcd.default.svc:2379
member 6c4d5a576dad55c5 is healthy: got healthy result from http://etcd-0002.etcd.default.svc:2379
member b885aaad8c46a728 is healthy: got healthy result from http://etcd-0000.etcd.default.svc:2379
```

However, I'm unable to use SRV discovery to connect to those services:

```
# for n in etcd etcd.default etcd.default.svc.cluster.local etcd-client etcd-client.default etcd-client.default.svc.cluster.local ; do ETCDCTL_DISCOVERY_SRV="$n" etcdctl cluster-health ; done
dns lookup errors: lookup _etcd-server-ssl._tcp.etcd on 10.11.240.10:53: no such host and lookup _etcd-server._tcp.etcd on 10.11.240.10:53: no such host
dns lookup errors: lookup _etcd-server-ssl._tcp.etcd.default on 10.11.240.10:53: no such host and lookup _etcd-server._tcp.etcd.default on 10.11.240.10:53: no such host
dns lookup errors: lookup _etcd-server-ssl._tcp.etcd.default.svc.cluster.local on 10.11.240.10:53: no such host and lookup _etcd-server._tcp.etcd.default.svc.cluster.local on 10.11.240.10:53: no such host
dns lookup errors: lookup _etcd-server-ssl._tcp.etcd-client on 10.11.240.10:53: no such host and lookup _etcd-server._tcp.etcd-client on 10.11.240.10:53: no such host
dns lookup errors: lookup _etcd-server-ssl._tcp.etcd-client.default on 10.11.240.10:53: no such host and lookup _etcd-server._tcp.etcd-client.default on 10.11.240.10:53: no such host
dns lookup errors: lookup _etcd-server-ssl._tcp.etcd-client.default.svc.cluster.local on 10.11.240.10:53: no such host and lookup _etcd-server._tcp.etcd-client.default.svc.cluster.local on 10.11.240.10:53: no such host
```

Is that intentional?

If I understand correctly, naming the ports `etcd-server` and `etcd-client` will generate the correct SRV records.

- https://coreos.com/etcd/docs/3.2.7/op-guide/clustering.html#dns-discovery
- https://github.com/kubernetes/dns/blob/1.14.4/docs/specification.md#232---srv-records

brunomcustodio:
Fixes #1381.

### DIFF
--- a/pkg/util/k8sutil/k8sutil.go
+++ b/pkg/util/k8sutil/k8sutil.go
@@ -117,9 +117,9 @@ func PodWithNodeSelector(p *v1.Pod, ns map[string]string) *v1.Pod {
 	return p
 }
 
-func CreateClientService(kubecli kubernetes.Interface, clusterName, ns string, owner metav1.OwnerReference) error {
+func CreateClientService(kubecli kubernetes.Interface, clusterName, ns string, clientPortName string, owner metav1.OwnerReference) error {
 	ports := []v1.ServicePort{{
-		Name:       "client",
+		Name:       clientPortName,
 		Port:       2379,
 		TargetPort: intstr.FromInt(2379),
 		Protocol:   v1.ProtocolTCP,
@@ -131,14 +131,14 @@ func ClientServiceName(clusterName string) string {
 	return clusterName + "-client"
 }
 
-func CreatePeerService(kubecli kubernetes.Interface, clusterName, ns string, owner metav1.OwnerReference) error {
+func CreatePeerService(kubecli kubernetes.Interface, clusterName, ns string, clientPortName, serverPortName string, owner metav1.OwnerReference) error {
 	ports := []v1.ServicePort{{
-		Name:       "client",
+		Name:       clientPortName,
 		Port:       2379,
 		TargetPort: intstr.FromInt(2379),
 		Protocol:   v1.ProtocolTCP,
 	}, {
-		Name:       "peer",
+		Name:       serverPortName,
 		Port:       2380,
 		TargetPort: intstr.FromInt(2380),
 		Protocol:   v1.ProtocolTCP,
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -52,6 +52,13 @@ const (
 	eventModifyCluster clusterEventType = "Modify"
 )
 
+const (
+	clientPortName    = "etcd-client"
+	clientPortTLSName = "etcd-client-ssl"
+	serverPortName    = "etcd-server"
+	serverPortTLSName = "etcd-server-ssl"
+)
+
 type clusterEvent struct {
 	typ     clusterEventType
 	cluster *spec.EtcdCluster
@@ -447,12 +454,26 @@ func (c *Cluster) delete() {
 }
 
 func (c *Cluster) setupServices() error {
-	err := k8sutil.CreateClientService(c.config.KubeCli, c.cluster.Name, c.cluster.Namespace, c.cluster.AsOwner())
+	var cpn, spn string
+
+	if !c.isSecureClient() {
+		cpn = clientPortName
+	} else {
+		cpn = clientPortTLSName
+	}
+
+	if !c.isSecurePeer() {
+		spn = serverPortName
+	} else {
+		spn = serverPortTLSName
+	}
+
+	err := k8sutil.CreateClientService(c.config.KubeCli, c.cluster.Name, c.cluster.Namespace, cpn, c.cluster.AsOwner())
 	if err != nil {
 		return err
 	}
 
-	return k8sutil.CreatePeerService(c.config.KubeCli, c.cluster.Name, c.cluster.Namespace, c.cluster.AsOwner())
+	return k8sutil.CreatePeerService(c.config.KubeCli, c.cluster.Name, c.cluster.Namespace, cpn, spn, c.cluster.AsOwner())
 }
 
 func (c *Cluster) createPod(members etcdutil.MemberSet, m *etcdutil.Member, state string, needRecovery bool) error {


pkg/util/k8sutil/k8sutil.go
```diff
@@ -117,9 +121,17 @@ func PodWithNodeSelector(p *v1.Pod, ns map[string]string) *v1.Pod {
 	return p
 }
 
-func CreateClientService(kubecli kubernetes.Interface, clusterName, ns string, owner metav1.OwnerReference) error {
+func CreateClientService(kubecli kubernetes.Interface, clusterName, ns string, isSecureClient bool, owner metav1.OwnerReference) error {
```
hongchaodeng: No need to change this func?

pkg/util/k8sutil/k8sutil.go
```diff
@@ -131,14 +143,28 @@ func ClientServiceName(clusterName string) string {
 	return clusterName + "-client"
 }
 
-func CreatePeerService(kubecli kubernetes.Interface, clusterName, ns string, owner metav1.OwnerReference) error {
+func CreatePeerService(kubecli kubernetes.Interface, clusterName, ns string, isSecureClient, isSecurePeer bool, owner metav1.OwnerReference) error {
```
hongchaodeng: Can you change this func to accept `clientPortName`, `serverPortName` instead?
This util shouldn't understand etcd logic.

pkg/util/k8sutil/k8sutil.go
brunomcustodio: I know, but I chose to change it to make things consistent.

pkg/util/k8sutil/k8sutil.go
brunomcustodio: Done, although if these utilities shouldn't understand etcd logic at all I'd suggest removing the hardcoded port numbers (on a separate PR, perhaps).

brunomcustodio:
@hongchaodeng I wouldn't submit a PR if I hadn't confirmed it works 🙂 I've addressed your points and squashed.

hongchaodeng:
@brunomcustodio 

>  I wouldn't submit a PR if I hadn't confirmed it works 

Great. Thanks!

Can you separate the test changes into a separate PR? I would like to merge the fix first.

For the test, we could coordinate ways to test it and do actions like setting up test infra. That need more time. Let's handle it later.

brunomcustodio:
@hongchaodeng done! I will issue the PR with the test when this lands on `master` (I think it makes more sense that way).

xiang90:
Note that this PR breaks compatibility. Users who depends on the old DNS name will be affected.

/cc @hongchaodeng @brunomcustodio

brunomcustodio:
@xiang90 that's correct, but I assumed that was desired after the discussion in #1381. Do you want to keep compatibility?

xiang90:
@brunomcustodio It is always not desired to break compatibility. Why not use native k8s discovery by putting a k8s service in front of etcd cluster but use DNS discovery at the first place?

xiang90:
i am not saying it is impossible to break it as etcd operator has not reached stable. but i want to understand the use case before breaking things.

brunomcustodio:
> Why not use native k8s discovery by putting a k8s service in front of etcd cluster but use DNS discovery at the first place?

@xiang90 I don't think I understood your suggestion. Could you please elaborate?

brunomcustodio:
@xiang90 the use case is to make [this](https://coreos.com/etcd/docs/latest/op-guide/clustering.html#dns-discovery) work.

xiang90:
> @xiang90 I don't think I understood your suggestion. Could you please elaborate?

k8s has its own way to do service discovery by utilizing label selectors: https://kubernetes.io/docs/concepts/services-networking/service/. the client, then, can use be pointed to a single stable endpoint. the k8s service will handle the actual discovery.

the process is the same when you provide the client a SRV record.

xiang90:
@brunomcustodio I want to know why you want to make DNS discovery work when you already have k8s native discovery.

brunomcustodio:
> k8s has its own way to do service discovery by utilizing label selectors

@xiang90 I certainly do know that and how discovery works in Kubernetes. 🙂

> I want to know why you want to make DNS discovery work when you already have k8s native discovery.

@xiang90 there are application (such as [Vault](https://www.vaultproject.io/)) which still provide the option to use a `discovery_srv` domain in configuration. Maybe some people would like to use that?

xiang90:
@brunomcustodio 

Even for Vault, it is easer to just provide one etcd endpoint (k8s service) (TLS is easier, sync is not necessary anymore, dynamic membership is also handled by k8s discovery layer).

I do not want to promote using DNS discovery when deploy etcd atop Kubernetes unless there is a concrete use case.

xiang90:
@brunomcustodio 

I am OK with changing the DNS name though. It matches with etcd conventions better. But, still, I do not want to promote DNS discovery for Kubernetes users.

/cc @hongchaodeng @fanminshi 

If we decide to merge this in, we need to add this as a breaking change in the change log.

hongchaodeng:
@etcd-bot ok to test

hongchaodeng:
@brunomcustodio @xiang90 

Why not create a new service called `<cluster-name>-discovery`?
The Peer service was used by peer communication (incl. client) for subdomain purpose.
It is not a wise option to couple these two purposes.

hongchaodeng:
We don't need to do peer URLs discovery too, right?

xiang90:
> Why not create a new service called <cluster-name>-discovery?

That is why I said above. There is no need to use DNS discovery when you already have an infrastructure that supports native discovery.

> The Peer service was used by peer communication (incl. client) for subdomain purpose.
It is not a wise option to couple these two purposes.

The only benefits I can see from this PR is to make peer/client SRV record more consistent with etcd convention.

brunomcustodio:
> It matches with etcd conventions better.

> The only benefits I can see from this PR is to make peer/client SRV record more consistent with etcd convention.

This was the single purpose of the PR—to follow the convention. I myself do not need this to be merged—just saw this being asked for in #1381 and decided to contribute since @cbandy didn't have the time.

hongchaodeng:
https://github.com/coreos/etcd-operator/pull/1400#issuecomment-329917475 is valid concern.

We can create a new service. Or maybe we could provide a new template.

Note that there is already a docs for pods' labels:
https://github.com/coreos/etcd-operator/blob/master/doc/user/resource_labels.md

cbandy:
@brunomcustodio, thank you for opening this PR _and_ writing a test!

I opened #1381 because I was surprised when the Kubernetes operator within the GitHub CoreOS organization generated SRV records that could not be used by `etcdctl`, a CoreOS tool.

I've been also surprised (or maybe confused) by the responses from the contributors to this repository. I am a novice regarding etcd deployments, but I'd like to understand more.

> Note that this PR breaks compatibility. Users who depends on the old DNS name will be affected.

True. Someone using the `_client._tcp` subdomain would be affected. The remainder, `{cluster-name}.{namespace}.svc.cluster.local`, is _not_ affected.


> k8s has its own way to do service discovery by utilizing label selectors: https://kubernetes.io/docs/concepts/services-networking/service/. the client, then, can use be pointed to a single stable endpoint. the k8s service will handle the actual discovery.

Agreed. The purpose of a Kubernetes service is to provide a stable endpoint that maps to one or more pods. That stable endpoint is one or more of the following:

- Cluster IP. `kube-proxy` directs traffic destined for this IP to one or more pods instead.
- DNS name. `kube-dns` resolves queries for this name (and named ports) to the cluster IP or to one or more pods.


> I do not want to promote using DNS discovery when deploy etcd atop Kubernetes unless there is a concrete use case.

> There is no need to use DNS discovery when you already have an infrastructure that supports native discovery.

Can you please elaborate? The service created by `CreatePeerService` is headless, so its _only effect_ is DNS records. (Headless services do not even appear in service [environment variables](https://kubernetes.io/docs/concepts/services-networking/service/#environment-variables).)

Perhaps we have multiple/different definitions of "DNS discovery"?


> The Peer service was used by peer communication (incl. client) for subdomain purpose. We don't need to do peer URLs discovery either.

Can you please elaborate? What's the difference between peer communication and peer URLs discovery?

> It is not a wise option to couple these two purposes.

Which two purposes? Client versus peer?

hongchaodeng:
@cbandy 

> I was surprised when the Kubernetes operator within the GitHub CoreOS organization generated SRV records that could not be used by etcdctl, a CoreOS tool.
> 
> I've been also surprised (or maybe confused) by the responses from the contributors to this repository.

This is free and open source project. Even if we do our best and spend spare time to maintain it, we can't guarantee everything works as expected. If you encounter any problem, just describe it as detailed and clearly as possible. For a solution, just be reasonable because we need to ensure it goes the right direction. Even better,  write docs, or submit a PR like @brunomcustodio did.

So back to [this comment](https://github.com/coreos/etcd-operator/pull/1400#issuecomment-329917475), the purpose of this service is not for discovery even if we make it headless. It's better to not couple another purpose.

Since we also have [docs](https://github.com/coreos/etcd-operator/blob/master/doc/user/resource_labels.md) around labels for etcd pods, it seems a better option to create just another service for discovery purpose. That's also an extensible option, that said, if etcd discovery name changes or k8s service SRV record changes, it is much easier to change.

It seems what we need is docs.

xiang90:
@cbandy 

> Perhaps we have multiple/different definitions of "DNS discovery"?

I am curious why do you want to use DNS discovery when you already run etcd in Kubernetes.

What I do normally is to add a service that select all my etcd pods, say it is `service-etcd`. Then I will point my etcdctl to `service-etcd:2379`. 

Again, I am fine with the change to make it consistent with etcd doc. However, I would like to understand the use case if there is any.

brunomcustodio:
> @brunomcustodio, thank you for opening this PR _and_ writing a test!
>
> I opened #1381 because I was surprised when the Kubernetes operator within the GitHub CoreOS organization generated SRV records that could not be used by etcdctl, a CoreOS tool.

@cbandy you're most welcome 🙂 I, too, was surprised with this...

brunomcustodio:
@cbandy @hongchaodeng @xiang90 let's all share a cup of tea and try to reach a final decision that is reasonable. ☕️

> I opened #1381 because I was surprised when the Kubernetes operator within the GitHub CoreOS organization generated SRV records that could not be used by etcdctl, a CoreOS tool.

I believe @cbandy's concern to be a valid concern. This is confusing, especially for newcomers, and mitigating it was the sole purpose of this PR in the first place. That being said, I believe we only have three options here:

* Leave everything as is, close this and #1381 _but_ somehow mention that clusters created by `etcd-operator` _intentionally_ do not support [this](https://coreos.com/etcd/docs/latest/op-guide/clustering.html#dns-discovery).
* Merge the PR as it is, breaking compatibility—something that is being made for other PRs.
* Amend this PR so that a third service is created for every cluster—although I think people will find it confusing due to its similarity with the existing one.

@hongchaodeng @xiang90 as maintainers, do you agree with any of the above solutions? I am, of course, available to amend the PR as you see fit.

hongchaodeng:
I think it's better to create a new service called "<cluster-name>-discovery" for discovery purpose.

It's not only to keep compatibility, but also we shouldn't couple this discovery service with existing peer service -- they are for different purposes. Coupling it will incur maintaining burden and confusion.

Additionally, I don't even think the discovery service should be created by default. It should be a template and docs first, or optional feature. In order to make a decision, we will need concrete use cases.

brunomcustodio:
> I think it's better to create a new service called "-discovery" for discovery purpose.

> Additionally, I don't even think the discovery service should be created by default. It should be a template and docs first, or optional feature.

This discussion got too confusing really soon. I see

> (...) we reserve the right to address bugs and API changes in a backwards incompatible way before the project is declared stable. 

stated as a big bold warning in `README.md`, and I see a PR that fixes a bug taking 30+ philosophical messages to get approved or rejected (and yes, it's OK to reject a PR).

I'm leaving the thread, partly due to lack of time to follow-up on this and partly due to the fact that I lost interest on this due to the "bureaucracy" involved. If you guys ever need something from me regarding this (e.g., the test I was asked to remove) just drop me a line.

hongchaodeng:
It's my fault.
I shouldn't ask a PR before we have agreed on a solution.
Let's move back the discussion to https://github.com/coreos/etcd-operator/issues/1381 .

PR #1404, opened 2017-09-14
doc: clarify s3 backup with restore spec

fanminshi:
[skip ci]

### DIFF
--- a/doc/user/spec_examples.md
+++ b/doc/user/spec_examples.md
@@ -90,7 +90,8 @@ spec:
 
 ### Three members cluster that restores from previous S3 backup
 
-Same as above but using "S3" as backup storage.
+Same as above but using "S3" as backup storage. The previous s3 backup must exist 
+in order for the restore to work; otherwise, the new cluster doesn't boot.
 
 ```yaml
 metadata:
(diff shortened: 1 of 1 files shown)


doc/user/spec_examples.md
```diff
@@ -101,11 +102,52 @@ spec:
...
+      awsSecret: <aws-secret-name>
+      prefix: <S3-prefix>
   restore:
     backupClusterName: "cluster-a"
     storageType: "S3"
 ```
 
+### Three members cluster that restores from different cluster's S3 backup
+
+etcd-operator can create a new cluster `cluster-b` from the backup of another cluster `cluster-a`. If backup doesn't exist, `cluster-b` starts as a new cluster. If backup exist, ensure that the spec of `cluster-b` has the same `s3Bucket`, `awsSecret`, and `prefix` as those on spec of `cluster-a`; so that etcd operator can find the correct s3 backup path from `cluster-a`.
```
hongchaodeng: > If backup doesn't exist, `cluster-b` starts as a new cluster.

Well, this is not true.. Because if no backup exists, why does a user want to restore.. Why not just create a new cluster.

fanminshi: When I restore from a cluster where the backup doesn't exist, I notice the new cluster just start as usual. I want to record this behavior so that the user knows what will happen. In contrast if I restore the same cluster and the backup doesn't exist, the cluster won't start because it complains the backup doesn't exist. So I think documenting those behaviors might know wha's expected.

hongchaodeng: > When I restore from a cluster where the backup doesn't exist, I notice the new cluster just start as usual.

The behavior is undefined so far. Instead of dealing with it and coupling the issue here, we should open a new issue/PR to define this scenario.

fanminshi: kk, I'll open another issue discussing on this.

hongchaodeng: Can you remove the the line about this special case? Then we can move forward this PR.

fanminshi: will do that!

fanminshi:
all fixed. PTAL

doc/user/spec_examples.md
```diff
@@ -101,11 +102,52 @@ spec:
...
+  size: 3
+  backup:
+    backupIntervalInSecond: 300
+    maxBackups: 5
+    storageType: "S3"
+    s3:
+      s3Bucket: <S3-bucket-name>
+      awsSecret: <aws-secret-name>
+      prefix: <S3-prefix>
+    restore:
```
hongchaodeng: indent is wrong

doc/user/spec_examples.md
```diff
@@ -101,11 +102,52 @@ spec:
...
+      awsSecret: <aws-secret-name>
+      prefix: <S3-prefix>
   restore:
     backupClusterName: "cluster-a"
     storageType: "S3"
 ```
 
+### Three members cluster that restores from different cluster's S3 backup
+
+etcd-operator can create a new cluster `cluster-b` from the backup of another cluster `cluster-a`. If backup exists, make sure the spec of `cluster-b` has the same `s3Bucket`, `awsSecret`, and `prefix` as those on the spec of `cluster-a`; so that the etcd operator can find the correct s3 backup path.
```
hongchaodeng: make sure the backup policy of `cluster-b` has the same S3 configuration as those of `cluster-a`; etcd operator would also use that config to access snapshots of `cluster-a`.

fanminshi:
fixed!

hongchaodeng:
lgtm

Issue #1412, opened 2017-09-18
Add events into EtcdCluster Resource


hongchaodeng:

hasbro17:
For a start we can create the following events common to the EtcdCluster lifecycle:
- Create:
  - A new member M is added
- Update
  - A member M is upgraded from v1 to v2
- Delete
  - A member M is removed
  - A dead member M is removed

Issue #1388, opened 2017-09-09
Refine status conditions


hongchaodeng:
EtcdCluster status conditions was originally hacked up for UI. It wasn't quite aligned with upstream yet, e.g. the rotation.. We need to refine it.

hongchaodeng:
The current "Conditions" are in fact events.. `kubectl describe` wasn't working for TPR before.
Now it seems to work with custom resource.

First of all, we need to change them to events. Second, we need to test k-describe.

hongchaodeng:
fixed

PR #1419, opened 2017-09-21
doc: events for EtcdCluster proposal

hasbro17:
[skip ci]
Define Events #1412
Refine conditions #1388


doc/design/conditions_and_events.md
```diff
@@ -0,0 +1,20 @@
+# Status Conditions and Events
```
xiang90: this is a user facing feature. probably worth moving to user doc?

hongchaodeng:
lgtm

PR #1433, opened 2017-09-22
*: use consts for operator pod name and namespace

hasbro17:
Add constants for using `MY_POD_NAME` and `MY_POD_NAMESPACE` in different places.


hongchaodeng:
lgtm

PR #1434, opened 2017-09-22
*: implement PersistentVolume for etcd data design part 1

sgotti:
this patch implements part 1 of the etcd data on persistent volumes design.

when pod pvsource is defined in the spec it'll create a PVC for every etcd
member and use it as the volume for etcd data.

pvc without a member will be removed during the reconcile.

**NOTES**

* As discussed during the design phase I haven't changed to pod restart policy to Always. So beside putting data on a PV this patch shoudln't change any other logic and behavior.

### DIFF
--- a/pkg/apis/etcd/v1beta2/cluster.go
+++ b/pkg/apis/etcd/v1beta2/cluster.go
@@ -25,6 +25,8 @@ import (
 const (
 	defaultBaseImage = "quay.io/coreos/etcd"
 	defaultVersion   = "3.1.8"
+
+	minPodPVSizeInMB = 512 // 512MiB
 )
 
 var (
@@ -170,7 +172,6 @@ type PodPolicy struct {
 
 	// PV represents a Persistent Volume resource.
 	// If defined new pods will use a persistent volume to store etcd data.
-	// TODO(sgotti) unimplemented
 	PV *PVSource `json:"pv,omitempty"`
 
 	// By default, kubernetes will mount a service account token into the etcd pods.
@@ -202,6 +203,11 @@ func (c *ClusterSpec) Validate() error {
 		for k := range c.Pod.Labels {
 			if k == "app" || strings.HasPrefix(k, "etcd_") {
 				return errors.New("spec: pod labels contains reserved label")
+			}
+		}
+		if c.Pod.PV != nil {
+			if c.Pod.PV.VolumeSizeInMB < minPodPVSizeInMB {
+				c.Pod.PV.VolumeSizeInMB = minPodPVSizeInMB
 			}
 		}
 	}
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -298,6 +298,13 @@ func (c *Cluster) run() {
 				continue
 			}
 
+			pvcs, err := c.pollPVCs()
+			if err != nil {
+				c.logger.Errorf("failed to poll pvcs: %v", err)
+				reconcileFailed.WithLabelValues("failed to poll vcs").Inc()
+				continue
+			}
+
 			if len(pending) > 0 {
 				// Pod startup might take long, e.g. pulling image. It would deterministically become running or succeeded/failed later.
 				c.logger.Infof("skip reconciliation: running (%v), pending (%v)", k8sutil.GetPodNames(running), k8sutil.GetPodNames(pending))
@@ -322,7 +329,7 @@ func (c *Cluster) run() {
 					break
 				}
 			}
-			rerr = c.reconcile(running)
+			rerr = c.reconcile(running, pvcs)
 			if rerr != nil {
 				c.logger.Errorf("failed to reconcile: %v", rerr)
 				break
@@ -415,6 +422,11 @@ func (c *Cluster) startSeedMember(recoverFromBackup bool) error {
 		SecureClient: c.isSecureClient(),
 	}
 	ms := etcdutil.NewMemberSet(m)
+	if c.IsPodPVEnabled() {
+		if err := c.createPVC(m); err != nil {
+			return fmt.Errorf("failed to create persistent volume claim for seed member (%s): %v", m.Name, err)
+		}
+	}
 	if err := c.createPod(ms, m, "new", recoverFromBackup); err != nil {
 		return fmt.Errorf("failed to create seed member (%s): %v", m.Name, err)
 	}
@@ -435,6 +447,10 @@ func (c *Cluster) isSecurePeer() bool {
 
 func (c *Cluster) isSecureClient() bool {
 	return c.cluster.Spec.TLS.IsSecureClient()
+}
+
+func (c *Cluster) IsPodPVEnabled() bool {
+	return c.cluster.Spec.Pod != nil && c.cluster.Spec.Pod.PV != nil
 }
 
 // bootstrap creates the seed etcd member for a new cluster.
@@ -475,6 +491,12 @@ func (c *Cluster) setupServices() error {
 	return k8sutil.CreatePeerService(c.config.KubeCli, c.cluster.Name, c.cluster.Namespace, c.cluster.AsOwner())
 }
 
+func (c *Cluster) createPVC(m *etcdutil.Member) error {
+	pvc := k8sutil.NewPVC(m, c.cluster.Spec, c.cluster.Name, c.cluster.Namespace, c.cluster.AsOwner())
+	_, err := c.config.KubeCli.Core().PersistentVolumeClaims(c.cluster.Namespace).Create(pvc)
+	return err
+}
+
 func (c *Cluster) createPod(members etcdutil.MemberSet, m *etcdutil.Member, state string, needRecovery bool) error {
 	var pod *v1.Pod
 	if state == "new" {
@@ -487,6 +509,7 @@ func (c *Cluster) createPod(members etcdutil.MemberSet, m *etcdutil.Member, stat
 	} else {
 		pod = k8sutil.NewEtcdPod(m, members.PeerURLPairs(), c.cluster.Name, state, "", c.cluster.Spec, c.cluster.AsOwner())
 	}
+	k8sutil.AddEtcdVolumeToPod(pod, m, c.IsPodPVEnabled())
 	_, err := c.config.KubeCli.Core().Pods(c.cluster.Namespace).Create(pod)
 	return err
 }
@@ -505,6 +528,21 @@ func (c *Cluster) removePod(name string) error {
 	}
 	if c.isDebugLoggerEnabled() {
 		c.debugLogger.LogPodDeletion(name)
+	}
+
+	return nil
+}
+
+func (c *Cluster) removePVC(name string) error {
+	ns := c.cluster.Namespace
+	err := c.config.KubeCli.Core().PersistentVolumeClaims(ns).Delete(name, nil)
+	if err != nil {
+		if !k8sutil.IsKubernetesResourceNotFoundError(err) {
+			return err
+		}
+		if c.isDebugLoggerEnabled() {
+			c.debugLogger.LogMessage(fmt.Sprintf("pvc (%s) not found while trying to delete it", name))
+		}
 	}
 	return nil
 }
@@ -535,6 +573,29 @@ func (c *Cluster) pollPods() (running, pending []*v1.Pod, err error) {
 	}
 
 	return running, pending, nil
+}
+
+func (c *Cluster) pollPVCs() (pvcs []*v1.PersistentVolumeClaim, err error) {
+	pvcList, err := c.config.KubeCli.Core().PersistentVolumeClaims(c.cluster.Namespace).List(k8sutil.ClusterListOpt(c.cluster.Name))
+	if err != nil {
+		return nil, fmt.Errorf("failed to list running pvcs: %v", err)
+	}
+
+	for i := range pvcList.Items {
+		pvc := &pvcList.Items[i]
+		if len(pvc.OwnerReferences) < 1 {
+			c.logger.Warningf("pollPVCs: ignore pvc %v: no owner", pvc.Name)
+			continue
+		}
+		if pvc.OwnerReferences[0].UID != c.cluster.UID {
+			c.logger.Warningf("pollPVCs: ignore pvc %v: owner (%v) is not %v",
+				pvc.Name, pvc.OwnerReferences[0].UID, c.cluster.UID)
+			continue
+		}
+		pvcs = append(pvcs, pvc)
+	}
+
+	return pvcs, nil
 }
 
 func (c *Cluster) updateMemberStatus(members etcdutil.MemberSet) {
--- a/pkg/cluster/reconcile.go
+++ b/pkg/cluster/reconcile.go
@@ -32,7 +32,7 @@ import (
 // reconcile reconciles cluster current state to desired state specified by spec.
 // - it tries to reconcile the cluster to desired size.
 // - if the cluster needs for upgrade, it tries to upgrade old member one by one.
-func (c *Cluster) reconcile(pods []*v1.Pod) error {
+func (c *Cluster) reconcile(pods []*v1.Pod, pvcs []*v1.PersistentVolumeClaim) error {
 	c.logger.Infoln("Start reconciling")
 	defer c.logger.Infoln("Finish reconciling")
 
@@ -46,6 +46,10 @@ func (c *Cluster) reconcile(pods []*v1.Pod) error {
 		return c.reconcileMembers(running)
 	}
 	c.status.ClearCondition(api.ClusterConditionScaling)
+
+	if err := c.reconcilePVCs(pvcs); err != nil {
+		return err
+	}
 
 	if needUpgrade(pods, sp) {
 		c.status.UpgradeVersionTo(sp.Version)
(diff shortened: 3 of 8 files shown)


hack/test
```diff
@@ -87,7 +87,10 @@ function e2e_pass {
 
 	# Run all the tests by default
 	E2E_TEST_SELECTOR=${E2E_TEST_SELECTOR:-.*}
+	# Run tests with PV support disabled
 	go test "./test/e2e/" -run "$E2E_TEST_SELECTOR" -timeout 30m --race --kubeconfig $KUBECONFIG --operator-image $OPERATOR_IMAGE --namespace ${TEST_NAMESPACE}
+	# Run tests with PV support enabled
+	PV_TEST=true go test "./test/e2e/" -run "$E2E_TEST_SELECTOR" -timeout 30m --race --kubeconfig $KUBECONFIG --operator-image $OPERATOR_IMAGE --namespace ${TEST_NAMESPACE}
```
sgotti: To see how the tests works on your jenkins I just rerun them enabling pv support. Do you have better ideas on how to do this without duplicating the code?

test/e2e/basic_test.go
```diff
@@ -40,7 +40,7 @@ func TestCreateCluster(t *testing.T) {
 		}
 	}()
 
-	if _, err := e2eutil.WaitUntilSizeReached(t, f.CRClient, 3, 6, testEtcd); err != nil {
+	if _, err := e2eutil.WaitUntilSizeReached(t, f.CRClient, 3, 20, testEtcd); err != nil {
```
sgotti: Since creating and mounting PVs takes some time I just blindly increasing the retries. Better ideas?

pkg/cluster/cluster.go
```diff
@@ -464,13 +480,41 @@ func (c *Cluster) setupServices() error {
...
+func (c *Cluster) createPVC(m *etcdutil.Member) error {
+	pvc := k8sutil.NewPVC(m, c.cluster.Spec, c.cluster.Name, c.cluster.Namespace, c.cluster.AsOwner())
+	_, err := c.config.KubeCli.Core().PersistentVolumeClaims(c.cluster.Namespace).Create(pvc)
+	if err != nil {
+		return err
+	}
+
+	// TODO: We set timeout to 60s here since PVC binding could take up to 60s for GCE/PD. See https://github.com/kubernetes/kubernetes/issues/40972 .
+	//       Change the wait time once there are official p99 SLA.
+	err = retryutil.Retry(4*time.Second, 15, func() (bool, error) {
```
sgotti: This part is copied from https://github.com/coreos/etcd-operator/blob/master/pkg/util/k8sutil/backup.go#L63. I think is shoul unified but this will need some more refactoring. One key difference is that etcd pods and (with this patch) PVCs are created (the api call) here while for backups they are created inside `pkg/util/k8sutil`.

pkg/cluster/cluster.go
```diff
@@ -464,13 +480,41 @@ func (c *Cluster) setupServices() error {
...
-func (c *Cluster) createPod(members etcdutil.MemberSet, m *etcdutil.Member, state string, needRecovery bool) error {
+func (c *Cluster) createPVC(m *etcdutil.Member) error {
+	pvc := k8sutil.NewPVC(m, c.cluster.Spec, c.cluster.Name, c.cluster.Namespace, c.cluster.AsOwner())
+	_, err := c.config.KubeCli.Core().PersistentVolumeClaims(c.cluster.Namespace).Create(pvc)
+	if err != nil {
+		return err
+	}
+
+	// TODO: We set timeout to 60s here since PVC binding could take up to 60s for GCE/PD. See https://github.com/kubernetes/kubernetes/issues/40972 .
+	//       Change the wait time once there are official p99 SLA.
```
sgotti: Should we wait for pvc to be created or should we just define them and then continue creating the pod? It'll work also without waiting and the pod will be scheduled only when the pvc goes in bound state.

pkg/util/k8sutil/k8sutil.go
```diff
@@ -211,7 +212,37 @@ func addOwnerRefToObject(o metav1.Object, r metav1.OwnerReference) {
...
+	name := m.PVCName()
+	pvc := &v1.PersistentVolumeClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Labels: map[string]string{
+				"etcd_node":    m.Name,
+				"etcd_cluster": clusterName,
+				"app":          "etcd",
+			},
```
sgotti: @hongchaodeng I remember that you said to just set the owner ref without using labels and I agree. I just noticed that with the current gc logic this could lead to a lot of logged warnings (https://github.com/coreos/etcd-operator/pull/1434/files#diff-40a34818cc02bcfb053e6102e57f3177R173 ) since every pvc in the namespace is inspected. So I added the same labels used for the other resources just to do an initial filtering.

pkg/apis/etcd/v1beta2/cluster.go
```diff
@@ -27,6 +27,8 @@ import (
 const (
 	defaultBaseImage = "quay.io/coreos/etcd"
 	defaultVersion   = "3.1.8"
+
+	minPVSize = 512 // 512MiB
```
xiang90: minPVSizeInMB

pkg/apis/etcd/v1beta2/cluster.go
```diff
@@ -206,6 +207,11 @@ func (c *ClusterSpec) Validate() error {
 				return errors.New("spec: pod labels contains reserved label")
 			}
 		}
+		if c.Pod.PV != nil {
+			if c.Pod.PV.VolumeSizeInMB < minPVSize {
```
xiang90: maybe we just set it back to min instead of erroring it? i forgot about how k8s resource limit works here for the similar case though. probably we want the consistency here.

pkg/cluster/cluster.go
```diff
@@ -404,7 +411,12 @@ func (c *Cluster) startSeedMember(recoverFromBackup bool) error {
 		SecureClient: c.isSecureClient(),
 	}
 	ms := etcdutil.NewMemberSet(m)
-	if err := c.createPod(ms, m, "new", recoverFromBackup); err != nil {
+	if c.UsePodPV() {
```
xiang90: ShouldUsePodPV or IsPVEnabled?

pkg/cluster/cluster.go
```diff
@@ -404,7 +411,12 @@ func (c *Cluster) startSeedMember(recoverFromBackup bool) error {
...
+		if err := c.createPVC(m); err != nil {
+			return fmt.Errorf("failed to create persistent volume claim for seed member (%s): %v", m.Name, err)
+		}
+	}
+	if err := c.createPod(ms, m, "new", recoverFromBackup, c.UsePodPV()); err != nil {
```
xiang90: no need to pass in usePodPV here. the receiver is still on`c`.

pkg/garbagecollection/gc.go
```diff
@@ -158,3 +161,28 @@ func (gc *GC) collectDeployment(option metav1.ListOptions, runningSet map[types.
 
 	return nil
 }
+
+func (gc *GC) collectPVCs(option metav1.ListOptions, runningSet map[types.UID]bool) error {
```
xiang90: @sgotti 

when this lands, we probably already move to k8s 1.8, which supports crd GC natively. so we do not need this thing anymore.

/cc @hongchaodeng

pkg/util/k8sutil/k8sutil.go
```diff
@@ -211,7 +212,37 @@ func addOwnerRefToObject(o metav1.Object, r metav1.OwnerReference) {
...
+		},
+	}
+
+	addOwnerRefToObject(pvc.GetObjectMeta(), owner)
+
+	return pvc
+}
+
+func NewEtcdPod(m *etcdutil.Member, initialCluster []string, clusterName, state, token string, cs api.ClusterSpec, usePVC bool, owner metav1.OwnerReference) *v1.Pod {
```
hongchaodeng: Don't add a new param.

Instead, add a new func `AddPVCIntoPod()` to compose the spec.

pkg/garbagecollection/gc.go
hongchaodeng: +1
We will `rm -rf` entire custom GC code :)

pkg/util/etcdutil/member.go
```diff
@@ -70,6 +70,10 @@ func (m *Member) PeerURL() string {
 	return fmt.Sprintf("%s://%s:2380", m.peerScheme(), m.Addr())
 }
 
+func (m *Member) PVCName() string {
```
hongchaodeng: This is not really etcd related method..
It should be in k8sutil.

test/e2e/basic_test.go
hongchaodeng: Don't change any timeout in test.
All test timeout is delicately tuned because we understand why we allocate this much time and have expectation how long each operation takes.

Let's create a separate set of test for etcd pv scenario.
Additionally, we will probably poll some PVs before test in the future. But we don't need to worry about this right now.

hack/test
sgotti: Since this part doesn't change the logic but only places etcd data on a PV I was thinking to redo all the tests with PV enabled. Using a different test files means a lot of duplicated code and maintenance burden. Perhaps we could have just different timeouts for the non PV and PV cases?

sgotti: @hongchaodeng gentle ICMP ECHO_REQUEST. I'm quite blocked trying to understand how you'd like to organize tests. See above comment.

hongchaodeng: Truly apologize... This message is inundated in too many notifications...

Yes. That sounds good. But The thing I worry about is that this way it creates too many temporary persistent disks on our GCE and is costly... For initial stage, let's just create one test to make sure it work. We can think of ideas to poll those PVs later?

mrIncompetent:
@sgotti Is there anything i can help with?
I would love to use this feature!

hongchaodeng:
@mrIncompetent 
Hasn't seen any update for a while.
Would you like to take it over? I'm fine to move forward if you can submit a PR.

sgotti:
@hongchaodeng @mrIncompetent Sorry, I've ben busy with other stuff and lost your comments notifications. I'm fine if you want to take over or I'll be able to update it in the next days.

hongchaodeng:
@sgotti 
No worries.

@mrIncompetent 
Just wait for the PR and help with review. Thanks!

pkg/garbagecollection/gc.go
sgotti: done

pkg/util/etcdutil/member.go
sgotti: done

sgotti:
@hongchaodeng Updated

test/e2e/e2eutil/spec_util.go
```diff
@@ -37,6 +37,15 @@ func NewCluster(genName string, size int) *api.EtcdCluster {
 	}
 }
 
+func AddPV(c *api.EtcdCluster, storageClass string) {
```
hongchaodeng: AddPVToCluster

pkg/cluster/cluster.go
```diff
@@ -415,6 +422,11 @@ func (c *Cluster) startSeedMember(recoverFromBackup bool) error {
 		SecureClient: c.isSecureClient(),
 	}
 	ms := etcdutil.NewMemberSet(m)
+	if c.IsPodPVEnabled() {
+		if err := c.createPVC(m); err != nil {
```
hongchaodeng: This is duplicated in two places. Can you combine it with createPod() into a new function? Maybe createResource?

pkg/cluster/cluster.go
```diff
@@ -506,6 +529,21 @@ func (c *Cluster) removePod(name string) error {
...
+}
+
+func (c *Cluster) removePVC(name string) error {
+	ns := c.cluster.Namespace
+	err := c.config.KubeCli.Core().PersistentVolumeClaims(ns).Delete(name, nil)
+	if err != nil {
+		if !k8sutil.IsKubernetesResourceNotFoundError(err) {
+			return err
+		}
+		if c.isDebugLoggerEnabled() {
```
hongchaodeng: Can you discard this path:
```
if c.isDebugLoggerEnabled() {...}
```
Just ignore `isDebugLoggerEnabled` now

pkg/cluster/cluster.go
```diff
@@ -506,6 +529,21 @@ func (c *Cluster) removePod(name string) error {
...
 		c.debugLogger.LogPodDeletion(name)
 	}
+
+	return nil
+}
+
+func (c *Cluster) removePVC(name string) error {
+	ns := c.cluster.Namespace
+	err := c.config.KubeCli.Core().PersistentVolumeClaims(ns).Delete(name, nil)
+	if err != nil {
```
hongchaodeng: if err !=nil && !NotFound(err){...

pkg/util/k8sutil/k8sutil.go
hongchaodeng: Not sure what's the context.
Re. labels, use:
```

"etcd_cluster": clusterName,
"app":          "etcd",
```
is good enough

pkg/cluster/cluster.go
```diff
@@ -437,6 +449,10 @@ func (c *Cluster) isSecureClient() bool {
 	return c.cluster.Spec.TLS.IsSecureClient()
 }
 
+func (c *Cluster) IsPodPVEnabled() bool {
```
hongchaodeng: Doc?

pkg/cluster/cluster.go
```diff
@@ -475,6 +491,12 @@ func (c *Cluster) setupServices() error {
 	return k8sutil.CreatePeerService(c.config.KubeCli, c.cluster.Name, c.cluster.Namespace, c.cluster.AsOwner())
 }
 
+func (c *Cluster) createPVC(m *etcdutil.Member) error {
```
hongchaodeng: doc?

pkg/cluster/cluster.go
```diff
@@ -506,6 +529,21 @@ func (c *Cluster) removePod(name string) error {
 	if c.isDebugLoggerEnabled() {
 		c.debugLogger.LogPodDeletion(name)
 	}
+
+	return nil
+}
+
+func (c *Cluster) removePVC(name string) error {
```
hongchaodeng: doc?

pkg/cluster/cluster.go
```diff
@@ -537,6 +575,29 @@ func (c *Cluster) pollPods() (running, pending []*v1.Pod, err error) {
 	return running, pending, nil
 }
 
+func (c *Cluster) pollPVCs() (pvcs []*v1.PersistentVolumeClaim, err error) {
```
hongchaodeng: doc?

pkg/cluster/reconcile.go
```diff
@@ -32,7 +32,7 @@ import (
 // reconcile reconciles cluster current state to desired state specified by spec.
 // - it tries to reconcile the cluster to desired size.
 // - if the cluster needs for upgrade, it tries to upgrade old member one by one.
-func (c *Cluster) reconcile(pods []*v1.Pod) error {
+func (c *Cluster) reconcile(pods []*v1.Pod, pvcs []*v1.PersistentVolumeClaim) error {
```
hongchaodeng: change doc?

pkg/util/k8sutil/k8sutil.go
```diff
@@ -209,6 +214,14 @@ func newEtcdServiceManifest(svcName, clusterName, clusterIP string, ports []v1.S
 	return svc
 }
 
+func AddEtcdVolumeToPod(pod *v1.Pod, m *etcdutil.Member, usePVC bool) {
```
hongchaodeng: doc?

pkg/util/k8sutil/k8sutil.go
```diff
@@ -228,6 +241,36 @@ func NewSeedMemberPod(clusterName string, ms etcdutil.MemberSet, m *etcdutil.Mem
 	return pod
 }
 
+func NewPVC(m *etcdutil.Member, cs api.ClusterSpec, clusterName, namespace string, owner metav1.OwnerReference) *v1.PersistentVolumeClaim {
```
hongchaodeng: doc?

test/e2e/etcd_on_pv_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"os"
+	"testing"
+
+	"github.com/coreos/etcd-operator/test/e2e/e2eutil"
+	"github.com/coreos/etcd-operator/test/e2e/framework"
+)
+
+func TestCreateClusterWithPV(t *testing.T) {
```
hongchaodeng: doc?

pkg/util/k8sutil/k8sutil.go
```diff
@@ -228,6 +241,36 @@ func NewSeedMemberPod(clusterName string, ms etcdutil.MemberSet, m *etcdutil.Mem
...
+			Name:      name,
+			Namespace: namespace,
+			Labels: map[string]string{
+				"etcd_node":    m.Name,
+				"etcd_cluster": clusterName,
+				"app":          "etcd",
+			},
+		},
+		Spec: v1.PersistentVolumeClaimSpec{
+			StorageClassName: &cs.Pod.PV.StorageClass,
```
swapnilgm: Since `storageClassName` is optional in cluster specification, unspecified `storageClassName` in cluster spec leads to `storageClassName = ""` in PVC specification. In case `storageClassName` is not specified in cluster specification, it should be removed from PVC to use default storageClass. Otherwise there won't be option for user to make use of default storage class. Please have look at this https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1

hongchaodeng: I didn't realize empty means default.
If so, the correct thing to do is to change the storageclads field into a pointer. But This is not urgent here. We can fix It after This pr

hongchaodeng:
Hi @sgotti 
Do you have time to address it in next few days? If not, I am having holidays and can help finish the work.

pkg/apis/etcd/v1beta2/cluster.go
```diff
@@ -170,7 +172,6 @@ type PodPolicy struct {
 
 	// PV represents a Persistent Volume resource.
 	// If defined new pods will use a persistent volume to store etcd data.
-	// TODO(sgotti) unimplemented
 	PV *PVSource `json:"pv,omitempty"`
```
hongchaodeng: better use 
`VolumeClaimTemplates []api.PersistentVolumeClaim`
see
https://github.com/kubernetes/api/blob/81aa34336d28aadc3a8e8da7dfd9258c5157e5e4/apps/v1beta1/types.go#L184

hongchaodeng:
Mind if I close this PR in favor of #1861?

hongchaodeng:
Done in https://github.com/coreos/etcd-operator/pull/1861

Issue #1315, opened 2017-07-22
disable v2 by default


xiang90:
etcd operator does not support backup v2 data.

we do not have a good story on v2 yet. probably we should disable v2 by default. and warn users v2 backup is not supported when they try to enable it.

raoofm:
@xiang90 if https://github.com/coreos/etcd/issues/6925 is implemented then yes we dont need v2, else it is very imp to support both v2 and v3.

hongchaodeng:
We can use this flag:
https://github.com/coreos/etcd/blob/master/Documentation/op-guide/configuration.md#--enable-v2

PR #1439, opened 2017-09-25
Add caveat regarding v2 API data.

swizzlr:
We're experimenting with the etcd operator on staging. We're only using it with reproducible data, but we wanted to test the backups all the same, and discovered that v2 data isn't backed up.

This is fine, since most people use the v3 API now anyway.

However, as a cluster operator I would like to disable v2, so consider this PR my +1 to #1315

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,8 @@ example-etcd-cluster-0003       1/1       Running   0          1m
 
 ## Disaster recovery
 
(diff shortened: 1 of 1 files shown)


xiang90:
lgtm

xiang90:
@swizzlr 

this is already mentioned in limitation by the way: https://github.com/coreos/etcd-operator#limitations

swizzlr:
Ah, I see. Thanks anyway.

PR #1448, opened 2017-09-28
etcd-backup-operator: support etcd backup from a new operator 

fanminshi:


fanminshi:
this pr depends on https://github.com/coreos/etcd-operator/pull/1451/

etcd-backup-operator/pkg/client/client.go
```diff
@@ -0,0 +1,118 @@
...
+
+	api "github.com/coreos/etcd-operator/pkg/apis/etcd/v1beta2"
+	"github.com/coreos/etcd-operator/pkg/util/k8sutil"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/client-go/rest"
+)
+
+type BackupCR interface {
```
hongchaodeng: Use generated client instead of rewriting one.

etcd-backup-operator/cmd/backup-operator/main.go
```diff
@@ -0,0 +1,86 @@
...
+	logrus.Infof("Go Version: %s", runtime.Version())
+	logrus.Infof("Go OS/Arch: %s/%s", runtime.GOOS, runtime.GOARCH)
+	logrus.Infof("etcd-backup-operator Version: %v", version.Version)
+	logrus.Infof("Git SHA: %s", version.GitSHA)
+
+	kubecli := k8sutil.MustNewKubeClient()
+	rl, err := resourcelock.New(
+		resourcelock.EndpointsResourceLock,
+		namespace,
+		"backup-operator",
```
hongchaodeng: etcd-backup-operator

etcd-backup-operator/pkg/operator/sync.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return err
+	}
+	eb := cobj.(*api.EtcdBackup)
+	b.handleBackup(&eb.Spec)
+	return nil
+}
+
+func (b *Backup) handleBackup(spec *api.EtcdBackupSpec) error {
+	if spec.StorageType != "s3" {
```
hongchaodeng: Use switch

etcd-backup-operator/pkg/operator/sync.go
```diff
@@ -0,0 +1,67 @@
...
+		return fmt.Errorf("unknown storage type: %v", spec.StorageType)
+	}
+	if spec.BackupStorageSource.S3 == nil {
```
hongchaodeng: This should be written as validation method and checked at the beginning.
It shouldn't return error here because workqueue will retry.

etcd-backup-operator/pkg/operator/s3backup.go
```diff
@@ -0,0 +1,213 @@
...
+	s3b := &s3Backup{
+		namespace:   b.namespace,
+		name:        b.name,
+		clusterName: clusterName,
+		s3bucket:    s3.S3Bucket,
+		prefix:      ToS3Prefix(s3.Prefix, b.namespace, clusterName),
+		awsSecret:   s3.AWSSecret,
+		kubecli:     b.kubecli,
+	}
+	return s3b.saveSnap()
```
hongchaodeng: It seems unnecessary to define a method for it.
We aren't returning *s3Backup anywhere.

etcd-backup-operator/cmd/backup-operator/main.go
```diff
@@ -0,0 +1,86 @@
+package main
+
+import (
+	"context"
+	"os"
+	"runtime"
+	"time"
+
+	"k8s.io/api/core/v1"
```
hongchaodeng: Order the imports proper.

etcd-backup-operator/pkg/apis/backup/v1beta2/register.go
```diff
@@ -0,0 +1,6 @@
+package v1beta2
+
+const (
+	CRDResourceKind   = "EtcdBackup"
```
hongchaodeng: EtcdBackupResourceXXX

etcd-backup-operator/cmd/backup-operator/main.go
```diff
@@ -0,0 +1,86 @@
...
+	eventBroadcaster.StartLogging(logrus.Infof)
+	eventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: v1core.New(kubecli.Core().RESTClient()).Events(namespace)})
+	return eventBroadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: name})
+}
+
+func run(stop <-chan struct{}) {
+	b := operator.New()
+	err := b.Start(context.TODO())
+	if err != nil {
+		logrus.Infof("operator stopped with %v", err)
```
hongchaodeng: In what case will Start() return an error?

hongchaodeng:
It would be better to split first the main() function and controller logic out into a separate PR as next step

etcd-backup-operator/pkg/operator/sync.go
fanminshi: how does workqueue retrying work?

etcd-backup-operator/cmd/backup-operator/main.go
fanminshi: when the following is implemented, then it might throw an error.
```go
func (b *Backup) init(ctx context.Context) error {
	// TODO: check if crd exists
	return nil
}
```

hongchaodeng: What specific error will it throw?

fanminshi:
close this in favor of https://github.com/coreos/etcd-operator/pull/1475

PR #1457, opened 2017-09-28
etcd-backup-operator: remove extra func in backup operator 

fanminshi:


etcd-backup-operator/hack/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM alpine
```
hongchaodeng: This pr should be  purely removing the extra function. Don't couple and mix  extra stuff.

fanminshi: kk, I'll have another pr that does the build stuff.

fanminshi:
only remove func PTAL

PR #1462, opened 2017-09-29
hack: fix Dockerfile path in etcd-backup-operator push

fanminshi:


fanminshi:
cc/ @hongchaodeng PTAL

PR #1463, opened 2017-09-29
*: add etcd backup controller

fanminshi:


fanminshi:
only controller logic cc/ @hongchaodeng

fanminshi:
checking crd logic added.

hongchaodeng:
It's better to let @hasbro17 review this once he comes back on Monday.
It is better to make him catch up as well.

pkg/etcd-backup-operator/controller/operator.go
```diff
@@ -0,0 +1,64 @@
+package controller
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/Sirupsen/logrus"
```
hasbro17: Group this together with non etcd-operator imports on line 15.

pkg/etcd-backup-operator/controller/operator.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return err
+	}
+	go b.run(ctx)
+	<-ctx.Done()
+	return ctx.Err()
+}
+
+func (b *Backup) init(ctx context.Context) error {
+	logrus.Fatalf("checking if crd %v exists", v1beta2.EtcdBackupCRDName)
```
hasbro17: You probably meant `logrus.Infof()`. Fatalf will exit.

Also I'm thinking the backup operator should have a logger similar to the etcd-operator, with a field like `pkg=backup-controller` or something. It makes it easier to search through the logs if they are tagged like that.
https://github.com/coreos/etcd-operator/blob/master/pkg/controller/controller.go#L76

fanminshi: sounds like a good idea.

pkg/etcd-backup-operator/controller/controller.go
```diff
@@ -0,0 +1,72 @@
...
+		DeleteFunc: b.onDelete,
+	}, cache.Indexers{})
+
+	defer b.queue.ShutDown()
+
+	logrus.Info("starting backup controller")
+	go b.informer.Run(ctx.Done())
+
+	if !cache.WaitForCacheSync(ctx.Done(), b.informer.HasSynced) {
+		logrus.Error("Timed out waiting for caches to sync")
```
hongchaodeng: This is not a timeout error, right?

pkg/etcd-backup-operator/controller/operator.go
```diff
@@ -0,0 +1,64 @@
+package controller
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/Sirupsen/logrus"
+	v1beta2 "github.com/coreos/etcd-operator/pkg/apis/etcd/v1beta2"
```
hongchaodeng:
```
api "github.com/coreos/etcd-operator/pkg/apis/etcd/v1beta2"
```

pkg/etcd-backup-operator/controller/operator.go
hongchaodeng: What is this for?

pkg/etcd-backup-operator/controller/operator.go
```diff
@@ -0,0 +1,64 @@
...
+	go b.run(ctx)
+	<-ctx.Done()
+	return ctx.Err()
+}
+
+func (b *Backup) init(ctx context.Context) error {
+	logrus.Fatalf("checking if crd %v exists", v1beta2.EtcdBackupCRDName)
+	_, err := b.kubeExtClient.ApiextensionsV1beta1().CustomResourceDefinitions().Get(v1beta2.EtcdBackupCRDName, v1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("crd %v is not found", v1beta2.EtcdBackupCRDName)
```
hongchaodeng: This is not only not found error?

pkg/etcd-backup-operator/controller/operator.go
```diff
@@ -0,0 +1,64 @@
...
+		return err
+	}
+	go b.run(ctx)
+	<-ctx.Done()
+	return ctx.Err()
+}
+
+func (b *Backup) init(ctx context.Context) error {
+	logrus.Fatalf("checking if crd %v exists", v1beta2.EtcdBackupCRDName)
+	_, err := b.kubeExtClient.ApiextensionsV1beta1().CustomResourceDefinitions().Get(v1beta2.EtcdBackupCRDName, v1.GetOptions{})
```
hongchaodeng: I think we can assume CRD exist for initial version.

pkg/etcd-backup-operator/controller/sync.go
```diff
@@ -0,0 +1,83 @@
...
+		b.queue.AddRateLimited(key)
+		return
+	}
+
+	b.queue.Forget(key)
+	// Report that, even after several retries, we could not successfully process this key
+	logrus.Infof("Dropping Backup (%v) out of the queue: %v", key, err)
+}
+
+// func (b *Backup) backupStatus() {
```
hongchaodeng: remove commented code

pkg/etcd-backup-operator/controller/sync.go
```diff
@@ -0,0 +1,83 @@
...
+	return nil
+}
+
+func (b *Backup) handleBackup(spec *api.EtcdBackupSpec) error {
+	// TODO:
+	return nil
+}
+
+// don't retry on any backup failure.
+const maxRetries = 0
```
hongchaodeng: Why retry 0?

pkg/etcd-backup-operator/controller/sync.go
```diff
@@ -0,0 +1,83 @@
...
+func (b *Backup) processItem(key string) error {
+	obj, exists, err := b.indexer.GetByKey(key)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		return nil
+	}
+
+	cobj, err := scheme.Scheme.DeepCopy(obj)
```
hongchaodeng: Use DeepCopy() method

hongchaodeng: Why copy it BTW?

pkg/etcd-backup-operator/controller/sync.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		return err
+	}
+	eb := cobj.(*api.EtcdBackup)
+	b.handleBackup(&eb.Spec)
+	return nil
+}
+
+func (b *Backup) handleBackup(spec *api.EtcdBackupSpec) error {
```
hongchaodeng: remove this method since it is not useful in this PR.

pkg/etcd-backup-operator/controller/operator.go
fanminshi: will remove since checking crd is not necessary

pkg/etcd-backup-operator/controller/controller.go
fanminshi: The doc says 
```
// WaitForCacheSync waits for caches to populate.  It returns true if it was successful, false
// if the controller should shutdown
```

Maybe `logrus.Error("cache failed to sync.")` is more appropriate? cc/ @hasbro17

hongchaodeng: Why even error? "It should shutdown" from the docs.

fanminshi: should we log the reason of shutting done?

hongchaodeng: Yes. We should.
But it's not relevant here. Right?

fanminshi:
comments addressed. PTAL cc/ @hongchaodeng @hasbro17

cmd/etcd-backup-operator/main.go
```diff
@@ -74,5 +76,9 @@ func createRecorder(kubecli kubernetes.Interface, name, namespace string) record
 }
 
 func run(stop <-chan struct{}) {
-	// TODO
+	c := controller.New()
+	err := c.Start(context.TODO())
+	if err != nil {
+		logrus.Fatalf("operator stopped with %v", err)
```
hongchaodeng: with error: %v

pkg/etcd-backup-operator/controller/controller.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/Sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+)
+
+func init(){
+	logrus.WithField("pkg", "controller"),
```
hongchaodeng: Why initialize logrus with such field?
All logrus logging will now contain this field even if it's outside "controller" package.

pkg/etcd-backup-operator/controller/operator.go
```diff
@@ -0,0 +1,48 @@
...
+		kubecli:       k8sutil.MustNewKubeClient(),
+		backupCRCli:   client.MustNewInCluster(),
+		kubeExtClient: k8sutil.MustNewKubeExtClient(),
+	}
+}
+
+// Start starts the Backup operator.
+func (b *Backup) Start(ctx context.Context) error {
+	// TODO: check if crd exists.
+	go b.run(ctx)
```
hongchaodeng: add a logging that it starts. Print out specific details like namespace/name

pkg/etcd-backup-operator/controller/controller.go
fanminshi: just realized that. removing this in favor of https://github.com/coreos/etcd-operator/blob/master/pkg/controller/controller.go#L76

pkg/etcd-backup-operator/controller/operator.go
fanminshi: no need since ` namespace/name` info belongs to pod's detail.

fanminshi:

hongchaodeng:
lgtm

defer to @hasbro17 for final approval

pkg/etcd-backup-operator/controller/operator.go
```diff
@@ -0,0 +1,52 @@
+package controller
+
+import (
+	"context"
+	"os"
+
+	"github.com/Sirupsen/logrus"
```
hasbro17: Move this down to line 13

fanminshi: done!

hasbro17:
LGTM

hongchaodeng:
@fanminshi 
I just realize that a lot of source code files don't have LICENSE headers.
Can you do a follow-up PR to fix them?

fanminshi:
@hongchaodeng good catch.

PR #1481, opened 2017-10-03
controller: remove unnecessary name from backup struct

fanminshi:


fanminshi:
cc/ @hongchaodeng @hasbro17

xiang90:
lgtm

PR #1486, opened 2017-10-04
*: add s3 backup

hongchaodeng:


fanminshi:
@hongchaodeng ptal.

fanminshi:
@hongchaodeng refactored code. PTAL.

pkg/backup/backup.go
```diff
@@ -30,10 +31,9 @@ import (
 	"github.com/coreos/etcd-operator/pkg/util/constants"
 	"github.com/coreos/etcd-operator/pkg/util/etcdutil"
 	"github.com/coreos/etcd-operator/pkg/util/k8sutil"
-
 	"github.com/coreos/etcd/clientv3"
```
hasbro17: Move this down to line 36 to group with non-operator imports.

pkg/backup/backup.go
```diff
@@ -179,94 +180,128 @@ func (b *Backup) Run() {
 	}
 }
 
+// saveSnap saves the latest snapshot if the given current revision is greater lastSnapRev.
 func (b *Backup) saveSnap(lastSnapRev int64) (int64, error) {
-	podList, err := b.kclient.Core().Pods(b.namespace).List(k8sutil.ClusterListOpt(b.clusterName))
+	logrus.Printf("saving backup for cluster (%s)", b.clusterName)
```
hongchaodeng: No need for this logging.

pkg/backup/backup.go
```diff
@@ -179,94 +180,128 @@ func (b *Backup) Run() {
...
+func WriteSnap(version string, rev int64, be backend, rc io.ReadCloser) (size int64, err error) {
+	n, err := be.save(version, rev, rc)
+	if err != nil {
+		return 0, err
+	}
+	return n, nil
+}
+
+// GetSnap gets the snapshot from the cluster member with the maximum revision.
+func GetSnap(ctx context.Context, kclient kubernetes.Interface, tls *tls.Config, namespace, clusterName string) (rc io.ReadCloser, etcdVersion string, rev int64, err error) {
```
hongchaodeng: Our convention is no var def at the end. 
```
(rc io.ReadCloser, etcdVersion string, rev int64, err error) 
```

pkg/backup/backup.go
```diff
@@ -179,94 +180,128 @@ func (b *Backup) Run() {
...
+
+func (b *Backup) getLatestBackupStatus() backupapi.BackupStatus {
+	return b.recentBackupsStatus[len(b.recentBackupsStatus)-1]
+}
 
-	return nil
+func getRunningPods(kclient kubernetes.Interface, namespace, clusterName string) ([]*v1.Pod, error) {
+	podList, err := kclient.Core().Pods(namespace).List(k8sutil.ClusterListOpt(clusterName))
+	if err != nil {
+		return nil, err
```
hasbro17: Wrap the error and then use it above in GetSnap():
`return nil, fmt.Errorf("failed to list pods: %v", err)`

pkg/backup/backup.go
```diff
@@ -179,94 +180,128 @@ func (b *Backup) Run() {
...
+		return 0, err
+	}
+	return n, nil
+}
+
+// GetSnap gets the snapshot from the cluster member with the maximum revision.
+func GetSnap(ctx context.Context, kclient kubernetes.Interface, tls *tls.Config, namespace, clusterName string) (rc io.ReadCloser, etcdVersion string, rev int64, err error) {
+	pods, err := getRunningPods(kclient, namespace, clusterName)
+	if err != nil {
+		return nil, "", 0, fmt.Errorf("no running etcd pods found")
```
hasbro17: Wrap the error from getRunningPods and return it. It would actually be some api error from `getRunningPods` and not that no running pods were found.
I think you forgot to check for `len(pods)==0`

hasbro17: `len(pods)==0` would be `no running etcd pods found`. So check for that here outside of `getRunningPods()`

pkg/etcd-backup-operator/controller/s3_backup.go
```diff
@@ -0,0 +1,77 @@
+package controller
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/aws/aws-sdk-go/aws/session"
```
hasbro17: Move this down from operator imports

pkg/etcd-backup-operator/controller/s3_backup.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create s3 backend: (%v)", err)
+	}
+	defer os.RemoveAll(s3Dir)
+
+	_, err = backup.WriteSnap(version, rev, be, rc)
+	if err != nil {
+		return fmt.Errorf("failed to write snap to s3 backend: (%v)", err)
+	}
+	return err
```
hasbro17: `return nil` to be explicit

pkg/etcd-backup-operator/controller/s3_backup.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func (b *Backup) setupAWSConfig(secret string) (awsDir string, so *session.Options, err error) {
+	awsDir, err = ioutil.TempDir(tmp, "")
+	if err != nil {
+		return "", nil, err
```
hasbro17: wrap error.

pkg/backup/s3.go
```diff
@@ -33,6 +33,14 @@ type s3Backend struct {
 	dir string
 }
 
+type S3Backend struct {
+	s3Backend
+}
+
+func NewS3Backend(S3 *s3.S3, dir string) *S3Backend {
```
hasbro17: How about we make the `backend` interface public?
https://github.com/coreos/etcd-operator/blob/master/pkg/backup/backend.go#L19
And keep `s3Backend` private. Then `NewS3Backend()` would just be:
```

func NewS3Backend(S3 *s3.S3, dir string) Backend {
    return &s3Backend{S3, dir}
}
```
@fanminshi Let me know what you think.

pkg/etcd-backup-operator/controller/s3_backup.go
```diff
@@ -0,0 +1,77 @@
...
+	defer os.RemoveAll(s3Dir)
+
+	_, err = backup.WriteSnap(version, rev, be, rc)
+	if err != nil {
+		return fmt.Errorf("failed to write snap to s3 backend: (%v)", err)
+	}
+	return err
+}
+
+func (b *Backup) setupAWSConfig(secret string) (awsDir string, so *session.Options, err error) {
```
hasbro17: Any reason to use named return variables? We're not doing a naked return here and since we're not using named return variables anywhere else it might be best to not name them to stay consistent.

fanminshi: @hasbro17 I hope that the named variables give some context on what to return. but I can remove them if needed.

hasbro17: Sure I get it. But it's probably best to not name them since that's the convention we're using through the rest of the codebase.
For most functions, it's already clear to understand from the parameter names of the return statement.
Like `awsDir`. And generally the type tells us more than the abbreviated name anyway, like `so` vs `*session.Options`.

fanminshi: sounds good.

pkg/backup/s3.go
fanminshi: @hasbro17 that's my plan in this https://github.com/coreos/etcd-operator/pull/1487.

pkg/etcd-backup-operator/controller/s3_backup.go
```diff
@@ -0,0 +1,77 @@
...
+
+	so, err = backupstorage.SetupAWSConfig(b.kubecli, b.namespace, secret, awsDir)
+	if err != nil {
+		return "", nil, err
+	}
+
+	return awsDir, so, nil
+}
+
+func (b *Backup) makeS3Backend(so *session.Options, prefix, bucket string) (s3Dir string, be *backup.S3Backend, err error) {
```
hasbro17: Does this need to be a method? It doesn't use anything of the Backup controller. Perhaps make it a function.

fanminshi: you are right.

fanminshi:
all fixed. PTAL cc/ @hongchaodeng @hasbro17

fanminshi:
@hasbro17 fixed backup ci issue. PTAL

pkg/backup/backup.go
```diff
@@ -162,6 +162,7 @@ func (b *Backup) Run() {
 		if err != nil {
 			logrus.Errorf("failed to save snapshot: %v", err)
 		}
+
```
hongchaodeng: remove this line

pkg/backup/backup.go
```diff
@@ -174,94 +175,137 @@ func (b *Backup) Run() {
...
-	podList, err := b.kclient.Core().Pods(b.namespace).List(k8sutil.ClusterListOpt(b.clusterName))
+	start := time.Now()
+
+	ctx, cancel := context.WithTimeout(context.Background(), constants.DefaultSnapshotTimeout)
+	rc, etcdVersion, rev, err := GetSnap(ctx, b.kclient, b.etcdTLSConfig, b.namespace, b.clusterName)
+	defer cancel()
+	defer rc.Close()
 	if err != nil {
-		return lastSnapRev, err
+		return lastSnapRev, fmt.Errorf("failed to get snapshot: (%v)", err)
```
hongchaodeng: `get cluster (%s)'s snapshot`

pkg/backup/backup.go
```diff
@@ -174,94 +175,137 @@ func (b *Backup) Run() {
...
 
-	log.Printf("saving backup for cluster (%s)", b.clusterName)
-	if err := b.writeSnap(member, rev); err != nil {
-		err = fmt.Errorf("write snapshot failed: %v", err)
-		return lastSnapRev, err
+	return 0, nil
+}
+
+// WriteSnap writes the snapshot to the backend.
+func WriteSnap(version string, rev int64, be backend.Backend, rc io.ReadCloser) (int64, error) {
```
hongchaodeng: Only need to use `io.Reader`, right?

pkg/backup/backup.go
```diff
@@ -174,94 +175,137 @@ func (b *Backup) Run() {
...
-	ctx, cancel := context.WithTimeout(context.Background(), constants.DefaultRequestTimeout)
-	resp, err := etcdcli.Maintenance.Status(ctx, m.ClientURL())
-	cancel()
+	// when GetSnap fails, close etcdcli.
+	defer func() {
+		if err != nil {
+			etcdcli.Close()
+		}
+	}()
+	cctx, ccancel := context.WithTimeout(context.Background(), constants.DefaultRequestTimeout)
```
hongchaodeng: cctx, ccancel -> ctx, cancel

pkg/backup/backup.go
```diff
@@ -174,94 +175,137 @@ func (b *Backup) Run() {
...
+	resp, err := etcdcli.Maintenance.Status(cctx, m.ClientURL())
+	ccancel()
 	if err != nil {
-		return err
+		return nil, "", 0, fmt.Errorf("failed to get member status (%v)", err)
 	}
 
-	ctx, cancel = context.WithTimeout(context.Background(), constants.DefaultSnapshotTimeout)
 	rc, err := etcdcli.Maintenance.Snapshot(ctx)
```
hongchaodeng: etcdcli.Close() here

pkg/backup/backup.go
```diff
@@ -174,94 +175,137 @@ func (b *Backup) Run() {
...
 	if err != nil {
-		return fmt.Errorf("failed to receive snapshot (%v)", err)
+		return nil, "", 0, fmt.Errorf("failed to get member snapshot (%v)", err)
 	}
-	defer cancel()
-	defer rc.Close()
+	// when context for getting/saving snapshot finishes, close the etcdcli.
+	go func() {
```
hongchaodeng: remove this. close etcdcli if it's not needed anymore

pkg/etcd-backup-operator/controller/s3_backup.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to create aws config/cred dir: (%v)", err)
+	}
+
+	so, err := backupstorage.SetupAWSConfig(b.kubecli, b.namespace, secret, awsDir)
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to setup aws config: (%v)", err)
+	}
+
+	return awsDir, so, nil
```
hongchaodeng: Why return `awsDir`? It's not used anywhere

pkg/etcd-backup-operator/controller/s3_backup.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func makeS3Backend(so *session.Options, prefix, bucket string) (string, backend.Backend, error) {
+	s3cli, err := backupS3.NewFromSessionOpt(bucket, prefix, *so)
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to create aws cli: (%v)", err)
+	}
+
+	s3Dir, err := ioutil.TempDir(tmp, "")
```
hongchaodeng: s3Dir -> backupDir

pkg/etcd-backup-operator/controller/s3_backup.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/aws/aws-sdk-go/aws/session"
+	"golang.org/x/net/context"
+)
+
+const (
+	tmp = "/tmp"
+)
+
+// TODO: remove this and use backend interface for other options (PV, Azure)
+func (b *Backup) handleS3(clusterName string, s3 *api.S3Source) error {
```
hongchaodeng: doc string this method

pkg/backup/backup.go
fanminshi: we can't close `etcdcli` if `WriteSnap` is not done with `rc`. Closing  `etcdcli` before `WriteSnap` results rc being closed.

pkg/etcd-backup-operator/controller/s3_backup.go
fanminshi: it is used 
```go
	s3Dir, be, err := makeS3Backend(so, prefix, s3.S3Bucket)
	if err != nil {
		return fmt.Errorf("failed to create s3 backend: (%v)", err)
	}
	defer os.RemoveAll(s3Dir) // used here.
````

File: pkg/etcd-backup-operator/controller/s3_backup.go (hunk 19 above)
Comment (fanminshi) [reply pkg/etcd-backup-operator/controller/s3_backup.go]:
```go
	awsDir, so, err := b.setupAWSConfig(s3.AWSSecret)
	if err != nil {
		return fmt.Errorf("failed to set up aws config: (%v)", err)
	}
	defer os.RemoveAll(awsDir) // used here.
```

pkg/etcd-backup-operator/controller/s3_backup.go
fanminshi:
```go
	awsDir, so, err := b.setupAWSConfig(s3.AWSSecret)
	if err != nil {
		return fmt.Errorf("failed to set up aws config: (%v)", err)
	}
	defer os.RemoveAll(awsDir) // used here
```

pkg/backup/backup.go
fanminshi: `ctx` is already defined in `GetSnap(ctx context.Context, ...)` I don't think we should override that.

pkg/backup/backup.go
fanminshi: yep.

pkg/etcd-backup-operator/controller/s3_backup.go
hongchaodeng:
```
os.RemoveAll(awsDir)
```
Can be done inside this func, right?

pkg/backup/backup.go
fanminshi: actually, we need rc io.ReadCloser. so we can call rc.Close()

hongchaodeng: Where is Close() called?

pkg/backup/backup.go
hongchaodeng: That's fine. Just override that.

pkg/backup/backup.go
```diff
@@ -174,94 +175,137 @@ func (b *Backup) Run() {
...
 	}
-	defer etcdcli.Close()
-
-	ctx, cancel := context.WithTimeout(context.Background(), constants.DefaultRequestTimeout)
-	resp, err := etcdcli.Maintenance.Status(ctx, m.ClientURL())
-	cancel()
+	// when GetSnap fails, close etcdcli.
+	defer func() {
+		if err != nil {
+			etcdcli.Close()
```
hongchaodeng: Why this close?

pkg/backup/backup.go
```diff
@@ -174,94 +175,137 @@ func (b *Backup) Run() {
...
+	if err != nil {
+		return 0, err
 	}
-	return rev, nil
+	return n, nil
 }
 
-func (b *Backup) writeSnap(m *etcdutil.Member, rev int64) error {
-	start := time.Now()
+// GetSnap gets the snapshot from the cluster member with the maximum revision.
```
hongchaodeng: document that ctx should be closed when returned io.ReadCloser is done reading.

hongchaodeng: Discussed with @fanminshi . My fault I didn't understand Snapshot() api good enough..
Need to refactor this part further to avoid background goroutine.

hongchaodeng:
Can you separate the refactor work into a separate PR first? Thanks!

fanminshi:
pr for only backup refactoring https://github.com/coreos/etcd-operator/pull/1489

pkg/backup/backup.go
```diff
@@ -174,94 +175,137 @@ func (b *Backup) Run() {
...
 	}
 
-	log.Printf("saving backup for cluster (%s)", b.clusterName)
-	if err := b.writeSnap(member, rev); err != nil {
-		err = fmt.Errorf("write snapshot failed: %v", err)
-		return lastSnapRev, err
+	return 0, nil
+}
+
+// WriteSnap writes the snapshot to the backend.
```
xiang90: explain what are the arg used for. if it only writes snapshot to be, why version string is needed?

pkg/backup/backup.go
```diff
@@ -174,94 +175,137 @@ func (b *Backup) Run() {
 	}
 }
 
+// saveSnap saves the latest snapshot if the given current revision is greater lastSnapRev.
```
xiang90: what happens if it is not greater?

pkg/backup/backup.go
```diff
@@ -174,94 +175,137 @@ func (b *Backup) Run() {
...
 	}
+	return rev
+}
+
+func (b *Backup) getLatestBackupStatus() backupapi.BackupStatus {
+	return b.recentBackupsStatus[len(b.recentBackupsStatus)-1]
+}
 
-	return nil
+func getRunningPods(kclient kubernetes.Interface, namespace, clusterName string) ([]*v1.Pod, error) {
```
xiang90: dont we already have this utility func somewhere in the codebase?

pkg/cluster/backupstorage/s3.go
```diff
@@ -87,7 +87,7 @@ func (s *s3) Delete() error {
 	return nil
 }
 
-func setupAWSConfig(kubecli kubernetes.Interface, ns, secret, dir string) (*session.Options, error) {
+func SetupAWSConfig(kubecli kubernetes.Interface, ns, secret, dir string) (*session.Options, error) {
```
xiang90: doc string on public func

hongchaodeng:
Closing this since it's now replaced by multiple smaller PRs

PR #1474, opened 2017-10-03
Changed github.com/Sirupsen to github.com/sirupsen

jduhamel:
Changed package name from github.com/Sirupsen
to github.com/sirupsen to reflect upstream package name change.
Changed go files to new import for logging.
Changed glide.lock and glide.yaml for new package name

PR #1488, opened 2017-10-06
glide: update hash of glide.lock

hasbro17:
The hash in glide.lock was out of date causing the following warning on doing `glide install --strip-vendor`
```
Lock file may be out of date. Hash check of YAML failed. You may need to run 'update'
```

Most probably it was caused by not updating the lock file properly in #1474 
https://github.com/coreos/etcd-operator/pull/1474/files#diff-f16a80eae23d5b298c2652448ec420cfR1


hongchaodeng:
lgtm

PR #1489, opened 2017-10-06
*: split backup into BackupController and BackupManager.

fanminshi:
Previously, controlling logic and snapshot retrieving/storing logic are coupled together into a struct called backup.

This pr separates controlling logic and snapshot retrieving/storing logic out of backup into BackupController and BackupManager:

BackupController is responsible for controlling when to perform backup.
BackupManager is responsible for retrieving/storing backup.

### DIFF
--- a/pkg/backup/backup.go
+++ b/pkg/backup/backup.go
@@ -53,7 +53,6 @@ type BackupManager struct {
 	clusterName   string
 	namespace     string
 	etcdTLSConfig *tls.Config
-	selfHosted    bool
 
 	be backend.Backend
 }
@@ -125,7 +124,6 @@ func NewBackupController(kclient kubernetes.Interface, clusterName, ns string, s
 		namespace:     ns,
 		be:            be,
 		etcdTLSConfig: tc,
-		selfHosted:    sp.SelfHosted != nil,
 	}
 
 	return &BackupController{
--- a/pkg/backup/backup.go
+++ b/pkg/backup/backup.go
@@ -46,25 +46,30 @@ const (
 	maxRecentBackupStatusCount = 10
 )
 
-type Backup struct {
+// BackupManager backups an etcd cluster.
+type BackupManager struct {
 	kclient kubernetes.Interface
 
 	clusterName   string
 	namespace     string
-	policy        api.BackupPolicy
-	listenAddr    string
 	etcdTLSConfig *tls.Config
 	selfHosted    bool
 
 	be backend.Backend
+}
 
-	backupNow chan chan backupNowAck
-
+// BackupController controls when to do backup based on backup policy and incoming HTTP backup requests.
+type BackupController struct {
+	listenAddr    string
+	backupNow     chan chan backupNowAck
+	policy        api.BackupPolicy
+	backupManager *BackupManager
 	// recentBackupStatus keeps the statuses of 'maxRecentBackupStatusCount' recent backups.
 	recentBackupsStatus []backupapi.BackupStatus
 }
 
-func New(kclient kubernetes.Interface, clusterName, ns string, sp api.ClusterSpec, listenAddr string) (*Backup, error) {
+// NewBackupController creates a BackupController.
+func NewBackupController(kclient kubernetes.Interface, clusterName, ns string, sp api.ClusterSpec, listenAddr string) (*BackupController, error) {
 	bdir := path.Join(constants.BackupMountDir, PVBackupV1, clusterName)
 	// We created not only backup dir and but also tmp dir under it.
 	// tmp dir is used to store intermediate snapshot files.
@@ -114,36 +119,41 @@ func New(kclient kubernetes.Interface, clusterName, ns string, sp api.ClusterSpe
 		}
 	}
 
-	return &Backup{
+	bm := &BackupManager{
 		kclient:       kclient,
 		clusterName:   clusterName,
 		namespace:     ns,
-		policy:        *sp.Backup,
-		listenAddr:    listenAddr,
 		be:            be,
 		etcdTLSConfig: tc,
 		selfHosted:    sp.SelfHosted != nil,
+	}
 
-		backupNow: make(chan chan backupNowAck),
+	return &BackupController{
+		listenAddr:    listenAddr,
+		backupNow:     make(chan chan backupNowAck),
+		policy:        *sp.Backup,
+		backupManager: bm,
 	}, nil
 }
 
-func (b *Backup) Run() {
-	go b.startHTTP()
+// Run starts BackupController controller where it
+// controlls backups based on backup policy and HTTP backup requests.
+func (bc *BackupController) Run() {
+	go bc.startHTTP()
 
-	lastSnapRev := b.getLatestBackupRev()
+	lastSnapRev := bc.backupManager.getLatestBackupRev()
 	interval := constants.DefaultSnapshotInterval
-	if b.policy.BackupIntervalInSecond != 0 {
-		interval = time.Duration(b.policy.BackupIntervalInSecond) * time.Second
+	if bc.policy.BackupIntervalInSecond != 0 {
+		interval = time.Duration(bc.policy.BackupIntervalInSecond) * time.Second
 	}
 
 	go func() {
-		if b.policy.MaxBackups == 0 {
+		if bc.policy.MaxBackups == 0 {
 			return
 		}
 		for {
 			<-time.After(10 * time.Second)
-			err := b.be.Purge(b.policy.MaxBackups)
+			err := bc.backupManager.be.Purge(bc.policy.MaxBackups)
 			if err != nil {
 				logrus.Errorf("fail to purge backups: %v", err)
 			}
@@ -154,30 +164,39 @@ func (b *Backup) Run() {
 		var ackchan chan backupNowAck
 		select {
 		case <-time.After(interval):
-		case ackchan = <-b.backupNow:
+		case ackchan = <-bc.backupNow:
 			logrus.Info("received a backup request")
 		}
 
-		rev, err := b.saveSnap(lastSnapRev)
+		bs, err := bc.backupManager.SaveSnap(lastSnapRev)
 		if err != nil {
 			logrus.Errorf("failed to save snapshot: %v", err)
 		}
-		lastSnapRev = rev
+
+		if bs != nil {
+			lastSnapRev = bs.Revision
+			bc.recentBackupsStatus = append(bc.recentBackupsStatus, *bs)
+			if len(bc.recentBackupsStatus) > maxRecentBackupStatusCount {
+				bc.recentBackupsStatus = bc.recentBackupsStatus[1:]
+			}
+		}
 
 		if ackchan != nil {
 			ack := backupNowAck{err: err}
 			if err == nil {
-				ack.status = b.getLatestBackupStatus()
+				ack.status = bc.recentBackupsStatus[len(bc.recentBackupsStatus)-1]
 			}
 			ackchan <- ack
 		}
 	}
 }
 
-func (b *Backup) saveSnap(lastSnapRev int64) (int64, error) {
-	podList, err := b.kclient.Core().Pods(b.namespace).List(k8sutil.ClusterListOpt(b.clusterName))
+// SaveSnap saves the snapshot if snapshot's revision is greater than the given lastSnapRev
+// and returns a BackupStatus containing saving backup metadata if SaveSnap succeeds.
+func (bm *BackupManager) SaveSnap(lastSnapRev int64) (*backupapi.BackupStatus, error) {
+	podList, err := bm.kclient.Core().Pods(bm.namespace).List(k8sutil.ClusterListOpt(bm.clusterName))
 	if err != nil {
-		return lastSnapRev, err
+		return nil, err
 	}
 
 	var pods []*v1.Pod
@@ -191,38 +210,39 @@ func (b *Backup) saveSnap(lastSnapRev int64) (int64, error) {
 	if len(pods) == 0 {
 		msg := "no running etcd pods found"
 		logrus.Warning(msg)
-		return lastSnapRev, fmt.Errorf(msg)
+		return nil, fmt.Errorf(msg)
 	}
-	member, rev := getMemberWithMaxRev(pods, b.etcdTLSConfig, b.selfHosted)
+	member, rev := getMemberWithMaxRev(pods, bm.etcdTLSConfig)
 	if member == nil {
 		logrus.Warning("no reachable member")
-		return lastSnapRev, fmt.Errorf("no reachable member")
+		return nil, fmt.Errorf("no reachable member")
 	}
 
 	if rev <= lastSnapRev {
 		logrus.Info("skipped creating new backup: no change since last time")
-		return lastSnapRev, nil
+		return nil, nil
 	}
 
-	log.Printf("saving backup for cluster (%s)", b.clusterName)
-	if err := b.writeSnap(member, rev); err != nil {
+	log.Printf("saving backup for cluster (%s)", bm.clusterName)
+	bs, err := bm.writeSnap(member, rev)
+	if err != nil {
 		err = fmt.Errorf("write snapshot failed: %v", err)
-		return lastSnapRev, err
+		return nil, err
 	}
-	return rev, nil
+	return bs, nil
 }
 
-func (b *Backup) writeSnap(m *etcdutil.Member, rev int64) error {
+func (bm *BackupManager) writeSnap(m *etcdutil.Member, rev int64) (*backupapi.BackupStatus, error) {
 	start := time.Now()
 
 	cfg := clientv3.Config{
 		Endpoints:   []string{m.ClientURL()},
 		DialTimeout: constants.DefaultDialTimeout,
-		TLS:         b.etcdTLSConfig,
+		TLS:         bm.etcdTLSConfig,
 	}
 	etcdcli, err := clientv3.New(cfg)
 	if err != nil {
-		return fmt.Errorf("failed to create etcd client (%v)", err)
+		return nil, fmt.Errorf("failed to create etcd client (%v)", err)
 	}
 	defer etcdcli.Close()
 
@@ -230,38 +250,34 @@ func (b *Backup) writeSnap(m *etcdutil.Member, rev int64) error {
 	resp, err := etcdcli.Maintenance.Status(ctx, m.ClientURL())
 	cancel()
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	ctx, cancel = context.WithTimeout(context.Background(), constants.DefaultSnapshotTimeout)
 	rc, err := etcdcli.Maintenance.Snapshot(ctx)
 	if err != nil {
-		return fmt.Errorf("failed to receive snapshot (%v)", err)
+		return nil, fmt.Errorf("failed to receive snapshot (%v)", err)
 	}
 	defer cancel()
 	defer rc.Close()
 
-	n, err := b.be.Save(resp.Version, rev, rc)
+	n, err := bm.be.Save(resp.Version, rev, rc)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	bs := backupapi.BackupStatus{
+	bs := &backupapi.BackupStatus{
 		CreationTime:     time.Now().Format(time.RFC3339),
 		Size:             util.ToMB(n),
 		Version:          resp.Version,
 		Revision:         rev,
 		TimeTookInSecond: int(time.Since(start).Seconds() + 1),
 	}
-	b.recentBackupsStatus = append(b.recentBackupsStatus, bs)
-	if len(b.recentBackupsStatus) > maxRecentBackupStatusCount {
-		b.recentBackupsStatus = b.recentBackupsStatus[1:]
-	}
 
-	return nil
+	return bs, nil
 }
 
-func getMemberWithMaxRev(pods []*v1.Pod, tc *tls.Config, selfHosted bool) (*etcdutil.Member, int64) {
+func getMemberWithMaxRev(pods []*v1.Pod, tc *tls.Config) (*etcdutil.Member, int64) {
 	var member *etcdutil.Member
 	maxRev := int64(0)
 	for _, pod := range pods {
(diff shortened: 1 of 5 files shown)


fanminshi:
cc/ @hongchaodeng @hasbro17

pkg/etcd-backup-operator/controller/sync.go
```diff
@@ -90,3 +86,13 @@ func (b *Backup) handleErr(err error, key interface{}) {
 	// Report that, even after several retries, we could not successfully process this key
 	b.logger.Infof("Dropping Backup (%v) out of the queue: %v", key, err)
 }
+
+func (b *Backup) handleBackup(spec *api.EtcdBackupSpec) error {
```
hongchaodeng: Don't include this

pkg/cluster/backupstorage/s3.go
```diff
@@ -37,7 +37,7 @@ func NewS3Storage(kubecli kubernetes.Interface, clusterName, ns string, p api.Ba
 		if err := os.MkdirAll(dir, 0700); err != nil {
 			return nil, err
 		}
-		options, err := setupAWSConfig(kubecli, ns, p.S3.AWSSecret, dir)
+		options, err := SetupAWSConfig(kubecli, ns, p.S3.AWSSecret, dir)
```
hongchaodeng: Don't include change in this file in this PR.

pkg/backup/backup.go
```diff
@@ -174,94 +176,168 @@ func (b *Backup) Run() {
 	}
 }
 
+// saveSnap saves the latest snapshot if the given current revision is greater lastSnapRev.
```
hongchaodeng: if latest revision is greater than the given lastSnapRev.

pkg/backup/backup.go
```diff
@@ -174,94 +176,168 @@ func (b *Backup) Run() {
...
-		logrus.Warning(msg)
-		return lastSnapRev, fmt.Errorf(msg)
-	}
-	member, rev := getMemberWithMaxRev(pods, b.etcdTLSConfig, b.selfHosted)
-	if member == nil {
-		logrus.Warning("no reachable member")
-		return lastSnapRev, fmt.Errorf("no reachable member")
+	ctx, cancel := context.WithTimeout(context.Background(), constants.DefaultRequestTimeout)
+	snap, etcdVersion, rev, err := GetSnap(ctx, b.kclient, b.etcdTLSConfig, b.namespace, b.clusterName)
+	defer cancel()
```
hongchaodeng: Why defer? Just cancel here?

pkg/backup/backup.go
```diff
@@ -174,94 +176,168 @@ func (b *Backup) Run() {
...
+	if err != nil {
+		return lastSnapRev, fmt.Errorf("failed to get snapshot: (%v)", err)
 	}
+	defer snap.Close()
```
hongchaodeng: Why defer? Why not just close after WriteSnap()?

pkg/backup/backup.go
```diff
@@ -174,94 +176,168 @@ func (b *Backup) Run() {
...
+	defer rc.Close()
+	n, err := be.Save(version, rev, rc)
+	if err != nil {
+		return 0, err
+	}
+	return n, nil
+}
+
+// GetSnap gets the snapshot from the cluster member with the maximum revision.
+func GetSnap(ctx context.Context, kclient kubernetes.Interface, tls *tls.Config, namespace, clusterName string) (*Snap, string, int64, error) {
```
hongchaodeng: Why pass in a ctx here?

pkg/backup/backup.go
```diff
@@ -174,94 +176,168 @@ func (b *Backup) Run() {
...
+		b.recentBackupsStatus = b.recentBackupsStatus[1:]
 	}
+
 	return rev, nil
 }
 
-func (b *Backup) writeSnap(m *etcdutil.Member, rev int64) error {
-	start := time.Now()
+// WriteSnap writes the snapshot to the backend.
+func WriteSnap(ctx context.Context, version string, rev int64, be backend.Backend, snap *Snap) (int64, error) {
```
hongchaodeng: Why pass in a ctx here?

pkg/backup/backup.go
```diff
@@ -174,94 +176,168 @@ func (b *Backup) Run() {
...
-	ctx, cancel = context.WithTimeout(context.Background(), constants.DefaultSnapshotTimeout)
-	rc, err := etcdcli.Maintenance.Snapshot(ctx)
+func NewSnap(cfg clientv3.Config) *Snap {
+	return &Snap{cfg: cfg}
+}
+
+type Snap struct {
+	cfg clientv3.Config
+	// mu protects cli
+	mu  sync.Mutex
```
hongchaodeng: > // mu protects cli

Doesn't make sense. We don't need to protect cli in this code.

pkg/backup/backup.go
```diff
@@ -174,94 +176,168 @@ func (b *Backup) Run() {
...
+	return NewSnap(cfg), resp.Version, rev, nil
+}
 
-	ctx, cancel = context.WithTimeout(context.Background(), constants.DefaultSnapshotTimeout)
-	rc, err := etcdcli.Maintenance.Snapshot(ctx)
+func NewSnap(cfg clientv3.Config) *Snap {
+	return &Snap{cfg: cfg}
+}
+
+type Snap struct {
```
hongchaodeng: This thing shouldn't call "Snap".
It should be SnapGetter or something.

hongchaodeng: If it's changed, same problem for GetSnap() and WriteSnap().

fanminshi:
All fixed. PTAL cc/ @hongchaodeng

pkg/backup/backup.go
```diff
@@ -174,94 +175,188 @@ func (b *Backup) Run() {
...
+
 	return rev, nil
 }
 
-func (b *Backup) writeSnap(m *etcdutil.Member, rev int64) error {
-	start := time.Now()
+// WriteSnap writes the snapshot to the backend.
+func WriteSnap(version string, rev int64, be backend.Backend, snap *Snap) (int64, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), constants.DefaultSnapshotTimeout)
+	defer cancel()
```
fanminshi: i used defer here so that `cancel` is called when both `rc, err := snap.Snapshot(ctx)` error out and `n, err := be.Save(version, rev, rc)` error out.

pkg/backup/backup.go
```diff
@@ -174,94 +175,188 @@ func (b *Backup) Run() {
...
 func (b *Backup) saveSnap(lastSnapRev int64) (int64, error) {
-	podList, err := b.kclient.Core().Pods(b.namespace).List(k8sutil.ClusterListOpt(b.clusterName))
+	start := time.Now()
+
+	snap, rev, err := GetSnap(b.kclient, b.etcdTLSConfig, b.namespace, b.clusterName)
 	if err != nil {
-		return lastSnapRev, err
+		return lastSnapRev, fmt.Errorf("failed to get snapshot: (%v)", err)
 	}
+	defer snap.Close()
```
fanminshi: call `defer snap.Close()` here because I need to close snap when `if rev <= lastSnapRev {` is true, `etcdVersion, err := snap.Version()` errors out, and `n, err := WriteSnap(etcdVersion, rev, b.be, snap)` errors out.

pkg/backup/backup.go
```diff
@@ -174,94 +175,186 @@ func (b *Backup) Run() {
...
 	if err != nil {
-		return fmt.Errorf("failed to create etcd client (%v)", err)
+		return nil, fmt.Errorf("failed to get member snapshot (%v)", err)
 	}
-	defer etcdcli.Close()
+	return rc, nil
+}
 
+// Version returns the etcd version of the etcd member.
+func (s *Snap) Version() (string, error) {
```
xiang90: this method does not make sense. why snapshot.version return member's version not snapshot's version? it seems too confusing.

pkg/backup/backup.go
```diff
@@ -174,94 +175,186 @@ func (b *Backup) Run() {
...
+		return "", fmt.Errorf("failed to get member status (%v)", err)
 	}
+	return resp.Version, nil
+}
 
-	ctx, cancel = context.WithTimeout(context.Background(), constants.DefaultSnapshotTimeout)
-	rc, err := etcdcli.Maintenance.Snapshot(ctx)
-	if err != nil {
-		return fmt.Errorf("failed to receive snapshot (%v)", err)
+func (s *Snap) lazyClient() (*clientv3.Client, error) {
```
xiang90: do not expose implementation detail through method naming. 

also why client is a method on snap?

pkg/backup/backup.go
```diff
@@ -174,94 +175,186 @@ func (b *Backup) Run() {
...
+		Endpoints:   []string{url},
 		DialTimeout: constants.DefaultDialTimeout,
-		TLS:         b.etcdTLSConfig,
+		TLS:         tls,
 	}
-	etcdcli, err := clientv3.New(cfg)
+	return &Snap{cfg: cfg}
+}
+
+type Snap struct {
```
xiang90: what does snap represent for? needs doc string

pkg/backup/backup.go
```diff
@@ -174,94 +175,186 @@ func (b *Backup) Run() {
...
+
+	m, rev := getMemberWithMaxRev(pods, tls)
+	if m == nil {
+		return nil, 0, fmt.Errorf("no reachable member")
+	}
+
+	return NewSnap(m.ClientURL(), tls), rev, nil
+}
 
+func NewSnap(url string, tls *tls.Config) *Snap {
```
xiang90: doc string

pkg/backup/backup.go
```diff
@@ -174,94 +175,186 @@ func (b *Backup) Run() {
...
+	n, err := be.Save(version, rev, rc)
+	rc.Close()
+	if err != nil {
+		return 0, err
+	}
+	return n, nil
+}
+
+// GetSnap gets the snapshot from the cluster member with the maximum revision.
+func GetSnap(kclient kubernetes.Interface, tls *tls.Config, namespace, clusterName string) (*Snap, int64, error) {
```
xiang90: does this func really return a snapshot?

pkg/backup/backup.go
```diff
@@ -174,94 +175,186 @@ func (b *Backup) Run() {
 	}
 }
 
+// saveSnap saves the latest snapshot if latest revision is greater than the given lastSnapRev.
 func (b *Backup) saveSnap(lastSnapRev int64) (int64, error) {
```
xiang90: probably should rename Backup to BackupManager to be more descriptive during this refactor.

pkg/backup/backup.go
fanminshi: Maybe change method name to `EtcdVersion()` or remove`Version()` in side snap struct?

xiang90: probably https://github.com/coreos/etcd-operator/pull/1489#discussion_r143321802 needs to be answered first.

pkg/backup/backup.go
fanminshi: `lazyClient()` is use to create client lazily and is used in both `Version()` and `Snapshot()`. So I refactor the common client initialization code as `lazyClient()`.

pkg/backup/backup.go
fanminshi: not anymore. will change the doc string

pkg/backup/backup.go
xiang90: i do not think the snap struct makes a lot of sense to me at the first place.

things are getting too confusing here.

first backup is the controller that manages backups. to make it less confusing, the actual backup is called snapshot.

now we are introducing two similar things again: snap and snapshot.

a backup can get a snap to get snapshots... it is just getting super confusing.

pkg/backup/backup.go
fanminshi: kk.

xiang90:
@fanminshi 

Can you write down the motivation of the refactor? What do we want to achieve? Why do we want to achieve? It helps me to understand the direction of this PR. thanks.

fanminshi:
@xiang90 kk, I'll write the motivation in the commit msg.

pkg/backup/backup.go
fanminshi: Snap provides an abstraction of getting snapshot from a member. 
 so calling snap.snapshot() returns the snapshot. calling snap.close() releases resources which is the etcd client.

xiang90: Then it should be named as SnapshotGetter. I am not sure if this abstraction is really necessary though.

fanminshi:
@xiang90 I added motivation in the commit msg section. let me know if you want me to be more detailed.

pkg/backup/backup.go
fanminshi: let me rethink about this.

xiang90:
> They needs two functionalities:
> get snapshot from etcd cluster
> save snapshot to the backend.

Basically, previously we couple the controlling logic and snapshot retrieving/storing logic together into a struct called backup.

We should do here are:

1. rename backup to backupController to make it clear that it is responsible for the controlling logic (a run func to drive the for loop)

2. move everything else to backupManager struct.

it should look like

```go

func (bc *BackupController) Run() {
	go bc.startHTTP()

	lastSnapRev := bc.backupManager.getLatestBackupRev()
	interval := constants.DefaultSnapshotInterval
	if bc.policy.BackupIntervalInSecond != 0 {
		interval = time.Duration(b.policy.BackupIntervalInSecond) * time.Second
	}

	go func() {
		if bc.policy.MaxBackups == 0 {
			return
		}
		for {
			<-time.After(10 * time.Second)
			err := bc.backupManager.purge(b.policy.MaxBackups)
			if err != nil {
				logrus.Errorf("fail to purge backups: %v", err)
			}
		}
	}()

	for {
		var ackchan chan backupNowAck
		select {
		case <-time.After(interval):
		case ackchan = <-b.backupNow:
			logrus.Info("received a backup request")
		}

		rev, err := bc.backupManager.saveSnap(lastSnapRev)
		if err != nil {
			logrus.Errorf("failed to save snapshot: %v", err)
		}
		lastSnapRev = rev

		if ackchan != nil {
			ack := backupNowAck{err: err}
			if err == nil {
				ack.status = b.backupManager.getLatestBackupStatus()
			}
			ackchan <- ack
		}
	}
}
```

fanminshi:
@xiang90 I get rid of GetSnap() and WriteSnap() and collapse both into SaveSnap() function where  SaveSnap() find snapshot and save it to the backend.

I also refactor the  backup -> backup controller and backup manager where backup controller when to backup and  backup manager  manages how to backup.

pkg/backup/backup.go
```diff
@@ -167,101 +174,128 @@ func (b *Backup) Run() {
...
 	}
+	return rev
+}
+
+func (b *BackupManager) getLatestBackupStatus() backupapi.BackupStatus {
+	return b.recentBackupsStatus[len(b.recentBackupsStatus)-1]
+}
 
-	return nil
+func getRunningPods(kclient kubernetes.Interface, namespace, clusterName string) ([]*v1.Pod, error) {
```
xiang90: revert this change? is this relevant to what we want to achieve?

pkg/backup/backup.go
```diff
@@ -298,23 +332,3 @@ func getMemberWithMaxRev(pods []*v1.Pod, tc *tls.Config, selfHosted bool) (*etcd
 	}
 	return member, maxRev
 }
-
-func (b *Backup) getLatestBackupRev() int64 {
```
xiang90: do not move this around. just change the receiver to backupmanager to ensure minimal change.

pkg/backup/backup.go
```diff
@@ -298,23 +332,3 @@ func getMemberWithMaxRev(pods []*v1.Pod, tc *tls.Config, selfHosted bool) (*etcd
...
-		return 0
-	}
-	rev, err := util.GetRev(name)
-	if err != nil {
-		logrus.Fatal(err)
-	}
-	return rev
-}
-
-func (b *Backup) getLatestBackupStatus() backupapi.BackupStatus {
```
xiang90: same here.

xiang90:
we almost get there. thanks.

pkg/backup/backup.go
fanminshi: not anymore. will revert.

pkg/backup/backup.go
```diff
@@ -46,25 +45,30 @@ const (
 	maxRecentBackupStatusCount = 10
 )
 
-type Backup struct {
+// BackupManager manages how to backup.
```
xiang90: what does how mean? need to be more descriptive.

pkg/backup/backup.go
```diff
@@ -114,36 +118,39 @@ func New(kclient kubernetes.Interface, clusterName, ns string, sp api.ClusterSpe
...
+		listenAddr:    listenAddr,
+		backupNow:     make(chan chan backupNowAck),
+		policy:        *sp.Backup,
+		backupManager: bm,
 	}, nil
 }
 
-func (b *Backup) Run() {
-	go b.startHTTP()
+func (bc *BackupController) Run() {
```
xiang90: doc string on run

pkg/backup/backup.go
```diff
@@ -167,101 +174,128 @@ func (b *Backup) Run() {
...
-	log.Printf("saving backup for cluster (%s)", b.clusterName)
-	if err := b.writeSnap(member, rev); err != nil {
-		err = fmt.Errorf("write snapshot failed: %v", err)
-		return lastSnapRev, err
+		return 0, 0, "", fmt.Errorf("skipped creating new backup: no change since last time")
 	}
-	return rev, nil
-}
-
-func (b *Backup) writeSnap(m *etcdutil.Member, rev int64) error {
```
xiang90: why this func is removed? only the receiver needs to be changed to backupManager to achieve our goal, no?

pkg/backup/backup.go
```diff
@@ -46,25 +45,30 @@ const (
...
-func New(kclient kubernetes.Interface, clusterName, ns string, sp api.ClusterSpec, listenAddr string) (*Backup, error) {
+// BackupController controls when to backup.
+type BackupController struct {
+	listenAddr    string
+	backupNow     chan chan backupNowAck
+	policy        api.BackupPolicy
+	backupManager *BackupManager
+}
+
+func New(kclient kubernetes.Interface, clusterName, ns string, sp api.ClusterSpec, listenAddr string) (*BackupController, error) {
```
xiang90: NewBackupController

pkg/backup/backup.go
```diff
@@ -167,101 +174,128 @@ func (b *Backup) Run() {
...
+	b.recentBackupsStatus = append(b.recentBackupsStatus, bs)
+	if len(b.recentBackupsStatus) > maxRecentBackupStatusCount {
+		b.recentBackupsStatus = b.recentBackupsStatus[1:]
+	}
+
+	return rev, nil
+}
 
+// SaveSnap saves snapshot to backend from the etcd member whose rev is greater than lastSnapRev.
+func SaveSnap(kclient kubernetes.Interface, tls *tls.Config, lastSnapRev int64, be backend.Backend, namespace, clusterName string) (int64, int64, string, error) {
```
xiang90: why do we need to create this func there? we should expose backupManager and its methods.

pkg/backup/backup.go
fanminshi: `BackupManager manages where to get etcd backups from and where save them`?

pkg/backup/backup.go
xiang90: /cc @fanminshi  this needs to be resolved

pkg/backup/backup.go
xiang90: /cc @fanminshi this needs to be resolved.

fanminshi: kk, let me see what I can do.

pkg/backup/backup.go
fanminshi: `only the receiver needs to be changed to backupManager to achieve our goal, no?`
possibly. let me think about how to use `backupManager` in the operator controller code.

pkg/backup/backup.go
```diff
@@ -154,30 +162,38 @@ func (b *Backup) Run() {
...
+				ack.status = bc.recentBackupsStatus[len(bc.recentBackupsStatus)-1]
 			}
 			ackchan <- ack
 		}
 	}
 }
 
-func (b *Backup) saveSnap(lastSnapRev int64) (int64, error) {
-	podList, err := b.kclient.Core().Pods(b.namespace).List(k8sutil.ClusterListOpt(b.clusterName))
+// SaveSnap saves the snapshot if snapshot's revision is greater than the given lastSnapRev.
```
xiang90: explain the return value.

xiang90: also does backupstate contain rev already?

xiang90:
lgtm after fixing the two minor issues.

pkg/backup/backup.go
fanminshi: will do what you suggested.

pkg/backup/backup.go
```diff
@@ -154,30 +162,39 @@ func (b *Backup) Run() {
...
 
-		rev, err := b.saveSnap(lastSnapRev)
+		bs, err := bc.backupManager.SaveSnap(lastSnapRev)
 		if err != nil {
 			logrus.Errorf("failed to save snapshot: %v", err)
 		}
-		lastSnapRev = rev
+
+		if bs != nil {
+			lastSnapRev = bs.Revision
```
fanminshi: @xiang90 I changed `SaveSnap` to return `BackupStatus` and `error`.
And only updates lastSnapRev if `BackupStatus` is not nil.

pkg/backup/backup.go
```diff
@@ -191,77 +208,74 @@ func (b *Backup) saveSnap(lastSnapRev int64) (int64, error) {
...
 	if member == nil {
 		logrus.Warning("no reachable member")
-		return lastSnapRev, fmt.Errorf("no reachable member")
+		return nil, fmt.Errorf("no reachable member")
 	}
 
 	if rev <= lastSnapRev {
 		logrus.Info("skipped creating new backup: no change since last time")
-		return lastSnapRev, nil
+		return nil, nil
```
fanminshi: @xiang90 I am unsure what to return here. I return `nil,nil` because no backup is performed and skipping is not considered an error by judgement of previous code.

xiang90:
lgtm

fanminshi:
merge when green.

PR #1492, opened 2017-10-09
*: expose setupAWSConfig and construct BackupManager with NewBackupManager

fanminshi:
This pr exposes SetupAWSConfig in backupstorage/s3.go so that etcd backup operator can use it to set up aws credentials.

In addition, it creates NewBackupManager to construct BackupManager so that BackupManager's private fields don't have to be exposed.


fanminshi:
cc/ @xiang90

hongchaodeng:
lgtm

PR #1512, opened 2017-10-12
*: change dir named with etcd-restore-operator to restore-operator

fanminshi:
change all dir named with etcd-restore-operator to restore-operator.


fanminshi:
cc/ @hongchaodeng @hasbro17

hack/build/restoreoperator/build
```diff
@@ -6,7 +6,7 @@ set -o pipefail
 
 function go_build {
 	echo "building "${1}"..."
-	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o ${bin_dir}/${1} -installsuffix cgo -ldflags "$go_ldflags" ./cmd/${1}/
+	GOOS=darwin GOARCH=amd64 CGO_ENABLED=0 go build -o ${bin_dir}/etcd-restore-operator -installsuffix cgo -ldflags "$go_ldflags" ./cmd/${1}/
```
hongchaodeng: Should be `GOOS=linux` ?

cmd/restoreoperator/main.go
```diff
@@ -22,7 +22,7 @@ import (
 
 	"github.com/coreos/etcd-operator/pkg/controller/restorecontroller"
 	"github.com/coreos/etcd-operator/pkg/util/k8sutil"
-	version "github.com/coreos/etcd-operator/version/etcd-restore-operator"
+	version "github.com/coreos/etcd-operator/version/restoreoperator"
```
xiang90: restore-operator is probably cleaner.

hack/build/restoreoperator/build
fanminshi: good catch.

cmd/restoreoperator/main.go
hongchaodeng: So some dirs are "restoreoperator" but some are "restore-operator".
Isn't it better to unify them?
I have no preference on either.

fanminshi:
all fixed. PTAL

hongchaodeng:
lgtm

PR #1519, opened 2017-10-13
pkg/controller: move etcd operator controller files to pkg/controller/operator

fanminshi:


fanminshi:
discuss in private. we should keep the previous etcd operator code structure untouched.

PR #1523, opened 2017-10-13
*: expose startHTTP

fanminshi:
Separate controller.Run() from startHTTP() allows backup to have a choice
on what to run and not to run.

This is important because backup will be adding a flag to enable only
serve backup http requests.


fanminshi:
cc/ @hongchaodeng @hasbro17

PR #1530, opened 2017-10-17
*: ignore event with invalid cluster spec at the controller

hasbro17:
~The backup policy should be validated not only when the cluster spec is created but also when the cluster spec's backup policy is updated.~

The cluster spec is now validated in the controller. Any event with an invalid spec is ignored and logged.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -134,11 +134,7 @@ func New(config Config, cl *api.EtcdCluster) *Cluster {
 }
 
 func (c *Cluster) setup() error {
-	err := c.cluster.Spec.Validate()
-	if err != nil {
-		return fmt.Errorf("invalid cluster spec: %v", err)
-	}
-
+	var err error
 	var shouldCreateCluster bool
 	switch c.status.Phase {
 	case api.ClusterPhaseNone:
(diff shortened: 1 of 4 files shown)


fanminshi:
lgtm defer to @hongchaodeng

pkg/cluster/cluster.go
```diff
@@ -369,6 +369,11 @@ func (c *Cluster) handleUpdateEvent(event *clusterEvent) error {
 	c.logSpecUpdate(*oldSpec, event.cluster.Spec)
 
 	ob, nb := oldSpec.Backup, event.cluster.Spec.Backup
+	if nb != nil {
+		if err := nb.Validate(); err != nil {
+			return fmt.Errorf("Invalid backup policy: %v", err)
```
xiang90: invalid backup policy

hasbro17:
After offline discussion I'm going to move the cluster spec validation out from the cluster level to the controller at:
https://github.com/coreos/etcd-operator/blob/master/pkg/controller/controller.go#L95

With that, any event with an invalid cluster spec, either Add or Modify will be ignored and logged.

hasbro17:
@hongchaodeng PTAL

hongchaodeng:
lgtm

Issue #1527, opened 2017-10-16
etcd-operator should require Role instead of ClusterRole 


hasbro17:
According to the [RBAC guide](https://github.com/coreos/etcd-operator/blob/master/doc/user/rbac.md#create-clusterrole) the etcd-operator requires a ClusterRole instead of a Role because it needs access to the following non-namespaced resources:
- `apiextensions.k8s.io/customresourcedefinitions`
- `storage.k8s.io/storageclasses`

### Motivation to use a Role instead of ClusterRole
When a user creates the ClusterRole required by the etcd-operator it is assumed that user has permissions at the cluster-wide scope for the above-mentioned resources.

As the [Kubernetes documentation](https://kubernetes.io/docs/tasks/administer-cluster/namespaces/#subdividing-your-cluster-using-kubernetes-namespaces) suggests this may not always be the case as a cluster could be subdivided by namespaces where a user is only authorized to access resources within a single namespace. So such a user will not be able to create the current ClusterRole and ClusterRoleBinding needed by the etcd-operator.

It is reasonable to expect that the user has sufficient permissions to create a Role and Rolebinding for the resources that are scoped to their namespace.

### Steps to use a Role instead of ClusterRole
To remove the dependency of the etcd-operator on ClusterRole permissions the resources `CustomResourceDefinition` and `StorageClass` will need to be managed outside the operator.

The etcd-operator will need to be modified to no longer create the EtcdCluster CRD on startup.
Additionally, the `--pv-provisioner` flag will need to be deprecated and the etcd-operator will no longer be responsible for creating the storage class and instead use the user-specified storage class.

The assumption is that an admin with the necessary cluster-wide permissions will create the following two resources before a user can create the etcd-operator in their namespace.

#### CustomResourceDefinition
```YAML
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: etcdclusters.etcd.database.coreos.com
spec:
  group: etcd.database.coreos.com
  scope: Namespaced
  version: v1beta2
  names:
    kind: EtcdCluster
    listKind: EtcdClusterList
    singular: etcdcluster
    plural: etcdclusters
    shortNames:
    - etcd
```
 
#### StorageClass

```YAML
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: <storage-class-name>
provisioner: <pv-provisioner>
```
[...]

PR #1533, opened 2017-10-17
*: add flags to not create global resources CRD and StorageClass

hasbro17:
Addresses #1527 

The default behavior of the etcd-operator is now to not create global resources (CRD, StorageClass). This is governed by the flag `no-global-cluster-resource` which is true by default.

In this default mode the `--pv-provisioner` flag will have no effect. Any backup policy of type `PV` without a `StorageClass` field will be invalid because the operator will not create any StorageClass resource.

To revert back to the old behavior of the operator the flag `no-global-cluster-resource` needs to be set to false.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -59,15 +59,22 @@ var (
 	chaosLevel int
 
 	printVersion bool
+
+	createCRD          bool
+	createStorageClass bool
 )
 
 func init() {
 	flag.StringVar(&debug.DebugFilePath, "debug-logfile-path", "", "only for a self hosted cluster, the path where the debug logfile will be written, recommended to be under: /var/tmp/etcd-operator/debug/ to avoid any issue with lack of write permissions")
-	flag.StringVar(&pvProvisioner, "pv-provisioner", constants.PVProvisionerGCEPD, "persistent volume provisioner type")
+	flag.StringVar(&pvProvisioner, "pv-provisioner", constants.PVProvisionerGCEPD, "persistent volume provisioner type. DEPRECATION WARNING: the pv-provisioner flag will be removed in next release.")
 	flag.StringVar(&listenAddr, "listen-addr", "0.0.0.0:8080", "The address on which the HTTP server will listen to")
 	// chaos level will be removed once we have a formal tool to inject failures.
 	flag.IntVar(&chaosLevel, "chaos-level", -1, "DO NOT USE IN PRODUCTION - level of chaos injected into the etcd clusters created by the operator.")
 	flag.BoolVar(&printVersion, "version", false, "Show version and quit")
+	// create-crd will be removed once the default behavior is not to create a CRD
+	flag.BoolVar(&createCRD, "create-crd", false, "The operator will not create the EtcdCluster CRD when this flag is set to false.")
+	// create-storage-class will be removed once the default behavior is not to create storage class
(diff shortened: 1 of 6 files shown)


hasbro17:
The tests should fail. Will fix the tests according to the new flag in a separate commit.

cmd/operator/main.go
```diff
@@ -68,6 +68,8 @@ func init() {
 	// chaos level will be removed once we have a formal tool to inject failures.
 	flag.IntVar(&chaosLevel, "chaos-level", -1, "DO NOT USE IN PRODUCTION - level of chaos injected into the etcd clusters created by the operator.")
 	flag.BoolVar(&printVersion, "version", false, "Show version and quit")
+	// no-global-cluster-resource will be removed once the default behavior to not create global cluster resources is enforced
+	flag.BoolVar(&constants.NoGlobalClusterResource, "no-global-cluster-resource", true, "This operator will not create global resources(CRD, StorageClass) when set to true. This flag needs to be false before the --pv-provisioner flag can be used.")
```
xiang90: this flag is not descriptive. right now, it is pretty much short-term task driven. it is hard for users who have no context to understand why this flag exists.

we need to make the flag more descriptive, and maybe separate the storageclass and crd to two different flags.

hasbro17: Going to change it to two flags to make it clearer.

cmd/operator/main.go
```diff
@@ -59,15 +59,22 @@ var (
...
 	printVersion bool
+
+	createCRD          bool
+	createStorageClass bool
 )
 
 func init() {
 	flag.StringVar(&debug.DebugFilePath, "debug-logfile-path", "", "only for a self hosted cluster, the path where the debug logfile will be written, recommended to be under: /var/tmp/etcd-operator/debug/ to avoid any issue with lack of write permissions")
-	flag.StringVar(&pvProvisioner, "pv-provisioner", constants.PVProvisionerGCEPD, "persistent volume provisioner type")
+	flag.StringVar(&pvProvisioner, "pv-provisioner", constants.PVProvisionerGCEPD, "persistent volume provisioner type. This flag will be deprecated in the next release")
```
hongchaodeng: DEPRECATION WARNING: pv-provisioner will be removed in next release.

pkg/apis/etcd/v1beta2/backup.go
```diff
@@ -67,6 +67,8 @@ func (bp *BackupPolicy) Validate() error {
 		if pv := bp.StorageSource.PV; pv == nil || pv.VolumeSizeInMB <= 0 {
 			return errPVZeroSize
 		}
+
```
hongchaodeng: remove blank line

test/e2e/framework/framework.go
```diff
@@ -103,7 +103,7 @@ func (f *Framework) setup() error {
 
 func (f *Framework) SetupEtcdOperator() error {
 	// TODO: unify this and the yaml file in example/
-	cmd := []string{"/usr/local/bin/etcd-operator"}
+	cmd := []string{"/bin/sh", "-ec", "/usr/local/bin/etcd-operator --create-crd=true --create-storage-class=true"}
```
hongchaodeng:
```
{"/usr/local/bin/etcd-operator", "...", ...
```

hongchaodeng:
lgtm

PR #1535, opened 2017-10-17
*: add serving HTTP backup request to restore operator

fanminshi:
enables restore operator to listen on port 19999 for serving HTTP backup request.
the request must be in form of endpoint:19999/v1/backup/etcd_cluster_name>/?etcdVersion=<etcd_version>


pkg/controller/restore-operator/backup_handler.go
```diff
@@ -0,0 +1,95 @@
+// Copyright 2017 The etcd-operator Authors
```
hongchaodeng: file name -> "http_handler.go" ?

xiang90:
cluster name should be part of  the resource path. it is not a query parameter. 

the path should look like /backups/clusterName/xxx

pkg/controller/restore-operator/backup_handler.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// close removes aws credential/config dir and backupDir dir which were
+// created for BackupServer.
+func (bsc *backupServerCloser) close() {
+	os.RemoveAll(bsc.awsDir)
+	os.RemoveAll(bsc.backupDir)
+}
+
+func (r *Restore) handleBackup(spec *api.BackupSpec) error {
```
hongchaodeng: handleBackup  -> handleRestore?

pkg/controller/restore-operator/backup_handler.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/coreos/etcd-operator/pkg/backup/backend"
+	"github.com/coreos/etcd-operator/pkg/backup/backupapi"
+	backupS3 "github.com/coreos/etcd-operator/pkg/backup/s3"
+	"github.com/coreos/etcd-operator/pkg/cluster/backupstorage"
+	"github.com/sirupsen/logrus"
+)
+
+// backupServerCloser wraps around backup.BackupServer with a close method
+// that cleans up resources associated with backup.BackupServer.
+type backupServerCloser struct {
```
xiang90: just backupServer. closer has no meaning here. no one will use it externally as an interface.

pkg/controller/restore-operator/operator.go
```diff
@@ -59,6 +67,38 @@ func New() *Restore {
...
+	clusterName := req.URL.Query().Get("cluster")
+	if len(clusterName) == 0 {
+		http.Error(w, "cluster is not specified", http.StatusBadRequest)
+		return
+	}
+	r.mu.RLock()
+	bs, ok := r.backupServers[clusterName]
+	r.mu.RUnlock()
+	if !ok {
+		http.Error(w, fmt.Sprintf("cluster %v backup server not found", clusterName), http.StatusInternalServerError)
```
xiang90: return here

pkg/controller/restore-operator/operator.go
```diff
@@ -42,6 +46,10 @@ type Restore struct {
 	kubecli       kubernetes.Interface
 	restoreCRCli  versioned.Interface
 	kubeExtClient apiextensionsclient.Interface
+
+	// mu protects backupServers.
+	mu            sync.RWMutex
+	backupServers map[string]*backupServerCloser
```
hongchaodeng: use concurrent map:
https://golang.org/doc/go1.9#sync-map

pkg/controller/restore-operator/sync.go
```diff
@@ -28,6 +24,7 @@ const (
 	//
 	// 5ms, 10ms, 20ms, 40ms, 80ms, 160ms, 320ms, 640ms, 1.3s, 2.6s, 5.1s, 10.2s, 20.4s, 41s, 82s
 	maxRetries = 15
+	tmpdir     = "/tmp"
```
hongchaodeng: What's this var for?
Isn't it duplicate?

fanminshi:
@xiang90 I am thinking about calling the backup http api like 
`curl -o <etcd_snap_name>  restore_operator_endpoint:19999/backup?cluster=example-cluster,etcdVersion=3.1.8`.
 I am not sure you meant by 
```
cluster name should be part of the resource path.
the path should look like /backups/clusterName/xxx
```
could you elaborate?

pkg/controller/restore-operator/backup_handler.go
```diff
@@ -0,0 +1,97 @@
...
+		logrus.Fatalf("unknown StorageType: %v", spec.StorageType)
+	}
+	r.mu.Lock()
+	r.backupServers[spec.ClusterName] = bsc
+	r.mu.Unlock()
+	return nil
+}
+
+func createS3BackupServer(kubecli kubernetes.Interface, s3 *api.S3Source, namespace, clusterName string) (*backupServerCloser, error) {
+	awsDir, err := ioutil.TempDir(tmpdir, "")
```
hongchaodeng: Duplicate to backup-operator/s3_backup.go ?
WIP Anyway. Will look again at final.

pkg/controller/restore-operator/sync.go
fanminshi: yeah, it is duplicated from backup-operator.

hongchaodeng:
> cluster name should be part of the resource path.
the path should look like /backups/clusterName/xxx

@fanminshi Just use endpoint:
```
/backups/<cluster-name>
```
For example, if cluster-name is "a", then it will query
```
/backups/a
```

pkg/controller/restore-operator/sync.go
hongchaodeng: I mean duplicate in restore-operator/backup_handler.go too

hongchaodeng:
I have written a prototype for the backup server in restore operator that does the cluster name path parsing. Please base on that:

branch: https://github.com/coreos/etcd-operator/tree/res_h

commit: https://github.com/coreos/etcd-operator/commit/bd3a6e9fadf62d42eef35225f506a34b126911b6

pkg/controller/restore-operator/s3_backup.go
```diff
@@ -0,0 +1,95 @@
...
+// that cleans up resources associated with backup.BackupServer.
+type backupServer struct {
+	*backup.BackupServer
+	awsDir    string
+	backupDir string
+}
+
+// close removes aws credential/config dir and backupDir dir which were
+// created for BackupServer.
+func (bsc *backupServer) close() {
```
xiang90: probably just remove.

fanminshi: the method is called `close` to indicate that this `backupServer` is finished and all its related resources are released.

hongchaodeng: I see..
@xiang90 
So the workflow here is that:
- when a Restore spec is created, it creates S3 client
- When finished serving backup, it destroys everything

The above two is concurrent, although they have implicit ordering.

I am fine with what it is now.

pkg/controller/restore-operator/http.go
```diff
@@ -0,0 +1,52 @@
...
+	if len(clusterName) == 0 {
+		http.Error(w, "cluster is not specified", http.StatusBadRequest)
+		return
+	}
+	v, ok := r.backupServers.Load(clusterName)
+	if !ok {
+		http.Error(w, fmt.Sprintf("cluster %v backup server not found", clusterName), http.StatusInternalServerError)
+		return
+	}
+	go func() {
```
hongchaodeng: Don't create go routine here?

pkg/controller/restore-operator/s3_backup.go
hongchaodeng: Actually after a second thought we shouldn't destroy it. A client might fail before and retry again.
We should keep it there for now. Let's handle cleanup later.

hongchaodeng: After offline discussion, we will save a map of existing Restore CR. Restore CR will be cleaned up once user deletes the CR.

On serveBackup(), we will pull the secret, create storage backend, pull backup, serve backup, and destroy. This eliminates concurrency issue, but might have performance issue. A tradeoff for simplicity now.

fanminshi:
cc/ @hongchaodeng I implemented the following

"On serveBackup(), we will pull the secret, create storage backend, pull backup, serve backup, and destroy. This eliminates concurrency issue, but might have performance issue. A tradeoff for simplicity now."

pkg/controller/restore-operator/http.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+func (r *Restore) startHTTP() {
+	http.HandleFunc(backupapi.APIV1+"/backup", r.serveBackup)
+	logrus.Infof("listening on %v", listenAddr)
+	panic(http.ListenAndServe(listenAddr, nil))
+}
+
+// path: /backup/<cluster-name>
+func (r *Restore) serveBackup(w http.ResponseWriter, req *http.Request) {
```
hongchaodeng: doc about what this method does?

pkg/controller/restore-operator/http.go
hongchaodeng: address this comment?

fanminshi: create go routine here can speed up multiple incoming requests.

pkg/controller/restore-operator/http.go
```diff
@@ -0,0 +1,75 @@
...
+		return
+	}
+	v, ok := r.restoreCRs.Load(clusterName)
+	if !ok {
+		http.Error(w, fmt.Sprintf("cluster %v backup server not found", clusterName), http.StatusInternalServerError)
+		return
+	}
+	// a new go routine is created to handle each request to speed up
+	// multi requests handling.
+	go func() {
```
hongchaodeng: Again, don't create a goroutine here.
The performance concern is unnecessary for now.

fanminshi: done.

hongchaodeng:
lgtm

fanminshi:
manual test:

create restore-operator:
```
kind: Deployment
metadata:
  name: etcd-restore-operator
spec:
  replicas: 1
  template:
    metadata:
      labels:
        name: etcd-restore-operator
    spec:
      containers:
      - name: etcd-restore-operator
        image: gcr.io/coreos-k8s-scale-testing/etcd-restore-operator:fanmin
        imagePullPolicy: Always
        command: 
        - /usr/local/bin/etcd-restore-operator
        env:
        - name: MY_POD_NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace
        - name: MY_POD_NAME
          valueFrom:
            fieldRef:
              fieldPath: metadata.name
```

create restore CR:
```yaml
kind: "EtcdBackup"
metadata:
  name: s3-cluster-backup
spec:
  clusterName: s3-cluster-backup
  storageType: S3
  s3:
    s3Bucket: jenkins-etcd-operator
    prefix: prefix
    awsSecret: aws
```

get snap:
```sh
curl -o snap http://172.17.0.5:19999/v1/backup/s3-cluster-backup?etcdVersion=3.1.8
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 24608    0 24608    0     0   123k      0 --:--:-- --:--:-- --:--:--  123k
```

fanminshi:
merge when green.

PR #1536, opened 2017-10-18
*: move creation of s3backend to a util pkg

fanminshi:
creation of s3backend can be shared for both backup operator and restore operator.
moving it out to a util pkg allows code reuse.


hongchaodeng:
lgtm

PR #1537, opened 2017-10-18
controller: fix CreateCRD condition check

hasbro17:
If the `create-crd` flag is false then the operator should also not be **waiting** for the crd to become ready since it does not have the rbac permissions to get a crd.

So if `create-crd` is false then this will skip the `initCRD()` altogether.


hasbro17:
@hongchaodeng PTAL

hongchaodeng:
lgtm

PR #1541, opened 2017-10-18
operator: create crd by default

hasbro17:


cmd/operator/main.go
```diff
@@ -72,7 +72,7 @@ func init() {
 	flag.IntVar(&chaosLevel, "chaos-level", -1, "DO NOT USE IN PRODUCTION - level of chaos injected into the etcd clusters created by the operator.")
 	flag.BoolVar(&printVersion, "version", false, "Show version and quit")
 	// create-crd will be removed once the default behavior is not to create a CRD
```
fanminshi: is this comment still accurate?
`// create-crd will be removed once the default behavior is not to create a CRD`

hasbro17: Ah, good catch. No this flag will now stay since the default is creating a crd.

fanminshi:
lgtm

PR #1551, opened 2017-10-20
*: update and unify rbac doc and rbac utils with rbac templates

hasbro17:
Addresses #1527 

This PR updates the RBAC setup guide to explain how to setup RBAC with either a ClusterRole or Role, depending on the `create-crd` flag.

Added RBAC templates at `example/rbac` so that both the RBAC doc and `hack/ci/rbac_utils.sh` use the same permissions.

The rbac_utils still sets up RBAC with a ClusterRole since our e2e tests test the etcd-operator for the default mode of `--create-crd=true`.

### DIFF
--- a/doc/user/rbac.md
+++ b/doc/user/rbac.md
@@ -43,133 +43,66 @@ $ kubectl delete ns $TEST_NAMESPACE
 
 ## Production setup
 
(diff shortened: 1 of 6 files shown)


hongchaodeng:
lgtm

PR #1556, opened 2017-10-23
doc: Add script to streamline RBAC setup

jpbetz:
Streamline the getting started workflow for developers using k8s 1.7+ with a single command to setup RBAC and clearly document it in the getting started docs.

### DIFF
--- a/doc/user/op_guide.md
+++ b/doc/user/op_guide.md
@@ -2,9 +2,13 @@
 
 ## Install etcd operator
 
-Before you create the etcd-operator make sure you setup the necessary [rbac rules](./rbac.md) if your Kubernetes version is 1.7 or higher.
+Set up basic [RBAC rules](./rbac.md) for etcd operator:
 
-Create deployment:
+```bash
+$ example/rbac/create_role.sh
+```
+
+Create a deployment for etcd operator:
 
 ```bash
 $ kubectl create -f example/deployment.yaml
--- a/doc/user/rbac.md
+++ b/doc/user/rbac.md
@@ -2,45 +2,6 @@
 
 If RBAC is in place, users need to setup RBAC rules for etcd operator. This doc serves a tutorial for it.
 
-## Quick setup
-
-If you just want to play with etcd operator, there is a quick setup.
-
-It assumes that your cluster has an admin role. For example, on [Tectonic](https://coreos.com/tectonic/),
-there is a `admin` ClusterRole. We are using that here.
-
-Modify or export env `$TEST_NAMESPACE` to a new namespace, then create it:
-
-```bash
-$ kubectl create ns $TEST_NAMESPACE
-```
-
-Then create cluster role binding:
-
-```bash
-$ cat <<EOF | kubectl create -f -
-apiVersion: rbac.authorization.k8s.io/v1beta1
-kind: ClusterRoleBinding
-metadata:
-  name: example-etcd-operator
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
-  name: admin
-subjects:
-- kind: ServiceAccount
-  name: default
-  namespace: $TEST_NAMESPACE
-EOF
-```
-
-Now you can start playing. One you are done, clean them up:
-
-```bash
-$ kubectl delete clusterrolebinding example-etcd-operator
-$ kubectl delete ns $TEST_NAMESPACE
-```
-
 ## Production setup
 
 For production, we recommend users to limit access to only the resources operator needs, and create a specific role, for the operator.
--- a/example/rbac/create_role.sh
+++ b/example/rbac/create_role.sh
@@ -1,0 +1,59 @@
+#!/usr/bin/env bash
+set -o errexit
+set -o nounset
+set -o pipefail
+
+ETCD_OPERATOR_ROOT=$(dirname "${BASH_SOURCE}")/../..
+
+print_usage() {
+  echo "$(basename "$0") - Create Kubernetes RBAC role and role bindings for etcd-operator
+Usage: $(basename "$0") [options...]
+Options:
+  --role-name=STRING         Name of Role or ClusterRole to create
+                               (default=\"etcd-operator\", environment variable: ROLE_NAME)
+  --role-binding-name=STRING Name of RoleBinding or ClusterRoleBinding to create
+                               (default=\"etcd-operator\", environment variable: ROLE_BINDING_NAME)
+  --namespace=STRING         namespace to create role and role binding in. Must already exist.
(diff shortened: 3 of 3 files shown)


hongchaodeng:
@jpbetz 
Last week we just re-structured rbac docs and separated out rbac yaml files:

- https://github.com/coreos/etcd-operator/tree/master/example/rbac
- https://github.com/coreos/etcd-operator/blob/master/doc/user/rbac.md

So this should have be solved?

jpbetz:
@hongchaodeng That's great. My goal here was to streamline the getting started flow. Right now, it's still quite a few steps to go through the `rbac.md` flow. Something like this can reduce it down to a single step. I'm not set on doing it this exact way, but there is a lot of value in reducing this down to to a single command, particularly for developers just trying it out for the first time.

Maybe now that the templates exist, we should also provide a script that runs them in the appropriate order automatically? If that sounds like a good approach let me know and I can update this PR to do that.

jpbetz:
I've rewritten this PR to use the rbac templates and a script to tie it all together for the getting started flow.

Is the a better place to put a script like this? It's currently in `example/rbac`, but maybe it would be better kept elsewhere?

example/rbac/create_role.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
hongchaodeng: `/usr/bin/env bash` .

Can you also provide docs what this script does and how to use it?

doc/user/op_guide.md
```diff
@@ -2,9 +2,16 @@
 
 ## Install etcd operator
 
-Before you create the etcd-operator make sure you setup the necessary [rbac rules](./rbac.md) if your Kubernetes version is 1.7 or higher.
+For Kubernetes 1.7 or higher, set up [rbac rules](./rbac.md) for an
```
hongchaodeng: This info is outdated.

> For Kubernetes 1.7 or higher

This is already mentioned in "Requirements" section in README. We should remove this.

>  set up rbac rules for an `etcd-operator` service account

This has changed. For simplicity, we use namespace default account.

The reason is that otherwise there would be many inconsistent parts in our examples, e.g. deployment.yaml. And we don't want to add to much complexity in open source etcd operator due to support capability. If a user cares about RBAC and using a different account, he needs to understand k8s RBAC and set up those himself.

example/rbac/create_role.sh
```diff
@@ -0,0 +1,32 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+ETCD_OPERATOR_ROOT=$(dirname "${BASH_SOURCE}")/../..
+
+ROLE_NAME="${ROLE_NAME:-etcd-operator}"
+ROLE_BINDING_NAME="${ROLE_BINDING_NAME:-etcd-operator}"
+NAMESPACE="${NAMESPACE:-default}"
+CREATE_CRD="${CREATE_CRD:-true}"
```
hongchaodeng: Just assume `CREATE_CRD=true` here and only handle that.

doc/user/rbac.md
```diff
@@ -4,9 +4,13 @@ If RBAC is in place, users need to setup RBAC rules for etcd operator. This doc
 
 ## Quick setup
 
-If you just want to play with etcd operator, there is a quick setup.
```
hongchaodeng: Just get rid of the "admin" setting?
Now the script is there, we should  just use it for "quick setup", right?

doc/user/op_guide.md
hongchaodeng: Maybe reduce this paragraph to only:

Setup RBAC rules for etcd-operator:
```
... run the script
```

doc/user/op_guide.md
```diff
@@ -2,9 +2,16 @@
...
+`etcd-operator` service account and deploy the operator. To get started quickly
+with a basic service account and etcd operator running in the default namespace,
+use:
 
-Create deployment:
+```bash
+$ example/rbac/create_role.sh
+```
+
+Next, create a deployment for etcd operator:
```
hongchaodeng: ~~Next,~~ Create an etcd operator deployment:

doc/user/rbac.md
```diff
@@ -69,40 +73,12 @@ export NAMESPACE=<namespace>
 
 ### RBAC with ClusterRole (create-crd=true)
 
-1. Create a ClusterRole:
```
hongchaodeng: Can you keep this?

Instead of having "Quick setup" and "Production setup", maybe change to "Quick setup" and "Detailed setup".

hongchaodeng:
After a second thought, I think we could improve the docs to avoid duplicates and be better structured as:

- Add the rbac setup script only for operator install guide. In the operator Install doc, use it to streamline the process with script.
- In the rbac.md doc, get rid of the "quick setup" entirely, and change the doc to only talk about Role vs ClusterRole setup. The "quick setup" was originally put there for easier operator install anyway. It's better off using the script. And thus we should get rid of "quick setup" section. The rbac doc should be the official guide and give all details for advanced users.

@jpbetz I understand it seems more work than you originally started with. So I'm fine if you just do the first item and leave the second. We can take care of the second job. Thanks!

jpbetz:
@hongchaodeng This sounds great. I like the idea of making a clean split between the getting started flow and the reference documentation.

doc/user/op_guide.md
jpbetz: Applied. I've also added a version check to the script so if one attempts to run it on 1.6 or earlier, it will print something like: "Kubernetes version 1.6 predates RBAC support. Skipping role creation."

jpbetz:
@hongchaodeng latest diff removes the "Quick setup" section from the RBAC reference documentation, simplifies the quick start guide like you suggested and adds documentation to the create_role.sh script.

Have a look. Additional feedback welcome.

example/rbac/create_role.sh
jpbetz: Great, that simplifies things.

example/rbac/create_role.sh
```diff
@@ -0,0 +1,67 @@
...
+      exit 0
+    ;;
+    *)
+      print_usage
+      exit 1
+    ;;
+esac
+done
+
+version_yaml=$(kubectl version --client -o yaml)
```
hongchaodeng: - This doesn't work for 1.7 kubectl.
- Can you remove this part of k8s version check? We already mention in "Requirements" section on k8s version. We don't need to duplicate this logic in every script which uses kubectl. Thanks!

jpbetz: Removed

hongchaodeng:
Thanks @jpbetz

PR #1558, opened 2017-10-23
*: update all documentation and examples on use of pv-provisioner

hasbro17:
[skip ci]

Updated the documentation wherever it is assumed that the operator will create a storage class by default.

The example backup spec `example/example-etcd-cluster-with-backup.yaml` has been changed to use the custom storage class `example/example-storage-class-gce-pd.yaml`.


hongchaodeng:
lgtm

PR #1564, opened 2017-10-24
controller/restore-operator: don't process cr if it has a status

fanminshi:
group handling restore cr into a function called handleCR
avoid processing cr twice if it has status. since updating cr's status triggers
the cr the be queued again, don't process the cr if that's the case.


fanminshi:
cc/ @hongchaodeng @hasbro17

hongchaodeng:
lgtm

PR #1565, opened 2017-10-24
controller/backup-operator: don't process cr if it has a status

fanminshi:
if cr has a status, it means that this cr has been process before.
In that case, don't process the cr again.

reportBackupStatus takes in berr instead of err to differentiate input err vs internal error.

### DIFF
--- a/pkg/controller/backup-operator/sync.go
+++ b/pkg/controller/backup-operator/sync.go
@@ -61,6 +61,11 @@ func (b *Backup) processItem(key string) error {
 	}
 
 	eb := obj.(*api.EtcdBackup)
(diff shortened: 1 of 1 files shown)


fanminshi:
cc/ @hongchaodeng  @hasbro17

hongchaodeng:
Also update this method:
```
 func (b *Backup) reportBackupStatus(err error, eb *api.EtcdBackup) {
```
to make the `err` not coupled with error below?

fanminshi:
All fixed. PTAL @hongchaodeng

hongchaodeng:
lgtm

fanminshi:
merge when green.

PR #1573, opened 2017-10-26
*: remove pv-provisioner and create-storage-class flags

hasbro17:
Removed all logic related to the `pv-provisioner` and `create-storage-class` flags.

The backup policy validation at the controller now includes checking that storage class must be set, else the event is not handled with a warning.
```
fail to handle event: invalid cluster spec. please fix the following problem with the cluster spec: PV backup must have a storage class set
```

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -50,30 +50,25 @@ import (
 )
 
 var (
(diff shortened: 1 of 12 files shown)


hongchaodeng:
lgtm

PR #1497, opened 2017-10-10
e2e: add etcd backup operator test

fanminshi:
Build unified image in build_pass allowing e2e test to run for backup_operator from the image.

Add an e2e test for etcd backup operator:

1. spawn etcd backup operator pod.
2. create etcd cluster.
3. invoke a backup cr for the etcd cluster.
4. verify that backup status shows succeeded.

PR #1576, opened 2017-10-26
hack: build unified image in fmt_build

fanminshi:
fmt_build now builds a unified operator image where it can be used to run backup_operator and restore_operator. 

e2e test can use this image to run backup_operator and restore_operator without requiring individual image.

### DIFF
--- a/hack/test
+++ b/hack/test
@@ -98,6 +98,7 @@ function build_pass {
 	IMAGE=$OPERATOR_IMAGE hack/build/operator/build
 	build_backup_operator
 	build_restore_operator
+	IMAGE=$OPERATOR_IMAGE hack/build/docker_push
 }
 
 function e2e_pass {


fanminshi:
cc/ @hongchaodeng @hasbro17

hongchaodeng:
Don't separate this out from https://github.com/coreos/etcd-operator/pull/1497.

We should only push the image only after we have the tests for the binaries.

fanminshi:
@hongchaodeng this also ensure etcd-operator from the unify image can be run for e2e test. I think separating it out here shouldn't be a problem.

hongchaodeng:
>  this also ensure etcd-operator from the unify image can be run for e2e test. 

It doesn't make sense to test the new unified image without the test.
Let's just close this PR and include this one line in the test PR.

hongchaodeng:
Closing this and merging the effort into #1497

PR #1577, opened 2017-10-26
*: generalize CreateCRD and WaitCRDReady

fanminshi:
CreateCRD now creates CRD based on crd name, crd kind, crd plural, and shortName.
WaitCRDReady now waits crd based on crd name.

Replace usage of keyword CRDName and related CRD defs to be prefixed with the keyword EtcdCluster.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -120,9 +120,9 @@ func (c *Controller) makeClusterConfig() cluster.Config {
 }
 
 func (c *Controller) initCRD() error {
-	err := k8sutil.CreateCRD(c.KubeExtCli)
+	err := k8sutil.CreateCRD(c.KubeExtCli, api.EtcdClusterCRDName, api.EtcdClusterResourceKind, api.EtcdClusterResourcePlural, "etcd")
 	if err != nil {
 		return fmt.Errorf("failed to create CRD: %v", err)
 	}
-	return k8sutil.WaitCRDReady(c.KubeExtCli)
+	return k8sutil.WaitCRDReady(c.KubeExtCli, api.EtcdClusterCRDName)
 }
--- a/pkg/util/k8sutil/crd.go
+++ b/pkg/util/k8sutil/crd.go
@@ -48,22 +48,22 @@ func GetClusterList(restcli rest.Interface, ns string) (*api.EtcdClusterList, er
 }
 
 func listClustersURI(ns string) string {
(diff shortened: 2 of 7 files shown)


fanminshi:
cc/ @hongchaodeng @hasbro17

pkg/util/k8sutil/crd.go
```diff
@@ -51,19 +51,18 @@ func listClustersURI(ns string) string {
...
+			Name: crdName,
 		},
 		Spec: apiextensionsv1beta1.CustomResourceDefinitionSpec{
 			Group:   api.SchemeGroupVersion.Group,
 			Version: api.SchemeGroupVersion.Version,
 			Scope:   apiextensionsv1beta1.NamespaceScoped,
 			Names: apiextensionsv1beta1.CustomResourceDefinitionNames{
-				Plural:     api.CRDResourcePlural,
-				Kind:       api.CRDResourceKind,
-				ShortNames: []string{"etcd"},
```
hongchaodeng: This removes *ShortNames*. It shouldn't, right?

pkg/controller/controller.go
```diff
@@ -120,9 +120,9 @@ func (c *Controller) makeClusterConfig() cluster.Config {
 }
 
 func (c *Controller) initCRD() error {
-	err := k8sutil.CreateCRD(c.KubeExtCli)
+	err := k8sutil.CreateCRD(c.KubeExtCli, api.CRDName, api.CRDResourceKind, api.CRDResourcePlural)
```
hongchaodeng: Can you also change `CRDName` to `EtcdClusterCRDName`?

pkg/util/k8sutil/crd.go
fanminshi: I am unsure on how short name is being used. do you know if is needed? maybe i can just add another parameter call `shortName`.

pkg/controller/controller.go
fanminshi: yeah

pkg/util/k8sutil/crd.go
hongchaodeng: > maybe i can just add another parameter call shortName.

Right.

fanminshi:
All fixed. PTAL cc/ @hongchaodeng @hasbro17

hongchaodeng:
lgtm

fanminshi:
merge when green.

PR #1581, opened 2017-10-26
controller/restore-operator: init crd when starting restore operator

fanminshi:
let restore operator creates the crd it needs when starting. 
In this way, user doesn't have to create the crd beforehand.


fanminshi:
cc/ @hongchaodeng @hasbro17

hasbro17:
LGTM

PR #1583, opened 2017-10-26
util: only add shortName if it is not empty

fanminshi:
ShortNames can't be empty. Or else kubernetes api server complains:

```
spec.names.shortNames[0]: Invalid value: "": a DNS-1035 label must consist of lower case alphanumeric characters or '-', and must start and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')"
```


fanminshi:
cc/ @hongchaodeng @hasbro17

hongchaodeng:
lgtm

PR #1589, opened 2017-10-27
*: support backup operator creation in e2e

fanminshi:
updates fmt_build to push all operators into one image.
enables e2e framework to run/delete etcd backup operator.

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -44,6 +44,8 @@ import (
 
 var Global *Framework
 
+const etcdBackupOperatorName = "etcd-backup-operator"
+
 type Framework struct {
 	opImage          string
 	KubeClient       kubernetes.Interface
@@ -88,6 +90,10 @@ func Teardown() error {
 	if err := Global.deleteEtcdOperator(); err != nil {
 		return err
 	}
+	err := Global.KubeClient.CoreV1().Pods(Global.Namespace).Delete(etcdBackupOperatorName, metav1.NewDeleteOptions(1))
+	if err != nil {
+		return fmt.Errorf("failed to delete etcd backup operator: %v", err)
+	}
 	// TODO: check all deleted and wait
 	Global = nil
 	logrus.Info("e2e teardown successfully")
@@ -107,6 +113,12 @@ func (f *Framework) setup() error {
 			return fmt.Errorf("fail to setup aws: %v", err)
 		}
 	}
+	err := f.SetupEtcdBackupOperator()
+	if err != nil {
+		return fmt.Errorf("failed to create etcd backup operator: %v", err)
+	}
+	logrus.Info("etcd backup operator created successfully")
+
 	logrus.Info("e2e setup successfully")
 	return nil
 }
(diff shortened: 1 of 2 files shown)


fanminshi:
cc/ @hongchaodeng @hasbro17

test/e2e/e2eutil/spec_util.go
```diff
@@ -52,6 +52,29 @@ func NewS3BackupPolicy(cleanup bool) *api.BackupPolicy {
 	}
 }
 
+// NewS3Backup creates a EtcdBackup object using clusterName.
```
hongchaodeng: Don't need this in this PR?

test/e2e/framework/framework.go
```diff
@@ -44,6 +44,8 @@ import (
 
 var Global *Framework
 
+const etcdBackupOperator = "etcd-backup-operator"
```
hongchaodeng: etcdBackupOperatorName

test/e2e/framework/framework.go
```diff
@@ -88,6 +90,9 @@ func Teardown() error {
 	if err := Global.deleteEtcdOperator(); err != nil {
 		return err
 	}
+	if err := Global.KubeClient.CoreV1().Pods(Global.Namespace).Delete(etcdBackupOperator, metav1.NewDeleteOptions(1)); err != nil {
```
hongchaodeng:
```
err = Delete()
if err != nil {...
```

test/e2e/framework/framework.go
```diff
@@ -107,6 +112,11 @@ func (f *Framework) setup() error {
 			return fmt.Errorf("fail to setup aws: %v", err)
 		}
 	}
+	if err := f.SetupEtcdBackupOperator(); err != nil {
```
hongchaodeng:
```
err = Delete()
if err != nil {...
```

fanminshi: hmm, isn't `if err := f.SetupEtcdBackupOperator(); err != nil {` shorter since it is one liner.

hongchaodeng: It's the current convention.
We should discourage any code like that:
```
if xxx := xxx(); xxx {
}
```

fanminshi: maybe I am missing something.
all setup functions has the following style.
```go
if xxx := xxx(); xxx {
}
```

```go
func (f *Framework) setup() error {
	if err := f.setupStorageClass(); err != nil {
		return fmt.Errorf("failed to setup storageclass(%v): %v", f.StorageClassName, err)
	}
	if err := f.SetupEtcdOperator(); err != nil {
		return fmt.Errorf("failed to setup etcd operator: %v", err)
	}
	logrus.Info("etcd operator created successfully")
	if os.Getenv("AWS_TEST_ENABLED") == "true" {
		if err := f.setupAWS(); err != nil {
			return fmt.Errorf("fail to setup aws: %v", err)
		}
	}
	if err := f.SetupEtcdBackupOperator(); err != nil {
		return fmt.Errorf("failed to create etcd backup operator: %v", err)
	}
	logrus.Info("etcd backup operator created successfully")

	logrus.Info("e2e setup successfully")
	return nil
}
```

Anyways, I think keeping the code consistent is the key here.

hongchaodeng: That should be discouraged

fanminshi: alright then.

fanminshi:

PR #1596, opened 2017-10-29
controller/restore-operator: http server takes in restore name to find restore source

hasbro17:
Addresses #1588 

The http server now accepts the restore-name in a request and uses that to look up the associated restore source, i.e restoreName -> (AWSSecret, S3Path).
The s3Reader is then used to return the backup file.

PR #1597, opened 2017-10-29
controller/restore-operator: use restore name instead of cluster name

hasbro17:
Part 1 of 2 of #1596

Changes the restore-operator to track a restore via restore CR name instead of spec.BackupSpec.ClusterName.

### DIFF
--- a/pkg/controller/restore-operator/http.go
+++ b/pkg/controller/restore-operator/http.go
@@ -24,6 +24,7 @@ import (
 	"github.com/coreos/etcd-operator/pkg/backup/backupapi"
 	backups3 "github.com/coreos/etcd-operator/pkg/backup/s3"
 	"github.com/coreos/etcd-operator/pkg/util/awsutil/s3factory"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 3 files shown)


pkg/controller/restore-operator/operator.go
```diff
@@ -45,10 +45,10 @@ type Restore struct {
 	etcdCRCli  versioned.Interface
 	kubeExtCli apiextensionsclient.Interface
 
-	// restoreCRs is a map of cluster name to restore cr.
+	// restoreCRs is a map of restore name to restore cr.
```
hongchaodeng: After a thorough digging, these two maps are outdated if we don't use clusterName but use restoreName.

All we need is an index that provide Get(restore_name) -> CR.
- We don't need `restoreCRs` and `clusterNames `, and we can get rid those.
- We don't need `restoreNames` to be added. Instead, we can make use of indexer.Store to get the object.

hongchaodeng: See https://github.com/kubernetes/client-go/blob/master/tools/cache/store.go#L194

pkg/controller/restore-operator/http.go
```diff
@@ -39,27 +39,27 @@ func (r *Restore) startHTTP() {
...
-	v, ok := r.restoreCRs.Load(clusterName)
+	v, ok := r.restoreCRs.Load(restoreName)
 	if !ok {
-		http.Error(w, fmt.Sprintf("cluster %v backup server not found", clusterName), http.StatusInternalServerError)
+		http.Error(w, fmt.Sprintf("restore %v backup server not found", restoreName), http.StatusInternalServerError)
 		return
 	}
 
-	logrus.Infof("serving backup for cluster %v", clusterName)
+	logrus.Infof("serving backup for cluster %v", restoreName)
```
fanminshi: maybe change this log to something else since we are not reasoning about cluster anymore.

hasbro17:
@hongchaodeng @fanminshi PTAL.
The http server now uses the restore CR name and the indexer to directly look up the restore CR object.

Manually tested by not creating the seed pod and manually requesting the backup with some restore-name. It successfully finds the restore CR associated with that restore-name. Returns an error if restore CR not found.

pkg/controller/restore-operator/http.go
```diff
@@ -39,27 +40,42 @@ func (r *Restore) startHTTP() {
...
+// serveBackup parses incoming request url of the form /backup/<restore-name>
+// get the etcd restore name.
 // Then it returns the etcd cluster backup snapshot to the caller.
 func (r *Restore) serveBackup(w http.ResponseWriter, req *http.Request) {
-	clusterName := string(req.URL.Path[len(backupHTTPPath):])
-	if len(clusterName) == 0 {
-		http.Error(w, "cluster is not specified", http.StatusBadRequest)
+	restoreName := string(req.URL.Path[len(backupHTTPPath):])
+	if len(restoreName) == 0 {
+		http.Error(w, "restore name is not specified", http.StatusBadRequest)
```
hongchaodeng: Should have been: http.StatusBadRequest -> http.StatusNotFound

Can you change it alongside?

PR #1599, opened 2017-10-30
*: expose ParseBucketAndKey

fanminshi:
ParseBucketAndKey can be used for backup writer.

### DIFF
--- a/pkg/backup/reader/s3_reader.go
+++ b/pkg/backup/reader/s3_reader.go
@@ -17,10 +17,10 @@ package reader
 import (
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


fanminshi:
cc/ @hasbro17 @hongchaodeng

hongchaodeng:
Can you submit a PR for writer interface and implementation? This PR should be part of that.
It's not preferred to change a private method to public without any actual use case.

fanminshi:
@hongchaodeng `ParseBucketAndKey` will be used in writer implementation. In that case, I'll include this pr with the writer implementation.

Issue #1588, opened 2017-10-27
Implementation plan for restore from S3path


hongchaodeng:
Plan:
- refactor S3 client layer (https://github.com/coreos/etcd-operator/pull/1587)
- clusterName -> restoreName (https://github.com/coreos/etcd-operator/pull/1597)
- Change the HTTP server that takes in restoreName -> (AWSSecret/S3client, S3Path) and returns backup. (https://github.com/coreos/etcd-operator/pull/1602)
- Change the seed member (https://github.com/coreos/etcd-operator/pull/1605)
  - http query
  - use restoreName to induce clusterName 
- remove old BackupSpec in restore CR (https://github.com/coreos/etcd-operator/pull/1610)

hongchaodeng:
All done

PR #1602, opened 2017-10-30
controller/restore-operator: use s3Reader to read backup

hasbro17:
Addresses #1588 
Part 2 of #1596 
restore-operator now uses the `S3RestoreSource` of a restore CR to read and return a backup via the s3Reader.
Also fixed the `RestoreSource ` to be embedded so the s3 field can be specified directly in the backup CR like:
```
spec:
  clusterSpec:
    size: 3
    version: "3.1.8"
  s3:
    path: jenkins-etcd-operator/test-prefix/v1/e2e-local/s3-cluster-backup/3.1.8_0000000000000001_etcd.backup
    awsSecret: aws
```

### DIFF
--- a/pkg/apis/etcd/v1beta2/restore_types.go
+++ b/pkg/apis/etcd/v1beta2/restore_types.go
@@ -48,7 +48,7 @@ type RestoreSpec struct {
 	BackupSpec BackupSpec `json:"backupSpec"`
 	// TODO: Remove BackupSpec once RestoreSource is implemented
 	// RestoreSource tells the where to get the backup and restore from.
-	RestoreSource RestoreSource `json:",inline"`
+	RestoreSource `json:",inline"`
 }
 
 type RestoreSource struct {
@@ -57,7 +57,9 @@ type RestoreSource struct {
 }
 
 type S3RestoreSource struct {
-	// The S3 path where the backup is saved.
+	// Path is the full s3 path where the backup is saved.
+	// The format of the path must be: "<s3-bucket-name>/<path-to-backup-file>"
+	// e.g: "etcd-backups/v1/default/example-etcd-cluster/3.1.8_0000000000000001_etcd.backup"
 	Path string `json:"path"`
 
 	// The name of the secret object that stores the AWS credential and config files.
--- a/pkg/controller/restore-operator/http.go
+++ b/pkg/controller/restore-operator/http.go
@@ -15,18 +15,18 @@
 package controller
 
 import (
+	"errors"
 	"fmt"
+	"io"
 	"net/http"
 
 	api "github.com/coreos/etcd-operator/pkg/apis/etcd/v1beta2"
-	"github.com/coreos/etcd-operator/pkg/backup"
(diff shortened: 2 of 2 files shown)


pkg/controller/restore-operator/http.go
```diff
@@ -72,31 +71,38 @@ func (r *Restore) serveBackup(w http.ResponseWriter, req *http.Request) {
...
 	cr := v.(*api.EtcdRestore)
-	spec := cr.Spec.BackupSpec
-	switch spec.StorageType {
-	case api.BackupStorageTypeS3:
-		bs, cli, err := r.makeBackupServer(spec.S3, spec.ClusterName)
-		if err != nil {
-			http.Error(w, "failed to create S3 backup server", http.StatusInternalServerError)
-			return
-		}
+	// TODO: case and switch on RestoreSource type when it exists
```
hongchaodeng: Maybe structure it like:
```
switch {
case restoreSource.S3 != nil:
  ...
case ...: ...
default: ...
}
```

pkg/apis/etcd/v1beta2/restore_types.go
```diff
@@ -57,7 +57,9 @@ type RestoreSource struct {
 }
 
 type S3RestoreSource struct {
-	// The S3 path where the backup is saved.
+	// The full S3 path where the backup is saved.
```
fanminshi: `Path is ful s3 path ...`?

pkg/apis/etcd/v1beta2/restore_types.go
```diff
@@ -57,7 +57,9 @@ type RestoreSource struct {
...
+	// The format of the path should be: "<s3-bucket-name>/<path-to-backup-file>"
```
fanminshi: `should be` -> `must be`?

pkg/controller/restore-operator/http.go
hongchaodeng: should be simple change :)
Can you also make it here?

fanminshi:
small nits. defer to @hongchaodeng

pkg/controller/restore-operator/http.go
```diff
@@ -72,31 +71,38 @@ func (r *Restore) serveBackup(w http.ResponseWriter, req *http.Request) {
...
 
-		bs.ServeBackup(w, req)
-		cli.Close()
-	default:
-		http.Error(w, fmt.Sprintf("unknown storage type %v", spec.StorageType), http.StatusBadRequest)
+	// Validate restore source
+	if s3RestoreSource == nil || len(s3RestoreSource.AWSSecret) == 0 || len(s3RestoreSource.Path) == 0 {
+		msg := fmt.Sprint("invalid restore source field (spec.s3), must specify all required fields")
+		logrus.Error(msg)
+		http.Error(w, msg, http.StatusInternalServerError)
```
hongchaodeng: Abstract them into a method so that we can handle error centrally?

hasbro17:
fixed PTAL

PR #1610, opened 2017-10-31
*: remove BackupSpec from restore CR spec

hasbro17:
Remove `BackupSpec` now that restore CR uses `RestoreSource`.


fanminshi:
lgtm defer @hongchaodeng

hongchaodeng:
lgtm

Issue #1611, opened 2017-10-31
doc: restore operator


hongchaodeng:
plan:
- add README
- add basic test (https://github.com/coreos/etcd-operator/pull/1616)

PR #1616, opened 2017-11-01
e2e: add e2e test for restore operator

hasbro17:
Addresses #1611 

The e2e test for the restore operator does the following:
- Create an EtcdRestore CR using the awsSecret and S3 path passed in
-  Wait until the CR status has `succeeded=true`
- Verify that the restored etcd cluster reaches the desired size

The restore e2e test needs the S3 path to be set. To do this the backup and restore tests are run sequentially:
- The backup operator test is `testEtcdBackupOperatorForS3Backup` is run first.
- That will return the S3 path if successful.
- The restore operator test runs after that using the S3 path returned by the previous test.

### DIFF
--- a/test/e2e/backup_test.go
+++ b/test/e2e/backup_test.go
@@ -21,20 +21,31 @@ import (
 	"testing"
 	"time"
 
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/service/s3"
+	api "github.com/coreos/etcd-operator/pkg/apis/etcd/v1beta2"
 	"github.com/coreos/etcd-operator/pkg/util/retryutil"
 	"github.com/coreos/etcd-operator/test/e2e/e2eutil"
 	"github.com/coreos/etcd-operator/test/e2e/framework"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/s3"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// TestEtcdBackupOperatorForS3Backup tests if etcd backup operator can save etcd backup to S3.
-func TestEtcdBackupOperatorForS3Backup(t *testing.T) {
+// TestBackupAndRestore runs the backup test first, and only runs the restore test after if the backup test succeeds and sets the S3 path
+func TestBackupAndRestore(t *testing.T) {
 	if os.Getenv(envParallelTest) == envParallelTestTrue {
 		t.Parallel()
 	}
+	s3Path := testEtcdBackupOperatorForS3Backup(t)
+	if len(s3Path) == 0 {
+		t.Fatal("skipping restore test: S3 path not set despite testEtcdBackupOperatorForS3Backup success")
+	}
+	testEtcdRestoreOperatorForS3Source(t, s3Path)
+}
+
+// testEtcdBackupOperatorForS3Backup tests if etcd backup operator can save etcd backup to S3.
+// It returns the full S3 path where the backup is saved.
+func testEtcdBackupOperatorForS3Backup(t *testing.T) string {
 	f := framework.Global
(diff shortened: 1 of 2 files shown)


hasbro17:
@hongchaodeng @fanminshi PTAL

test/e2e/backup_test.go
```diff
@@ -21,20 +21,37 @@ import (
...
+	"github.com/aws/aws-sdk-go/service/s3"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// TestEtcdBackupOperatorForS3Backup tests if etcd backup operator can save etcd backup to S3.
-func TestEtcdBackupOperatorForS3Backup(t *testing.T) {
-	if os.Getenv(envParallelTest) == envParallelTestTrue {
-		t.Parallel()
+// TestS3Path is the S3 path set by testEtcdBackupOperatorForS3Backup and used by testEtcdRestoreOperatorForS3Source
+var TestS3Path = ""
```
fanminshi: no need to expose `TestS3Path`?

test/e2e/backup_test.go
```diff
@@ -21,20 +21,37 @@ import (
...
+
+// TestBackupAndRestore runs the backup test first, and only runs the restore test after if the backup test succeeds and sets the S3 path
+func TestBackupAndRestore(t *testing.T) {
+	ok := t.Run("TestEtcdBackupOperatorForS3Backup", func(t *testing.T) {
+		testEtcdBackupOperatorForS3Backup(t)
```
hongchaodeng: Make `testEtcdBackupOperatorForS3Backup()` return s3Path.
Don't use go subtest

test/e2e/e2eutil/spec_util.go
```diff
@@ -75,6 +75,31 @@ func NewS3Backup(clusterName string) *api.EtcdBackup {
 	}
 }
 
+// NewS3EtcdRestore returns an EtcdRestore CR with an S3 restore source
+func NewS3EtcdRestore(restoreName, version, path, awsSecret string, size int) *api.EtcdRestore {
```
hongchaodeng: NewEtcdRestore(... *RestoreSource)

test/e2e/e2eutil/spec_util.go
```diff
@@ -75,6 +75,31 @@ func NewS3Backup(clusterName string) *api.EtcdBackup {
...
+		Spec: api.RestoreSpec{
+			ClusterSpec: api.ClusterSpec{
+				Size:    size,
+				Version: version,
+			},
+			RestoreSource: api.RestoreSource{
+				S3: &api.S3RestoreSource{
+					Path:      path,
+					AWSSecret: awsSecret,
+				},
```
fanminshi: can we use `os.Getenv("TEST_AWS_SECRET")` here?

test/e2e/backup_test.go
```diff
@@ -87,3 +105,62 @@ func TestEtcdBackupOperatorForS3Backup(t *testing.T) {
...
+		t.Fatalf("failed to create etcd restore cr: %v", err)
+	}
+	defer func() {
+		if err := f.CRClient.EtcdV1beta2().EtcdRestores(f.Namespace).Delete(er.Name, nil); err != nil {
+			t.Fatalf("failed to delete etcd restore cr: %v", err)
+		}
+	}()
+
+	// Verify the EtcdRestore CR status "succeeded=true"
+	err = retryutil.Retry(time.Second, 5, func() (bool, error) {
```
hongchaodeng: Can you write down what's the normal finished time?

hasbro17: In practice, it usually takes about 1 second for `succeeded=true` from the time of the restore CR creation.
I can add a comment. Do you want to keep the retry time and attempts the same?

hongchaodeng: Just write down the time is enough. In case of flake, there is reference to look at.

hasbro17:
fixed. PTAL.

Issue #1617, opened 2017-11-02
Allow members to advertise using IPs


jamiehannaford:
Is it possible for etcd members to advertise themselves via their IPs rather than DNS hostnames? Since they're on hostNetwork, I assume they will advertise their host IP. 

The reason this is needed is that we're trying to integrate etcd-operator into kubeadm, but we do not have a working DNS server on bootstrap (because there is no CNI yet). The user sets up CNI _after_ the control plane is working. We also want kube-apiserver on each node to hit its own etcd pod via `localhost`, so we really have no need to use services in our use case.

@hongchaodeng @xiang90 WDYT?

/cc @luxas

jamiehannaford:
I'm suggesting making this configurable, with a new field in `EtcdCluster` or something.

jamiehannaford:
Looks like this was already considered in https://github.com/coreos/etcd-operator/pull/122. 

It seems that using the operator without a working DNS server is possible, just requires a bit of hacking because the host IP is only known after the pod is scheduled. @xiang90 what would be needed to make that work?

hongchaodeng:
Hi @jamiehannaford 

It's possible to use Pod IP. Actually we initially use pod IP in self hosted etcd to avoid depending on kubedns. The issue arises when adding TLS certs because IPs are dynamic.

With using of k8s csr endpoint we can switch to use pod IP: https://github.com/coreos/etcd-operator/issues/1465

jamiehannaford:
@hongchaodeng Oh, cool! I think we use CSRs in kubeadm too with auto-approve so that might work really well. So I'm imagining how this would work...

1. We'd need to update [`addOneSelfHostedMember`](https://github.com/coreos/etcd-operator/blob/c7131f581f00bf0901d24d23fc1768f7103ba01d/pkg/cluster/self_hosted.go#L120) to find the node with no etcd pod on it. Then pass that node's IP into `NewSelfHostedEtcdPod` which can generate the CSR and make the pod command. 

2. Add new fields to the spec struct for dynamic TLS. I'm thinking the RBAC user or group name that the operator will use to send the CSRs. 

3. Use `status.podIP` if the user wants to use pod IPs instead of DNS.

Did I miss anything?

hongchaodeng:
> We'd need to update addOneSelfHostedMember to find the node with no etcd pod on it. 

Scheduler does this for us via anti-affinity.

[...]

Issue #1465, opened 2017-10-01
Use Kubernetes CA for certificates


F21:
Since Kubernetes 1.7, it is possible to ask [Kubernetes to sign certificates for us](https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/).

It would be nice if the operator can automatically generate certificates, ask Kubernetes to sign them, automatically approve the certificate and use them in the deployed etcd clusters.

This would allow the following:
- Use the Kubernetes CA to secure workloads on the cluster.
- Simplify the setup of secure etcd clusters in Kubernetes.
- Support certificates with low lifetimes (24hrs) and automatically issue new ones and rotate them.

hongchaodeng:
There is already a design for dynamic TLS: https://github.com/coreos/etcd-operator/blob/master/doc/design/cluster_tls.md

Note that it is been a while and a bit outdated. We can base off the original design and make use of k8s CSR API to implement dynamic TLS

jamiehannaford:
Auto-approval only works if the user that's requesting the CSR is in the `system:node` group (e.g. a kubelet). You can see the approval logic here:

client: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/certificates/approver/sarapprove.go#L189-L191
server: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/certificates/approver/sarapprove.go#L221-L226

So I'm not sure this can work right now unless you write another CSR approver controller

hongchaodeng:
Hi @jamiehannaford 

Yes. I understand auto-approval only works for node TLS bootstrap now. But this is very generic and in long term should be able to extend to auto-approve other groups too. In short term, manual approval or custom auto-approval worker should be better choice.

Does that make sense? If so let's create a feature request on that :)

jamiehannaford:
@hongchaodeng Creating a custom approval controller is a lot of work, wouldn't it just be simpler as a first step to generate a CA for etcd and provide it in the EtcdCluster spec? That would allow dynamic TLS, since you're generating certs for each etcd pod as it's being created.

hongchaodeng:
> you're generating certs for each etcd pod as it's being created.

That's duplicating the work by k8s CSR API. And what this issue is created for and what we all wanted from this issue is to make use of that API.

jamiehannaford:
@hongchaodeng Okay. Should a custom controller that approves etcd CSRs be added to this codebase?

hongchaodeng:
[...]

PR #1625, opened 2017-11-06
[WIP] Add init container script

jamiehannaford:
This init script does a few things:

- Generates client, peer and serving certs from a provided CA pair (hostPath PV)
- Adds the new member using its POD_IP to the cluster (this will result in temporary failures until the real container starts running)

Some stuff left to do:

- [ ] Add logic if certs already exist
- [ ] DRY things up a bit 
- [ ] Test this with modified [NewSelfHostedEtcdPod](https://github.com/coreos/etcd-operator/blob/master/pkg/util/k8sutil/self_hosted.go#L40) and make sure it works together

@hongchaodeng I wanted to submit my initial work to make sure it's correct. WDYT?

### DIFF
--- a/cmd/init/main.go
+++ b/cmd/init/main.go
@@ -1,0 +1,174 @@
+package main
+
+import (
+	"context"
+	"crypto/rsa"
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"net"
+	"os"
+	"path/filepath"
+	"reflect"
+	"strings"
+	"time"
+
+	"github.com/coreos/etcd/clientv3"
+	"github.com/coreos/etcd/pkg/transport"
+	certutil "k8s.io/client-go/util/cert"
+)
+
+const (
+	certDir        = "/certs"
+	caCertFile     = "ca.crt"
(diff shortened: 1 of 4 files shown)


hongchaodeng:
What's the motivation for this PR? Can you explain more?

jamiehannaford:
@hongchaodeng You said we needed an init container in [this comment](https://github.com/coreos/etcd-operator/issues/1617#issuecomment-341488152):

> DNS would be get ridden of. Instead of adding member in etcd operator, there will be an init container that does the add member command.

So this PR adds a new binary to do the MemberAdd and cert generation.

hongchaodeng:
Hi @jamiehannaford 

We need to implement dynamic TLS first before switching to host IP (https://github.com/coreos/etcd-operator/issues/1617#issuecomment-341492025).
Our use case requires that: https://github.com/coreos/etcd-operator/issues/1465#issuecomment-342124690

jamiehannaford:
Closing in favour of the dynamic TLS design. We can reopen later if necessary.

Issue #1626, opened 2017-11-06
Decoupling Backup and Restore Logic from Etcd Operator


hongchaodeng:
# Decoupling Backup and Restore Logic from Etcd Operator

Currently the [backup][backup_policy] and [restore][restore_policy] are provided as features in the EtcdCluster API. These features have been useful and adopted by open source users. Over time, there are increasing requests to extend the functionalities (#1316, #1241). On the other hand, there is a limit for adding the features due to: 1. extending the API needs to happen in upstream codebase; 2. it is conflicted with making current API stable. Especially (2), because it blocks stabilizing the project.

To solve the above problems, we propose new ways to do backup and restore:

- Backup: Currently EtcdCluster includes [BackupPolicy][backup_policy] to periodically save etcd snapshots. Instead, we propose to add a new [EtcdBackup][etcd_bakcup_api] API which includes a reference to the EtcdCluster that it wants to make backup from. We built [etcd-backup-operator][backup_operator_doc] as an example to save backups onto S3. As a result, etcd operator doesn’t need to know anything about backup.
- Restore: Currently EtcdCluster includes [RestorePolicy][restore_policy] to restore an etcd member from a backup. Instead, we propose to add a new [EtcdRestore][etcd_restore_api] API which includes the backup resource information and the EtcdCluster to be restored. We built [etcd-restore-operator][restore_operator_doc] as an example to restore the data and create a new EtcdCluster. etcd operator can still manage the restored cluster, but it doesn’t need to know anything about the restore process.

By using the above methods, we can decouple backup and restore logic from etcd operator, and thus can remove BackupPolicy and RestorePolicy from EtcdCluster API.

## User effect

Existing backup and restore features in EtcdCluster API won’t be supported after 0.7.0 release. Please check [migration docs][migrate_doc] before upgrading.

The open source etcd-backup-operator and etcd-restore-operator are provided as examples for how to use EtcdCluster API. Based on what we know, most Internet company users have their own ways to do etcd backup and restore. We expect these users would write their own tooling and make their work easier by decoupling them from our codebase. For enterprise users, we recommend trying out [Tectonic](https://coreos.com/tectonic/) and make [contacts](https://coreos.com/contact/) for customized etcd backup/restore solutions.

Moving forward, we will focus on stabilizing etcd operator and make a stable release soon.

[...]

PR #1628, opened 2017-11-07
hack/test: remove init for minio and azurite

fanminshi:
ref: https://github.com/coreos/etcd-operator/issues/1626

minio and azurite act as a mock backend storage for S3 and ABS. it is used
for s3/abs backend unit testing.
removing those is part of the effort to remove s3/abs backend from etcd opreator.

### DIFF
--- a/hack/test
+++ b/hack/test
@@ -137,53 +137,7 @@ function upgrade_pass {
 		--new-image=$UPGRADE_TO
 }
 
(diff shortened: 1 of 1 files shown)


fanminshi:
cc/ @hongchaodeng @hasbro17

hasbro17:
LGTM

PR #1630, opened 2017-11-07
*: remove backup side car

fanminshi:
ref:https://github.com/coreos/etcd-operator/issues/1626


fanminshi:
cc/ @hongchaodeng @hasbro17

hack/build/operator/build
```diff
@@ -40,7 +40,6 @@ gitHash="github.com/coreos/etcd-operator/version.GitSHA=${GIT_SHA}"
 go_ldflags="-X ${ldKVPairs} -X ${gitHash}"
```
hongchaodeng: Change this script to only do build. Now it also does push.

fanminshi: @hongchaodeng do you want me to do the above in this pr or a separate pr?

hongchaodeng: Do it in separate pr.

hongchaodeng:
lgtm

PR #1639, opened 2017-11-08
*: Improve go build

fanminshi:
remove go install from build scripts.
go build for operators can now take-in extra flags.
refactor go_build as a standalone lib in hack/lib.

### DIFF
--- a/hack/build/backup-operator/build
+++ b/hack/build/backup-operator/build
@@ -1,19 +1,12 @@
 #!/usr/bin/env bash
 
+# Note: add `-i` to enable build with caching. e.g ./hack/backup-operator/build -i
+
 set -o errexit
 set -o nounset
 set -o pipefail
 
-function go_build {
-	echo "building "${1}"..."
-	if [ ! -z ${GOINSTALL+x} ] && [ "${GOINSTALL}" = "y" ]
-	then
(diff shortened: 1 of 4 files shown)


fanminshi:
cc/ @hongchaodeng

hack/test
```diff
@@ -3,6 +3,7 @@
 set -o errexit
 set -o nounset
 set -o pipefail
+set -x
```
hongchaodeng: Remove this?

hack/build/backup-operator/build
```diff
@@ -6,13 +6,7 @@ set -o pipefail
 
 function go_build {
```
hongchaodeng: Can you put this into a lib?

hack/test
```diff
@@ -95,9 +96,10 @@ function build_restore_operator  {
 }
 
 function build_pass {
-	IMAGE=$OPERATOR_IMAGE hack/build/operator/build
-	build_backup_operator
-	build_restore_operator
+	# cache all builds by passing '-i'. 
+	IMAGE=$OPERATOR_IMAGE hack/build/operator/build -i
```
hongchaodeng: Don't enable install in CI.

fanminshi: should we enable cache in CI so that is will be faster to run? I thought that's the purpose of the support caching for go build.

hongchaodeng: >  I thought that's the purpose of the support caching for go build.

No. We add it for local dev purpose.

fanminshi:
PTAL cc/ @hongchaodeng

hongchaodeng:
lgtm

PR #1641, opened 2017-11-08
hack: remove push from operator build

fanminshi:
operator build should only build operator binary not pushing it.


fanminshi:
cc/ @hongchaodeng

hongchaodeng:
lgtm

fanminshi:
merge when green.

PR #1647, opened 2017-11-09
util: remove unused backup until functions

fanminshi:
ref: #1626


fanminshi:
cc/ @hongchaodeng @hasbro17

pkg/util/k8sutil/backup.go
```diff
@@ -14,244 +14,5 @@
...
-	if err != nil {
-		wErr := fmt.Errorf("fail to wait PVC (%s) '(%v)/Bound': %v", name, claim.Status.Phase, err)
-		return wErr
-	}
-
-	return nil
-}
-
+// What's this?
 var BackupImage = "quay.io/coreos/etcd-operator:latest"
```
hasbro17: I think this was the image used for the backup sidecar pod. So we probably don't need this as well.

hongchaodeng: Yep. Just delete it.

fanminshi: @hasbro17 it is used here https://github.com/coreos/etcd-operator/blob/master/hack/build/operator/build#L27

hongchaodeng: That's the build flag used for backup sidecar previously.
Delete that as well

fanminshi: done. PTAL @hongchaodeng @hasbro17

hongchaodeng:
lgtm

PR #1655, opened 2017-11-10
e2e: remove storage class creation from frameworks

hasbro17:
ref: #1626 
The test frameworks(e2e and upgrade) no longer need to create the StorageClass since there are no PV backup tests.

With the [current rbac permissions](https://github.com/coreos/etcd-operator/blob/master/example/rbac/cluster-role-template.yaml) the tests should have failed up until now due to a lack of permissions for storageclass but since GKE has no rbac, the tests would run fine.

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -19,7 +19,6 @@ import (
 	"flag"
 	"fmt"
 	"os/exec"
-	"path"
 	"strconv"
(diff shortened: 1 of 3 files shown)


hongchaodeng:
ref: https://github.com/coreos/etcd-operator/issues/1626

hongchaodeng:
lgtm

Issue #1640, opened 2017-11-08
Provide scripts to run e2e tests inside a pod


hasbro17:
Being able to run e2e tests inside a pod on a kubernetes cluster would make it easy to run the e2e tests during local development without testing against the CI.

- [x] Run e2e tests (https://github.com/coreos/etcd-operator/pull/1638)
- [x] Have the `run-test-pod` script check if the test-pod succeeds or fails
- [x] Have the `run-test-pod` script setup the RBAC for the test-pod and tests 
- [x] Log collector program to collect all interested logs (e.g. e2e test, etcd operator)
- [x] Consolidate build scripts for operator and test-pod and run them in a container.
- [x] Use it in Jenkins
  - [x] Change existing workflow: Build the operator image first, then run e2e tests in a pod
- [x] use hack/ci/run to:
  - build operator and test-pod images
  - generate test-pod spec from the template
  - run test-pod
- [x] Run upgrade tests
- [x] Remove the build pass from `hack/test` once all CI jobs are using the pod testing workflow.
- [x] Remove `hack/ci/run`

hongchaodeng:
@hasbro17 
I think it already works today:
https://github.com/coreos/etcd-operator/tree/master/test/pod#run-the-test-pod
Right?

There are two more steps:
- logs
- concurrent jobs

I would recommend to finish the logs first and then switch our existing jenkins jobs to use it.

Currently we have two jenkins jobs and we can combine them into one and solve concurrent jobs in one jenkins job as a second step.

hasbro17:
Yes the `run-test-pod` script already works right now by running each pass serially. So if you run a single pass `e2e` it work. The PR(https://github.com/coreos/etcd-operator/pull/1685) was meant to run the passes concurrently. But as suggested we can tackle that separately.

For now I'll change the `run-test-pod` script to save logs to a file and then update the jenkins jobs for each pass to use the script.

hongchaodeng:
The primary goal is to containerize all tests, incl. unit, upgrade.
After that, we can think about how to improve them. Running them concurrently in one jenkins job is an improvement.

hongchaodeng:
I added the task for collecting logs.
My proposal: build a new program "log collector". Replacing [these lines](https://github.com/coreos/etcd-operator/blob/9bdfb337f92f349ba495e21e379367bddebfee6b/test/pod/run-test-pod#L74-L78) to collect all interested logs, e.g. e2e test, etcd operator.

[...]

PR #1661, opened 2017-11-13
test/pod: timeout if test-pod does not become running

hasbro17:
[skip ci]
ref: #1640 
The test-pod could fail to become running due to a bad image name or some other issue. So a timeout is needed in run-test-pod otherwise it keeps waiting forever.

### DIFF
--- a/test/pod/run-test-pod
+++ b/test/pod/run-test-pod
@@ -52,12 +52,22 @@ PHASE_RUNNING="Running"
 PHASE_SUCCEEDED="Succeeded"
 RETRY_INTERVAL=5
 
-# Wait until pod is running
+# Wait until pod is running or timeout
 echo "Waiting for test-pod to start runnning"
+TIMEOUT=90
+ELAPSED=0
 POD_PHASE=""
 until [ "${POD_PHASE}" == "${PHASE_RUNNING}" ]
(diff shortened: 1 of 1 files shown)


test/pod/run-test-pod
```diff
@@ -52,12 +52,22 @@ PHASE_RUNNING="Running"
...
 
-# Wait until pod is running
+# Wait until pod is running or timeout
 echo "Waiting for test-pod to start runnning"
+TIMEOUT=90
+ELAPSED=0
 POD_PHASE=""
 until [ "${POD_PHASE}" == "${PHASE_RUNNING}" ]
 do
+    if [ "${ELAPSED}" -eq "${TIMEOUT}" ]; then
```
fanminshi: `[ "${ELAPSED}" -ge "${TIMEOUT}" ]`?
use >=? I think it is possible that `ELAPSED` != `TIMEOUT` if `RETRY_INTERVAL` and `TIMEOUT` are not picked correctly.

hasbro17: Good catch!

fanminshi:
lgtm defer @hongchaodeng

hongchaodeng:
lgtm

Issue #1252, opened 2017-07-06
Self-hosted tests should apply anti-affinity to all namespace


hongchaodeng:
ref:
- https://github.com/coreos/etcd-operator/pull/1148
- https://github.com/coreos/etcd-operator/issues/1147

Currently two self-hosted cluster from different test jobs can colocate etcd member onto same node. This can be solved by:
- putting `Ports` info in etcd Pod. Unfortunately, pod-checkpointer also checkpoints such information and on reboot, scheduler will fail to schedule etcd pod because checkpointed etcd pod also sits there.
- apply anti-affinity to all namespace for self-hosted etcd pod. This is blocked since upstream doesn't provide it yet.

PR #1662, opened 2017-11-14
Use podIPs rather than DNS hostnames

jamiehannaford:
First step for #1644 

/cc @hongchaodeng @hasbro17

### DIFF
--- a/pkg/cluster/member.go
+++ b/pkg/cluster/member.go
@@ -24,6 +24,7 @@ import (
 	"github.com/pkg/errors"
 
 	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func (c *Cluster) updateMembers(known etcdutil.MemberSet) error {
@@ -31,6 +32,20 @@ func (c *Cluster) updateMembers(known etcdutil.MemberSet) error {
 	if err != nil {
 		return err
 	}
+
+	peerPodIPs := map[string]string{}
+	if c.cluster.Spec.SelfHosted != nil {
+		pods, err := c.config.KubeCli.CoreV1().Pods(c.cluster.Namespace).List(metav1.ListOptions{})
+		if err != nil {
+			return err
+		}
+		for _, p := range pods.Items {
+			if p.Labels["etcd_cluster"] == c.cluster.Name {
+				peerPodIPs[p.Labels["etcd_node"]] = p.Status.PodIP
+			}
+		}
+	}
+
 	members := etcdutil.MemberSet{}
 	for _, m := range resp.Members {
 		name, err := getMemberName(m, c.cluster.GetName(), c.cluster.Spec.SelfHosted)
@@ -45,13 +60,24 @@ func (c *Cluster) updateMembers(known etcdutil.MemberSet) error {
 			c.memberCounter = ct + 1
 		}
 
-		members[name] = &etcdutil.Member{
+		member := &etcdutil.Member{
 			Name:         name,
 			Namespace:    c.cluster.Namespace,
 			ID:           m.ID,
 			SecurePeer:   c.isSecurePeer(),
 			SecureClient: c.isSecureClient(),
 		}
+
+		if c.cluster.Spec.SelfHosted != nil {
+			podIP, ok := peerPodIPs[m.Name]
+			if !ok {
+				return newFatalError(fmt.Sprintf("could not get podIP for %s member (ID %d)", m.Name, m.ID))
+			}
+			member.PodIP = podIP
+			member.SelfHosted = true
+		}
+
+		members[name] = member
 	}
 	c.members = members
 	return nil
@@ -64,13 +90,19 @@ func (c *Cluster) newMember(id int) *etcdutil.Member {
 		Namespace:    c.cluster.Namespace,
 		SecurePeer:   c.isSecurePeer(),
 		SecureClient: c.isSecureClient(),
+		SelfHosted:   c.cluster.Spec.SelfHosted != nil,
 	}
 }
 
 func podsToMemberSet(pods []*v1.Pod, sc bool) etcdutil.MemberSet {
 	members := etcdutil.MemberSet{}
 	for _, pod := range pods {
-		m := &etcdutil.Member{Name: pod.Name, Namespace: pod.Namespace, SecureClient: sc}
+		m := &etcdutil.Member{
+			Name:         pod.Name,
+			Namespace:    pod.Namespace,
+			SecureClient: sc,
+			PodIP:        pod.Status.PodIP,
+		}
 		members.Add(m)
 	}
 	return members
--- a/pkg/cluster/member.go
+++ b/pkg/cluster/member.go
@@ -16,6 +16,7 @@ package cluster
 
 import (
 	"fmt"
+	"net/url"
 	"strings"
 
 	api "github.com/coreos/etcd-operator/pkg/apis/etcd/v1beta2"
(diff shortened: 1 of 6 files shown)


jamiehannaford:
@hongchaodeng Okay done, PTAL

hongchaodeng:
@etcd-bot ok to test

hongchaodeng:
@hasbro17 
We need to add the set of self-hosted test into the CI today. It only runs in nightly now.
Previously it was removed from CI due to this issue: https://github.com/coreos/etcd-operator/issues/1252
We will probably run it in a single namespace sequentially.

pkg/util/etcdutil/member.go
```diff
@@ -67,7 +73,11 @@ func (m *Member) ListenPeerURL() string {
 }
 
 func (m *Member) PeerURL() string {
-	return fmt.Sprintf("%s://%s:2380", m.peerScheme(), m.Addr())
+	host := m.Addr()
+	if m.SelfHosted {
+		host = "$(POD_IP)"
+	}
+	return fmt.Sprintf("%s://%s:2380", host, m.peerScheme())
```
hongchaodeng: m.peerScheme(), host

pkg/util/etcdutil/member.go
```diff
@@ -42,7 +44,11 @@ func (m *Member) Addr() string {
 
 // ClientURL is the client URL for this member
 func (m *Member) ClientURL() string {
-	return fmt.Sprintf("%s://%s:2379", m.clientScheme(), m.Addr())
+	host := m.Addr()
```
hongchaodeng: Don't change this thing for now?
Let's tackle peer TLS first.

hongchaodeng:
@jamiehannaford 
Self-hosted test is added to the CI now.
Can you rebase? Thanks!

jamiehannaford:
@hongchaodeng @hasbro17 How are you setting up the nodes to run the self-hosted tests? It seems you're deploying multiple etcd pods to the same node. I'm seeing this in the e2e logs:

> initial cluster test-etcd-773gx-0000=http://10.240.0.3:2380,test-etcd-773gx-0001=http://10.240.0.3:2380 has duplicate url

hongchaodeng:
@jamiehannaford 
We have one jenkins job allowing only one test at a time.

hongchaodeng:
That's weird.. Self hosted etcd pods should have anti-affinity.
Let's try your patch locally.

pkg/util/etcdutil/member.go
```diff
@@ -67,7 +69,11 @@ func (m *Member) ListenPeerURL() string {
 }
 
 func (m *Member) PeerURL() string {
-	return fmt.Sprintf("%s://%s:2380", m.peerScheme(), m.Addr())
+	host := m.Addr()
+	if m.SelfHosted {
+		host = "$(POD_IP)"
```
hongchaodeng: In PeerURLPairs(), it calls this method to construct `initialCluster` string.
For pod "0001", it will set other peers to use its own Pod IP even if they have different Pod IP.

We will need to update Member struct with its pod IP when a new Pod and new member is created.

pkg/cluster/member.go
```diff
@@ -64,13 +84,20 @@ func (c *Cluster) newMember(id int) *etcdutil.Member {
...
+		SelfHosted:   c.cluster.Spec.SelfHosted != nil,
 	}
 }
 
-func podsToMemberSet(pods []*v1.Pod, sc bool) etcdutil.MemberSet {
+func podsToMemberSet(c *Cluster, pods []*v1.Pod, sc bool) etcdutil.MemberSet {
 	members := etcdutil.MemberSet{}
 	for _, pod := range pods {
-		m := &etcdutil.Member{Name: pod.Name, Namespace: pod.Namespace, SecureClient: sc}
+		m := &etcdutil.Member{
```
hongchaodeng: Here, we can add each pod's IP into `*Member`

hongchaodeng: Ah.. Seems like it's added. Ignored

pkg/cluster/member.go
```diff
@@ -45,12 +58,19 @@ func (c *Cluster) updateMembers(known etcdutil.MemberSet) error {
 			c.memberCounter = ct + 1
 		}
 
+		podIP, ok := peerPodIPs[m.Name]
```
hongchaodeng: We should **only** get each member's IP here **from Member.PeerURLs**.

jamiehannaford:
@hongchaodeng this is RFR again

pkg/cluster/member.go
```diff
@@ -45,13 +60,24 @@ func (c *Cluster) updateMembers(known etcdutil.MemberSet) error {
...
+		member := &etcdutil.Member{
 			Name:         name,
 			Namespace:    c.cluster.Namespace,
 			ID:           m.ID,
 			SecurePeer:   c.isSecurePeer(),
 			SecureClient: c.isSecureClient(),
 		}
+
+		if c.cluster.Spec.SelfHosted != nil {
+			podIP, ok := peerPodIPs[m.Name]
```
hongchaodeng: Don't list pod and create another map here.
Use the PeerURLs from ListMember response: https://github.com/coreos/etcd/blob/0f4a535c2fb7a2920e13e2e19b9eaf6b2e9285e5/etcdserver/etcdserverpb/rpc.pb.go#L1379-L1388

jamiehannaford: Ah okay, good idea. That's in the format of `name1=podIP,name2=podIP` right? 

So split by `,` then match by member name

hongchaodeng: It's the PeerURLs given to the etcd process.
For example, `PeerURLs[0]=http://10.1.0.1:2380`

hongchaodeng: Assume only one PeerURL.
Also only need to handle self-hosted case.

jamiehannaford: For 3 member clusters, there will be a slice of strings, e.g. `[]string{"http://10.1.0.1:2380","http://10.1.0.2:2380","http://10.1.0.3:2380"}`. 

How does the update method know which URL belongs to which member? One option is for member-0000 to have index 0, member-0001 have index 1, etc. Is it safe to assume the slice is always ordered?

hongchaodeng: I don't catch your question :(
So let me be more detailed about what changes should be made here:

```Go
peerURL := m.PeerURLs[0] // "http://10.1.0.1:2380"
podIP := getIP(peerURL) // "10.1.0.1"
member.PodIP = podIP
```

jamiehannaford: can you give me an example of `resp.Members`?

Is it like:

```
{Name: "member1", PeerURLs: []string{"http://10.1.0.1:2380","http://10.1.0.2:2380","http://10.1.0.3:2380"},
{Name: "member2", PeerURLs: []string{"http://10.1.0.1:2380","http://10.1.0.2:2380","http://10.1.0.3:2380"},
{Name: "member3", PeerURLs: []string{"http://10.1.0.1:2380","http://10.1.0.2:2380","http://10.1.0.3:2380"},
```

or:

```
{Name: "member1", PeerURLs: []string{"http://10.1.0.1:2380"},
{Name: "member2", PeerURLs: []string{"http://10.1.0.2:2380"},
{Name: "member3", PeerURLs: []string{"http://10.1.0.3:2380"},
```

?

hongchaodeng: The second.
It's the PeerURLs that the member exposes.

jamiehannaford: Ah okay, my mistake! I'll fix this tomorrow :)

jamiehannaford:
@hongchaodeng RFR again

pkg/cluster/member.go
```diff
@@ -45,13 +47,24 @@ func (c *Cluster) updateMembers(known etcdutil.MemberSet) error {
...
 			Namespace:    c.cluster.Namespace,
 			ID:           m.ID,
 			SecurePeer:   c.isSecurePeer(),
 			SecureClient: c.isSecureClient(),
 		}
+
+		if c.cluster.Spec.SelfHosted != nil {
+			podIP, err := getPodIP(m.PeerURLs[0])
+			if err != nil {
+				return newFatalError(fmt.Sprintf("error parsing peer URL (%v) for %s: %v", m.PeerURLs[0], m.Name, err))
```
hongchaodeng: Just panic.
FatalError will delete the cluster. No need to return the error.

pkg/util/etcdutil/member.go
```diff
@@ -67,7 +71,15 @@ func (m *Member) ListenPeerURL() string {
...
 func (m *Member) PeerURL() string {
-	return fmt.Sprintf("%s://%s:2380", m.peerScheme(), m.Addr())
+	host := m.Addr()
+	if m.SelfHosted {
+		host = m.PodIP
+	}
+	return m.CustomPeerURL(host)
+}
+
+func (m *Member) CustomPeerURL(host string) string {
```
hongchaodeng: doc

hongchaodeng:
Sorry. This change is breaking compatibility. I failed to realize that before merging. I take the responsibility for this and have to revert this change.

More discussion to follow on https://github.com/coreos/etcd-operator/issues/1617
@jamiehannaford @xiang90 @ericchiang

Issue #1355, opened 2017-08-10
Maintain helm chart


raoofm:
helm chart is outdated, is it being maintained or has a plan to maintain by etcd-operator team.

for example

- etcd is outdated 3.1.8
- etcd-operator is outdated 0.3.3
- it still uses tpr not crd

hongchaodeng:
No.
We don't have the bandwidth to maintain it at the moment.

xiang90:
@raoofm @hongchaodeng 

i labeled this as future. it is a good thing to do, but not high priority for now.

raoofm:
@xiang90 thnx, was just checking as it was confusing to refer the chart and the examples here.

raoofm:
@xiang90 @hongchaodeng I assumed it is being maintained as it was referred here https://github.com/coreos/etcd-operator/blob/master/doc/user/op_guide.md#installation-via-helm without any warning of usage :)

Nitrogenycs:
In case you do not maintain the chart anymore, please put a message in the readme stating so to save people wasting time.

I've spent quite a few hours now getting it working. It breaks completely with kubectl v1.8 due to the use of TPRs and it also does not spawn a cluster, i.e. cluster.enabled=true does not work. This breaks my helm setup.

xiang90:
we never try to maintain it. someone sent us a pr to add it there.

we can:

1. make it clear that it is an externally maintained project
2. remove it from etcd-operator reference.

@Nitrogenycs 

you should file a PR to where the chart is. The maintainer is probably more active there.

raoofm:
@Nitrogenycs you are right that chart is kind of broken or it isn't the right way of doing it. I had spent a day as well thinking that is an approved chart as I found it in etcd-operator guide and figuring out what am I missing as it was in the stable directory. Probably it should be moved to incubator again.

> we never try to maintain it

@xiang90 it would be a great help to have it there. I even gave it a shot but the spec changed quite frequently. I think option 1 is good so that at least if any user wants to maintain it they see the link in the project.

xiang90:
we made it clear that we are not maintaining it in #1663.

PR #1663, opened 2017-11-14
install_guide: add disclaimer for helm chart

hasbro17:
[skip ci]
ref: #1355 
/cc @xiang90 @hongchaodeng

### DIFF
--- a/doc/user/install_guide.md
+++ b/doc/user/install_guide.md
@@ -38,6 +38,7 @@ kubectl delete clusterrolebinding etcd-operator
 ```
 
 ## Installation via Helm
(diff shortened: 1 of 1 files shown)


xiang90:
lgtm

doc/user/install_guide.md
```diff
@@ -38,6 +38,7 @@ kubectl delete clusterrolebinding etcd-operator
 ```
 
 ## Installation via Helm
+**Disclaimer:** The following Helm chart is an external project not maintained by the etcd-operator maintainers. It may not be up to date and so is not the recommended way to install the etcd-operator.
```
xiang90: i do not think we have any recommended way to install etcd operator. although we have one inside tectonic, but it is not public available. so we should remove the last half sentence?

PR #1668, opened 2017-11-15
Add auto-approver controller

jamiehannaford:
This PR adds a new binary which acts as the auto-approve controller for CSRs. It runs as a container inside the operator pod. 

In order for CSRs to be auto-approved, two conditions need to be met:

1. The ServiceName for new etcd pods is set to `etcd-cluster` (this value can be changed if there's a better alternative). If the init container is a member of a Pod with this SA, its CSRs will be approved.
2. The CSR itself needs to specify `etcd` as the Organization in the CSR (again, this can be changed if there's a better alternative).

Most of this has been cribbed from https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/certificates/approver/sarapprove.go.

/cc @hongchaodeng

### DIFF
--- a/cmd/auto-approve-csr/main.go
+++ b/cmd/auto-approve-csr/main.go
@@ -1,0 +1,234 @@
+package main
+
+import (
+	"crypto/x509"
+	"fmt"
+	"reflect"
+	"time"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/coreos/etcd-operator/pkg/util/k8sutil"
+
+	authorization "k8s.io/api/authorization/v1beta1"
+	capi "k8s.io/api/certificates/v1beta1"
+	"k8s.io/client-go/informers"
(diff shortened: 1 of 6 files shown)


xiang90:
@jamiehannaford 

is this specific to etcd operator? or this can be a generic auto-approver thing?

/cc @ericchiang

ericchiang:
I don't think we should give the etcd-operator the ability to approve CSRs. Ideally the upstream csrapprover could just handle these kind of cases https://github.com/kubernetes/kubernetes/issues/55822

jamiehannaford:
Yeah, a more flexible core controller would be preferred. I asked about this in the sig-auth channel and was told to create my own controller instead. I got the impression that the upstream one will always be concerned with kubelet bootstrapping only.

This controller could be made generic if we just match by CN.

ericchiang:
Yeah. I think whitelisting SANs is going to be the hard part there...

PR #1685, opened 2017-11-21
test/pod: add script for running each test pass in a separate namespace

hasbro17:
[skip ci]

The `run-test-passes` script runs each test pass in a separate pod and namespace. It saves the logs for each pass to a file. 
After waiting for all passes to complete it prints out the logs for all failed passes.

/cc @hongchaodeng @fanminshi

### DIFF
--- a/test/pod/run-test-passes
+++ b/test/pod/run-test-passes
@@ -1,0 +1,88 @@
+#!/usr/bin/env bash
+
+# This script runs the each test pass(e2e, e2eslow, upgrade) in separate pods each in its own namespace.
+# It essentially sets up the namespace and calls the run-test-pod script for each pass.
+# The logs for each pass are stored at $(pwd)/_output/test-logs/
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# The ROOT_NAMESPACE should contain the TEST_AWS_SECRET
+# The ROOT_NAMESPACE is appended as a suffix to the namespace for each pass
+: ${ROOT_NAMESPACE:?"Need to set ROOT_NAMESPACE"}
+: ${TEST_AWS_SECRET:?"Need to set TEST_AWS_SECRET"}
+: ${TEST_S3_BUCKET:?"Need to set TEST_S3_BUCKET"}
+: ${TEST_IMAGE:?"Need to set TEST_IMAGE, e.g. gcr.io/coreos-k8s-scale-testing/etcd-operator-tests"}
+
+PASSES=${PASSES:-"e2e e2eslow"}
+OPERATOR_IMAGE=${OPERATOR_IMAGE:-"quay.io/coreos/etcd-operator:dev"}
+
+# Make the output directory for the log files
+LOG_DIR="$(pwd)/_output/test-logs/${ROOT_NAMESPACE}"
+mkdir -p ${LOG_DIR} || true
+
+
+# Cleanup for the per pass namespace
+function cleanup_namespaces {
+    for p in ${PASSES}; do
+        PASS_NAMESPACE=${p}-${ROOT_NAMESPACE}
+        kubectl delete ns ${PASS_NAMESPACE} || true
+    done
+}
+trap cleanup_namespaces EXIT
+
+for p in $PASSES; do
+    echo "${p} pass"
+
+    # Create the pass namespace
+    PASS_NAMESPACE=${p}-${ROOT_NAMESPACE}
+    kubectl create ns ${PASS_NAMESPACE}
+
+    # Copy the aws secret into pass namespace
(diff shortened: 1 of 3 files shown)


test/pod/run-test-pod
```diff
@@ -76,12 +75,3 @@ echo "test-pod logs"
 echo "=============="
 kubectl -n ${TEST_NAMESPACE} logs -f ${POD_NAME}
 echo "=============="
-
-# Check for pod success or failure
```
hongchaodeng: Why delete them?

test/pod/run-test-passes
```diff
@@ -0,0 +1,88 @@
...
+
+for p in $PASSES; do
+    echo "${p} pass"
+
+    # Create the pass namespace
+    PASS_NAMESPACE=${p}-${ROOT_NAMESPACE}
+    kubectl create ns ${PASS_NAMESPACE}
+
+    # Copy the aws secret into pass namespace
+    kubectl -n ${ROOT_NAMESPACE} get secrets ${TEST_AWS_SECRET} -o yaml | sed "s|${ROOT_NAMESPACE}|${PASS_NAMESPACE}|g" | kubectl create -f -
```
hongchaodeng: sed "s|namespace: xxx | namespace: xxx|g"

test/pod/run-test-passes
```diff
@@ -0,0 +1,88 @@
...
+
+    # Run the run-test-pod script for this pass in the pass namespace and save the logs
+    POD_NAME=${p}-test-pod
+	echo "Running test-pod (${POD_NAME}) in namespace (${PASS_NAMESPACE})"
```
hongchaodeng: indent

test/pod/run-test-passes
```diff
@@ -0,0 +1,88 @@
...
+# Cleanup for the per pass namespace
+function cleanup_namespaces {
+    for p in ${PASSES}; do
+        PASS_NAMESPACE=${p}-${ROOT_NAMESPACE}
+        kubectl delete ns ${PASS_NAMESPACE} || true
+    done
+}
+trap cleanup_namespaces EXIT
+
+for p in $PASSES; do
```
hongchaodeng: Better rename `PASSES` inside the script?

test/pod/run-test-passes
```diff
@@ -0,0 +1,88 @@
...
+    kubectl create ns ${PASS_NAMESPACE}
+
+    # Copy the aws secret into pass namespace
+    kubectl -n ${ROOT_NAMESPACE} get secrets ${TEST_AWS_SECRET} -o yaml | sed "s|${ROOT_NAMESPACE}|${PASS_NAMESPACE}|g" | kubectl create -f -
+
+    # Run the run-test-pod script for this pass in the pass namespace and save the logs
+    POD_NAME=${p}-test-pod
+	echo "Running test-pod (${POD_NAME}) in namespace (${PASS_NAMESPACE})"
+	PASSES=${p} \
+    POD_NAME=${POD_NAME} \
```
hongchaodeng: indent:
```
POD_NAME=xxx \
    TEST_NAMESPACE=xxx \
    ...
```

test/pod/run-test-passes
```diff
@@ -0,0 +1,88 @@
...
+    TEST_NAMESPACE=${PASS_NAMESPACE} \
+    TEST_IMAGE=${TEST_IMAGE} \
+    TEST_S3_BUCKET=${TEST_S3_BUCKET} \
+    TEST_AWS_SECRET=${TEST_AWS_SECRET} \
+    OPERATOR_IMAGE=${OPERATOR_IMAGE} \
+    test/pod/run-test-pod > ${LOG_DIR}/${p}_pass.log 2>&1 &
+    echo ""
```
hongchaodeng: what's this for?
Seems unneeded. Remove it?

test/pod/run-test-passes
```diff
@@ -0,0 +1,88 @@
...
+done
+
+echo "Waiting for all passes to complete..."
+wait
```
hongchaodeng: It seems a bad idea to do concurrency control in bash script.
We should use a go program.
But again, I think this scope is bigger than needed. Just do `e2e` initially without caring about concurrency.

hongchaodeng:
@hasbro17 
Let's not deal with two concurrent jobs initially, but only handle `e2e` alone.
We should discuss concurrent jobs and define what to be done in https://github.com/coreos/etcd-operator/issues/1640

hasbro17:
@hongchaodeng Sure let me follow this up on the issue, describe the current status, and we can decide how to proceed from there.

hasbro17:
Closing this until we decide on a strategy for running the phases concurrently.

PR #1688, opened 2017-11-22
test/pod: save test-pod logs to file

hasbro17:
[skip ci]
ref: #1640 

Open to any opinions on making the logfile name/location configurable.

### DIFF
--- a/test/pod/run-test-pod
+++ b/test/pod/run-test-pod
@@ -71,10 +71,10 @@ do
     POD_PHASE=$(kubectl -n ${TEST_NAMESPACE} get pod ${POD_NAME} -o jsonpath='{.status.phase}')
 done
 
-# Print out logs until pod stops running
+# Print out and save logs to a file until the pod stops running
 echo "test-pod logs"
 echo "=============="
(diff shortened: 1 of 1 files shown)


hongchaodeng:
We need to save the "test pod's logs" as well.

hasbro17:
@hongchaodeng can you please clarify. This is saving only the test pod's logs. 
`run-test-pod` runs the test pod and prints out its logs to the screen. Now it saves it to a file as well.

hongchaodeng:
> This is saving only the test pod's logs.

No. It's not. Test pod's log is:
```
kubectl logs test-pod
```

hongchaodeng:
Sorry My mistake.. It's saving the test log now.

hongchaodeng:
OK. LGTM

fanminshi:
lgtm

hongchaodeng:
As a follow-up, can you save operators' logs too?

hasbro17:
@hongchaodeng I can try adding that but I suspect that might introduce concurrency in the bash script in trying to get the logs of multiple pods before they are removed.

hongchaodeng:
Let's add that into umbrella issue :)

PR #1692, opened 2017-11-22
Revert "Use podIPs rather than DNS hostnames"

hongchaodeng:
Reverts coreos/etcd-operator#1662


hongchaodeng:
ref: https://github.com/coreos/etcd-operator/pull/1662#issuecomment-346501816

Issue #1700, opened 2017-11-28
e2e master is failing: TestTLS


hongchaodeng:
https://jenkins-etcd.prod.coreos.systems/view/operator/job/etcd-operator-master/704/consoleText
```
--- FAIL: TestTLS (2.32s)
	tls_test.go:36: exit status 1
```

hongchaodeng:
@fanminshi Can you take a look?

fanminshi:
taking a look.

fanminshi:
I run TestTLS locally and couldn't reproduce. I'll investigate more.

hasbro17:
## How to reproduce
If you run the following tests in the same namespace in the following order then the issue arises.
1. e2e: `TestBackupAndRestore`
2. e2eslow: `TestTLS`

Run the tests with `PASSES=e2e e2eslow` and `E2E_TEST_SELECTOR=TLS|BackupAndRestore`

## Issue with tests
The first test `TestBackupAndRestore` creates the TLS secrets but doesn't remove them when the test is done.
https://github.com/coreos/etcd-operator/blob/master/test/e2e/backup_test.go#L65-L74

When the second test `TestTLS` is run it tries to create the TLS secrets as well.
https://github.com/coreos/etcd-operator/blob/master/test/e2e/e2eslow/tls_test.go#L28-L37
The problem is that the random suffix generated is the same because `rand.Unit64()` generates random values from the default source. So it ends up trying to create TLS secrets of the same name as the previous test. 

And so the `kubectl create secrets` command fails and returns exit 1.
https://github.com/coreos/etcd-operator/blob/master/test/e2e/e2eutil/tls.go#L63-L68

## Fix
The fix is to:
1. Seed rand correctly `rand.Seed(time.Now().UTC().UnixNano())`
2. Clean up the secrets in `TestBackupAndRestore`

Also noticed that the cluster name in `TestBackupAndRestore` needs to be fixed as well
https://github.com/coreos/etcd-operator/blob/master/test/e2e/backup_test.go#L66

hongchaodeng:
@hasbro17 
Great findings!
The fix makes sense to me.
We also need to wrap the error in PrepareTLS() and print error message from os command.

PR #1703, opened 2017-11-28
e2e: fix tls secret creation and cleanup in tests

hasbro17:
fixes: #1700 

The framework now correctly seeds the random number generator.
The backup test cleans up its TLS secrets as well.
The error from executing commands is now wrapped as well in the tls util functions.

### DIFF
--- a/test/e2e/backup_test.go
+++ b/test/e2e/backup_test.go
@@ -33,6 +33,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+func init() {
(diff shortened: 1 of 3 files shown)


hongchaodeng:
lgtm

We should also print the os command output if err != nil: https://golang.org/pkg/os/exec/#Cmd.CombinedOutput

fanminshi:
lgtm

hasbro17:
@hongchaodeng PTAL

hongchaodeng:
lgtm

PR #1708, opened 2017-11-30
docs: editorial pass on conditions_and_events.md

zbwright:
[skip ci]

### DIFF
--- a/doc/user/conditions_and_events.md
+++ b/doc/user/conditions_and_events.md
@@ -1,18 +1,19 @@
-# Status Conditions and Events
+# Status Events and Conditions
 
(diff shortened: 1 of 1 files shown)


doc/user/conditions_and_events.md
```diff
@@ -1,18 +1,19 @@
-# Status Conditions and Events
+# Status Events and Conditions
```
fanminshi: is there a reason for switch the order ?

hongchaodeng: I think it is because we talked events first and conditions second in this doc.
So this is fine to me.

hongchaodeng:
lgtm

doc/user/conditions_and_events.md
fanminshi: kk. sounds good.

fanminshi:
lgtm.

PR #1709, opened 2017-11-30
docs: editorial pass on install_guide.md

zbwright:
[skip ci]

### DIFF
--- a/doc/user/install_guide.md
+++ b/doc/user/install_guide.md
@@ -1,8 +1,8 @@
-# Operation Guide
+# Installation guide
 
-## Setup RBAC
+## Set up RBAC
 
-Set up basic [RBAC rules](./rbac.md) for etcd operator:
+Set up basic [RBAC rules][rbac-rules] for etcd-operator:
 
 ```bash
 $ example/rbac/create_role.sh
@@ -10,13 +10,13 @@ $ example/rbac/create_role.sh
 
 ## Install etcd operator
 
-Create a deployment for etcd operator:
+Create a deployment for etcd-operator:
 
 ```bash
 $ kubectl create -f example/deployment.yaml
 ```
 
-etcd operator will automatically create a Kubernetes Custom Resource Definition (CRD):
+etcd-operator will automatically create a Kubernetes Custom Resource Definition (CRD):
 
 ```bash
 $ kubectl get customresourcedefinitions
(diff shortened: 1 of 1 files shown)


hongchaodeng:
lgtm

doc/user/install_guide.md
```diff
@@ -1,36 +1,38 @@
-# Operation Guide
+# Installation guide
 
-## Setup RBAC
+## Set up RBAC
 
-Set up basic [RBAC rules](./rbac.md) for etcd operator:
+Set up basic [RBAC rules][rbac-rules] for etcd-operator:
```
fanminshi: i wonder what's rule for using `-`.
should we use `etcd operator` or ` etcd-operator` for the doc?
I think either one is fine. but we need to keep everything consistent.

hongchaodeng: Sorry that I merged it fast.
I'm fine with both actually.
Is it OK for both to co-exist?

hongchaodeng: Or maybe we should change "etcd-operator" here back to "etcd operator"?

hongchaodeng: @zbwright 
Can you change "etcd-operator" in this doc back to "etcd operator"? Thanks!

fanminshi: i think we just need to stick with one.

fanminshi: @zbwright what's the best practice in this case?

zbwright: best practice in this instance is for all three OCS operator docs to be consistent - either they are 'vault-operator, etcd-operator, and prometheus-operator', or they are 'vault operator, etcd operator, and prometheus operator'. Product should pick one. If we drop the hyphen, we should probably call it 'the etcd operator', rather than 'etcd-operator'. I'm leaving all this fine tuning for a later date. we'll drop the hyphen in all these docs for now.

fanminshi: sounds good!

raoofm: @fanminshi I don't see a vault operator anywhere in coreos. Is it in any other repository?

hongchaodeng: @raoofm 
Vault operator is private.
@zbwright is our internal technical editor so she needed to consider all products as a whole.

raoofm: @hongchaodeng thank you

PR #1738, opened 2017-12-06
*: run logcollector in container from pre-built image

hasbro17:
The logcollector is now pre-built as an image at `gcr.io/coreos-k8s-scale-testing/logcollector` and `run-test-pod` now runs the logcollector in container from that image.


hasbro17:
@hongchaodeng PTAL

hongchaodeng:
LGTM

PR #1753, opened 2017-12-11
test/pod: decouple pod template and run script

hasbro17:
The test-pod spec is now generated from the template in `hack/ci/run2` and passed to `run-test-pod` to use.


hongchaodeng:
LGTM

Can you follow up:
- rename ci/run2 to ci/run_e2e
- rename test-pod.yaml to test-pod-tmpl.yaml

PR #1758, opened 2017-12-12
run_e2e: default env variables

hasbro17:
Defaulting all possible env values required by the test-pod since the CI job might not specify all envs if it is running a particular test.

For e.g in upgrade tests `BUILD_IMAGE` would be false but `OPERATOR_IMAGE` will be unset since it is not needed. And our script does not allow unset variables.


hasbro17:
Testing this branch against upgrade and master e2e tests.

hongchaodeng:
LGTM

PR #1759, opened 2017-12-12
hack: remove ci/run and build step from hack/test

hasbro17:
ref: #1640 
With `hack/ci/run_e2e` and `test/pod/run-test-pod` the older scripts are no longer needed.


hongchaodeng:
lgtm

PR #1762, opened 2017-12-13
controller: add cluster spec validation in restore operator

fanminshi:
add spec validation in restore operator.


fanminshi:
cc/ @hongchaodeng

hongchaodeng:
LGTM

PR #1765, opened 2017-12-13
test/pod: update README for upgrade test variables

hasbro17:
[skip ci]

The `test/pod/test-pod-tmpl.yaml` has env variables related to the upgrade tests that need to be set even if the upgrade tests are not run. The README has been updated to show that.


hongchaodeng:
lgtm

Issue #1320, opened 2017-07-24
Put readiness check into pod


hongchaodeng:
Currently we do readiness check in operator in each reconcile. This could pause operator for 5s in each health check if failed, and make reconcile's predictability worse. We could put the readiness check into pod spec.

hasbro17:
After the readiness check is part of the etcd pod's spec, we can update the EtcdCluster member status of ready/unready members based on each pod's ready condition.

PR #1767, opened 2017-12-13
*: use pod readiness check to update member status

hasbro17:
ref #1320 

Added readiness probe to etcd pod. It's the same as the liveliness probe but the time period for readiness is 10s, shorter than the liveliness 60s.

The operator now checks if all member pods are Ready while updating the member status for ready/unready members.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -26,7 +26,6 @@ import (
 	api "github.com/coreos/etcd-operator/pkg/apis/etcd/v1beta2"
 	"github.com/coreos/etcd-operator/pkg/debug"
 	"github.com/coreos/etcd-operator/pkg/generated/clientset/versioned"
-	"github.com/coreos/etcd-operator/pkg/util"
 	"github.com/coreos/etcd-operator/pkg/util/etcdutil"
 	"github.com/coreos/etcd-operator/pkg/util/k8sutil"
 	"github.com/coreos/etcd-operator/pkg/util/retryutil"
@@ -280,7 +279,7 @@ func (c *Cluster) run() {
 				c.logger.Errorf("failed to reconcile: %v", rerr)
 				break
 			}
-			c.updateMemberStatus(c.members, k8sutil.GetPodNames(running))
+			c.updateMemberStatus(running)
 			if err := c.updateCRStatus(); err != nil {
 				c.logger.Warningf("periodic update CR status failed: %v", err)
 			}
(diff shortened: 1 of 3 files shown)


pkg/cluster/cluster.go
```diff
@@ -433,14 +432,28 @@ func (c *Cluster) pollPods() (running, pending []*v1.Pod, err error) {
 	return running, pending, nil
 }
 
-func (c *Cluster) updateMemberStatus(members etcdutil.MemberSet, running []string) {
+func (c *Cluster) updateMemberStatus(members etcdutil.MemberSet, running []*v1.Pod) {
```
hongchaodeng: pass in `(ready, running []string)`

hongchaodeng: Sorry. I misunderstand the code... Disregard above.

pkg/util/k8sutil/pod_util.go
```diff
@@ -67,6 +68,26 @@ func containerWithRequirements(c v1.Container, r v1.ResourceRequirements) v1.Con
 	return c
 }
 
+func etcdReadinessProbe(isSecure bool) *v1.Probe {
+	// etcd pod is ready only if a linearizable get succeeds.
+	cmd := "ETCDCTL_API=3 etcdctl get foo"
```
hongchaodeng: Unify this with liveness probe code?

hasbro17: The initial delay, timeout, period, and failure threshold are smaller for readiness probe than they are for liveness probe. 
Or do you still want to unify them into a common function and specify those values are params?

hongchaodeng: Compose them:
```
pb := NewProber(xxx)
SetProberInitialDelay(pb, 5)
...
```

hasbro17:
@hongchaodeng PTAL. I don't know if we need explicit setter methods for setting the probe delays and timeouts. Can make it less verbose by just setting them  directly:
```
readinessProbe := newEtcdProbe(cs.TLS.IsSecureClient())
readinessProbe.InitialDelaySeconds = 5
. . .
```

hongchaodeng:
> I don't know if we need explicit setter methods for setting the probe delays and timeouts.

Either is fine.

hasbro17:
Going to set them directly then.

hasbro17:
fixed. PTAL.

hongchaodeng:
Initial Delay too large? Maybe 1s?
```
  readinessProbe:
      exec:
        command:
        - /bin/sh
        - -ec
        - ETCDCTL_API=3 etcdctl get foo
      failureThreshold: 1
      initialDelaySeconds: 5
      periodSeconds: 5
      successThreshold: 1
      timeoutSeconds: 5
```

hasbro17:
Changed initial delay to 1s. Tried it manually. Can't really tell if it makes too much of a difference though so it's fine either way.

hongchaodeng:
failureThreshold: 1 -> 3?
I tried it and it seemed a little volatile at the beginning switching to unready.

hongchaodeng:
LGTM

Issue #1721, opened 2017-12-02
[feature/backup] EtcdBackup: take etcd endpoint


hongchaodeng:
Currently EtcdBackup takes in "ClusterName" which is the name of EtcdCluster. This is coupled with EtcdCluster API.

It'd better take an etcd endpoint instead. This would also allow to take backup on. external, non-k8s etcd cluster.

hongchaodeng:
fixed

PR #1772, opened 2017-12-14
*: etcdbackup spec uses etcd endpoints to retrieve snapshot 

fanminshi:
etcdbackup spec now takes in etcd endpoints to retrieve snapshot.
etcd backup operator takes in etcd endpoints and retrieves snapshot with the endpoint that point to a member with max revision.

ref: https://github.com/coreos/etcd-operator/issues/1721


fanminshi:
@etcd-bot retest this please reason: need to run ci after removing [skip ci] tag.

pkg/apis/etcd/v1beta2/backup_types.go
```diff
@@ -47,8 +47,8 @@ type EtcdBackup struct {
 
 // BackupSpec contains a backup specification for an etcd cluster.
 type BackupSpec struct {
-	// ClusterName is the etcd cluster name.
-	ClusterName string `json:"clusterName,omitempty"`
+	// EtcdEndpoints specifies the etcd cluster endpoints.
```
xiang90: let us document how will the endpoint be used. will the snapshot come from the first available endpoint? or it will get from the endpoints with the largest revision? can i put endpoints of different clusters? what happens if i do?

pkg/backup/backup_manager.go
```diff
@@ -76,54 +73,33 @@ func (bm *BackupManager) SaveSnap(s3Path string) error {
...
-		return nil, 0, err
-	}
-
-	var pods []*v1.Pod
-	for i := range podList.Items {
-		pod := &podList.Items[i]
-		if pod.Status.Phase == v1.PodRunning {
-			pods = append(pods, pod)
-		}
+	client, rev := getEndpointWithMaxRev(bm.endpoints, bm.etcdTLSConfig)
```
hongchaodeng: We probably don't need such complexity of sorting.
Just pass the endpoints as the same as `--endpoints` param?

fanminshi: so let etcd client decides then. that makes sense.

hongchaodeng: Cool. 👍

pkg/apis/etcd/v1beta2/backup_types.go
fanminshi: talking to hongqiao. it seems like we just pass endpoints to etcd client and let etcd client decide which endpoint to get etcd snapshot from.

xiang90: that changes the previous logic. the previous logic tries to snapshot from the server that has the largest revision. 

why do we want to make it less smarter?

pkg/backup/backup_manager.go
xiang90: why dont we need this logic? how will etcd client library help us to get the same thing done?

pkg/apis/etcd/v1beta2/backup_types.go
fanminshi: @xiang90 should we keep the getting snapshot logic as simple as possible first? 

user can still achieve saving the snapshot with max revision by giving an endpoint that points the max member.

xiang90: how will the user be able to do that? i do not think the goal is to make the logic as simple as possible. the goal is to decouple the backup-operator from the etcd-operator logic by not taking the cluster name.

hongchaodeng: Here's the use case I have in mind.
Instead of getting all transient etcd pods' IP, initially user will give the load balancer URL to us. It doesn't make sense to retrieve the max revision.

xiang90: > Instead of getting all transient etcd pods' IP, initially user will give the load balancer URL to us. It doesn't make sense to retrieve the max revision.

If that is the case, the snapshot logic has to be changed. we need to:

1. do a l-get to ensure the member we are interacting with is up-to-date
2. do a snapshot

whatever we choose to do here, we should not do that in the PR. we should first finish the decouple work.

hongchaodeng: The issue is that if user give us a load balancer URL, then there is a race condition that you first get a revision of X, but then you get a snapshot of Y.

xiang90: > The issue is that if user give us a load balancer URL, then there is a race condition that you first get a revision of X, but then you get a snapshot of Y.

no. that is not how etcd client works. the client wont change connection between requests.

hongchaodeng: Cool. It seems I have made the wrong assumption.
So it's fine then.

fanminshi: `how will the user be able to do that? `
I am thinking that user manually figure out which member has highest rev using etcdctl, but that seems tedious.

xiang90: > I am thinking that user manually figure out which member has highest rev using etcdctl, but that seems tedious.

right. then i would rather to request the snapshot with etcdctl and pipe that into s3util to upload :P

hongchaodeng: Actually the below code has the race I mentioned above..
So needs to fix that too.

pkg/backup/backup_manager.go
fanminshi: I guess the main concern here is should etcd back operator be smart on what endpoint to choose or not. let's figure that out first @hongchaodeng  @xiang90 .

pkg/apis/etcd/v1beta2/backup_types.go
xiang90: > Actually the below code has the race I mentioned above..

just to be clear, the goal of the sorting is to not get a super out of date snapshot rather than trying to ensure we get the most recent snapshot. an endpoint with smaller revision can always increase its revision faster than the one was the largest. we do not need to handle that race, and we cannot.

pkg/backup/backup_manager.go
fanminshi: it seems like we are going to keep the maxRev logic. i'll do that then.

fanminshi:
merge when green.

fanminshi:
cc/ @xiang90

test/e2e/backup_test.go
```diff
@@ -95,8 +130,12 @@ func testEtcdBackupOperatorForS3Backup(t *testing.T) string {
 		t.Fatalf("failed to create 3 members etcd cluster: %v", err)
 	}
 
+	endpints, err := getEndpoints(f.KubeClient, true, f.Namespace, clusterName)
```
hasbro17: endpints ==> endpoints

pkg/apis/etcd/v1beta2/backup_types.go
```diff
@@ -47,8 +47,11 @@ type EtcdBackup struct {
...
+	// etcd backup operator determines a endpoint that points
```
xiang90: when multiple endpoints are given, the backup operator retrieves the backup from the endpoint that has the most up-to-date state. the given endpoints must belong to the same etcd cluster.

fanminshi:
all fixed PTAL

pkg/apis/etcd/v1beta2/backup_types.go
```diff
@@ -47,8 +47,11 @@ type EtcdBackup struct {
 
 // BackupSpec contains a backup specification for an etcd cluster.
 type BackupSpec struct {
-	// ClusterName is the etcd cluster name.
-	ClusterName string `json:"clusterName,omitempty"`
+	// EtcdEndpoints specifies the endpoints of an etcd cluster.
+	// when multiple endpoints are given, the backup operator retrieves
```
xiang90: when -> When

the given -> The given

xiang90:
the doc part looks good to me.

pkg/apis/etcd/v1beta2/backup_types.go
fanminshi: fixed!

hongchaodeng:
Minor comments change. Merging.

Issue #1770, opened 2017-12-14
EtcdRestore should take in EtcdCluster reference


hongchaodeng:
Currently, EtcdRestore takes in ClusterSpec in order to restore EtcdCluster. This incurs a problem when the configuration of EtcdCluster to restore is huge, for example labels, ownerRef, TLS policy, pod policy, etc.

In order to simplify the process and reduce duplication, it'd better put a reference to EtcdCluster in EtcdRestore instead:
```yaml
spec:
  s3: ...
  etcdCluster:
    name: example-etcd-cluster
    namespace: etcd-services
```

TODO:
- [x] restore-operator should ensure that there are no etcd pods and services in its namespace before recreating the restored `EtcdCluster` CR.
- [x] update the restore-operator walkthrough

hasbro17:
TODO: Update the restore-operator walkthrough doc as well.

hasbro17:
Now that the restore-operator deletes the reference `EtcdCluster` it should also first delete any existing etcd pods and services in its namespace before creating the new restored `EtcdCluster` CR.
Adding a TODO above.

PR #1778, opened 2017-12-15
*: add EtcdClusterRef to EtcdRestore

hasbro17:
ref: #1770 

The restore-operator can now restore a failed cluster specified by the EtcdClusterRef `spec.etcdCluster`.

The restore workflow is as follows:
- Fetch and delete the reference EtcdCluster
- Create the restored EtcdCluster with the same name and spec as the reference cluster.
- Follow the remaining restore process as before(create seed member etc)

The backup and restore e2e test now use the same EtcdCluster with the test workflow:
- Create a TLS enabled cluster
- Run backup test and create backup
- Fail the cluster by deleting its pods and services
- Run the restore test to restore the cluster

/cc @hongchaodeng


hongchaodeng:
@hasbro17 
Awesome! Thanks for the hard work!
LGTM

PR #1780, opened 2017-12-15
pkg/backup: change getEndpointWithMaxRev to getClientWithMaxRev

fanminshi:
Since getEndpointWithMaxRev already creates etcdclient in order to determine the endpoint with max rev, it can smartly return the etcdclient with max rev instead. 
Then backupmanager can use that client directly to retrieve snapshot instead of create a new client based on the max rev endpoint that getEndpointWithMaxRev originally returns.

### DIFF
--- a/pkg/backup/backup_manager.go
+++ b/pkg/backup/backup_manager.go
@@ -17,7 +17,6 @@ package backup
 import (
 	"context"
 	"crypto/tls"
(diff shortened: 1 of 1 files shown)


fanminshi:
cc/ @hongchaodeng

hongchaodeng:
LGTM

PR #1785, opened 2017-12-16
*: restore-operator deletes reference EtcdCluster completely before restore

hasbro17:
ref: #1770 

The restore-operator will now delete the reference EtcdCluster and its pods and services completely before creating the restored EtcdCluster.
With this change the e2e test no longer needs to cleanup the pods and services of the reference cluster first. 

Additionally `EtcdClusterRef` has also been simplified by removing the namespace field. The namespace for the reference cluster is always assumed to be the same namespace as the restore-operator.
Support for restoring from an EtcdCluster in a different namespace can be added later.

/cc @hongchaodeng

PR #1789, opened 2017-12-18
*: remove namespace from EtcdClusterRef

hasbro17:
ref: #1770 

Part 1/2 of #1785 

`EtcdClusterRef` has been simplified by removing the namespace field. The namespace for the reference cluster is always assumed to be the same namespace as the restore-operator.

Support for restoring from an EtcdCluster in a different namespace can be added later.

/cc @hongchaodeng


hongchaodeng:
LGTM

PR #1790, opened 2017-12-18
*: restore-operator deletes EtcdClusterRef completely

hasbro17:
ref: #1770

Part 2/2 of #1785

The restore-operator will now delete the reference EtcdCluster and its pods and services completely before creating the restored EtcdCluster.
With this change, the e2e test no longer needs to clean up the pods and services of the reference cluster first.

/cc @hongchaodeng


hongchaodeng:
lgtm

Issue #1763, opened 2017-12-13
EtcdBackup/EtcdRestore API groomming


hongchaodeng:
EtcdBackup was originally decoupled from EtcdCluster and had many legacy parts. We did that for speed.

We need to further improve EtcdBackup/EtcdRestore API:
- [x] take in etcd endpoint instead of clusterName: https://github.com/coreos/etcd-operator/issues/1721
   - PR: https://github.com/coreos/etcd-operator/pull/1772, https://github.com/coreos/etcd-operator/pull/1779
- [x] write to S3 path instead of relying on combinations of clusterName and internal scheme
  - PR: https://github.com/coreos/etcd-operator/pull/1774, https://github.com/coreos/etcd-operator/pull/1775
- [x] EtcdRestore should take in EtcdCluster ref: https://github.com/coreos/etcd-operator/issues/1770
- [x] expose etcd version and revision in backup status.
  - PR: https://github.com/coreos/etcd-operator/pull/1792
- [x] EtcdRestore should take in BackupStorageType.
  - Fields added. Still needs validation.
  - PR: https://github.com/coreos/etcd-operator/pull/1794

hongchaodeng:
I'm removing the "delete CR would delete backup" item from this issue.
Because it is not obvious how that would be a safe way -- how would I know if a CR was deleted and its backup still existed?
We'd better create an new issue for this specifically.

xiang90:
> Because it is not obvious how that would be a safe way -- how would I know if a CR was deleted and its backup still existed?
We'd better create an new issue for this specifically.

agreed. can you create a new issue now?

PR #1794, opened 2017-12-19
*: validate BackupStorageType in restore operator

fanminshi:
ref: https://github.com/coreos/etcd-operator/issues/1763

When RestoreSpec specifies  backupStorageType, restore operator needs to check if backupStorageType matches the RestoreSource.

### DIFF
--- a/pkg/controller/restore-operator/http.go
+++ b/pkg/controller/restore-operator/http.go
@@ -73,12 +73,18 @@ func (r *Restore) serveBackup(w http.ResponseWriter, req *http.Request) error {
 
 	logrus.Infof("serving backup for restore CR %v", restoreName)
 	cr := v.(*api.EtcdRestore)
-	restoreSource := cr.Spec.RestoreSource
-	var backupReader reader.Reader
-	var path string
 
-	switch {
-	case restoreSource.S3 != nil:
+	var (
+		backupReader reader.Reader
+		path         string
+	)
+
+	switch cr.Spec.BackupStorageType {
+	case api.BackupStorageTypeS3:
+		restoreSource := cr.Spec.RestoreSource
+		if restoreSource.S3 == nil {
+			return errors.New("empty s3 restore source")
+		}
 		s3RestoreSource := restoreSource.S3
 		if len(s3RestoreSource.AWSSecret) == 0 || len(s3RestoreSource.Path) == 0 {
 			return errors.New("invalid s3 restore source field (spec.s3), must specify all required subfields")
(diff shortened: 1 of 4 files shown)


pkg/controller/restore-operator/http.go
```diff
@@ -73,12 +73,18 @@ func (r *Restore) serveBackup(w http.ResponseWriter, req *http.Request) error {
...
+		backupReader reader.Reader
+		path         string
+	)
+
+	switch cr.Spec.BackupStorageType {
+	case api.BackupStorageTypeS3:
+		restoreSource := cr.Spec.RestoreSource
+		if restoreSource.S3 == nil {
+			return errors.New("empty s3 restore source")
+		}
```
hongchaodeng:
```
default:
  return ("unknown backup storage type (%s) for restore CR (%s)
```

hongchaodeng:
LGTM
Please also change the default error case.

pkg/controller/restore-operator/http.go
fanminshi: @hongchaodeng do we want to print everything about restore CR? that can be very verbose.

hongchaodeng: just the name

fanminshi: kk
i have this
`return fmt.Errorf("unknown backup storage type (%s) for restore-name (%v)", cr.Spec.BackupStorageType, restoreName)`

hongchaodeng: restore-name -> restore CR

fanminshi:
merge when green.

pkg/controller/restore-operator/http.go
```diff
@@ -71,14 +71,20 @@ func (r *Restore) serveBackup(w http.ResponseWriter, req *http.Request) error {
 		return fmt.Errorf("no restore CR found for restore-name (%v)", restoreName)
 	}
 
-	logrus.Infof("serving backup for restore CR %v", restoreName)
+	logrus.Infof("serving backup for restore-name (%v)", restoreName)
```
hongchaodeng: restore-name -> restore CR

:)

pkg/controller/restore-operator/http.go
fanminshi: @hongchaodeng  it should be `restore-name` to be consistent. code uses `restore-name` in other places too.

pkg/controller/restore-operator/http.go
fanminshi: hmm, we have mixed `restore-name` and `restore CR` within the code base. which one to use then?

pkg/controller/restore-operator/http.go
hongchaodeng: That's based on different contexts:
```
		return fmt.Errorf("failed to get restore CR for restore-name (%v): %v", restoreName, err)
...
		return fmt.Errorf("no restore CR found for restore-name (%v)", restoreName)
```

pkg/controller/restore-operator/http.go
hongchaodeng: It's for different contexts.
For this context, please use `restore CR`.

fanminshi: alright then

fanminshi:
fixed!

hongchaodeng:
lgtm

fanminshi:
merge when green.

PR #1815, opened 2018-01-02
hack/run_e2e: support skipping tests

hasbro17:
With `PASSES=""` hack/run_e2e can be used to skip tests and only build images.

### DIFF
--- a/hack/ci/run_e2e
+++ b/hack/ci/run_e2e
@@ -10,7 +10,6 @@ export PATH=$GOROOT/bin:$PATH
 go version
 kubectl version
 
-: ${PASSES:?"Need to set PASSES"}
 : ${TEST_NAMESPACE:?"Need to set TEST_NAMESPACE"}
 : ${TEST_S3_BUCKET:?"Need to set TEST_S3_BUCKET"}
 : ${TEST_AWS_SECRET:?"Need to set TEST_AWS_SECRET"}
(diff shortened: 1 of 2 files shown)


hack/ci/run_e2e
```diff
@@ -40,6 +39,11 @@ if [[ ${BUILD_E2E} == "true" ]]; then
   TEST_IMAGE=${TEST_IMAGE} hack/build/e2e/docker_push
 fi
 
+if [ -z "${PASSES-}" ]; then
+	echo "No PASSES specified. Skipping tests"
+  exit 0
```
hongchaodeng: indent

hack/ci/run_e2e
```diff
@@ -40,6 +39,11 @@ if [[ ${BUILD_E2E} == "true" ]]; then
   TEST_IMAGE=${TEST_IMAGE} hack/build/e2e/docker_push
 fi
 
+if [ -z "${PASSES-}" ]; then
+  echo "No PASSES specified. Skipping tests"
```
hongchaodeng: Might want to use tab instead?

hasbro17: I can fix this and other indents in a follow-up PR. Testing the master jenkins job right now with `PASSES=""`

hongchaodeng: No worry about it.
Just remind that when we have PR for jenkins scripts.

PR #1705, opened 2017-11-30
backup: custom S3 endpoint support

givia:
this enables backup on other implementations of S3, such as ceph and minio.

PR #1820, opened 2018-01-03
S3 custom endpoint

alexandrem:
This adds the capability to upload backup files to a compatible S3 service api endpoint, such as OpenStack Swift or Minio.

The endpoint details are defined in a new secret file named "endpoints" which is appended to the existing operator AWS secret file in Kubernetes.

The current expected format of this file is in TOML. We could change this to something else if it makes more sense.

The endpoints file can support other options, such as disabling SSL cert verification and have room for future options if needed.

Example:

backup.yaml
```
kind: "EtcdBackup"
metadata:
  name: etcd-cluster-backup
  namespace: test
spec:
  etcdEndpoints:
  - http://etcd-cluster-client:2379
  storageType: S3
  s3:
    path: test/cluster-backup
    awsSecret: aws
```

endpoints secret:
```
[s3]
endpoint = "http://minio-minio-svc.default.svc.cluster.local:9000/"
disable_ssl = true
```


alexandrem:
Will require updating the example configs in https://github.com/coreos/etcd-operator/blob/master/doc/user/walkthrough/backup-operator.md#setup-aws-secret

hongchaodeng:
Hi @alexandrem 

Thanks for submitting the PR. Can you create an issue for this feature and describe your use case for this first? This will keep it in track and we would regularly run through the list for release planning.

alexandrem:
I noticed someone had developed a similar patch in #1705

That one contained a nice e2e test. We should merge efforts.

hongchaodeng:
> We should merge efforts.

That sounds perfect. Thanks for pushing this forward.

givia:
My e2e test (with minor changes) is applicable to this pr. I think we should add e2e test from my pr to this one and keep on here.

hongchaodeng:
An e2e test would be wasting since much of the test logic would be duplicate to existing S3 backup test.

For this change, it only needs to test the difference between normal AWS S3 and custom endpoints:
- the new endpoints parsing code
- the new S3 client with custom endpoints

A unit test with integration flag/env would be good enough. Like we did previously, we can set up minion in background before testing and passes credentials in.

hongchaodeng:
If decided, @fanminshi can help with minio setup -- we had done this before.

drewwells:
We would also like to see this feature land. Oracle Cloud also has an s3 compatible object store: https://docs.us-phoenix-1.oraclecloud.com/Content/Object/Tasks/s3compatibleapi.htm#AmazonS3CompatibilityAPI

It would be beneficial to customize the URL so non-aws ones can be used. It helps reach the  backup/restore operator design goals of supporting other non-S3 backup solutions.

xiang90:
@drewwells @givia @alexandrem 

I believe https://github.com/coreos/etcd-operator/pull/1705 is closer to what we want, and closer to get merged. 

Can you please put some effort onto that PR? I also want to get it merged soon.

xiang90:
close in favor of https://github.com/coreos/etcd-operator/pull/1855

PR #1844, opened 2018-01-10
document use of clientTLSSecret

khrisrichardson:
Document `clientTLSSecret` attribute in `EtcdBackup` spec.

### DIFF
--- a/doc/user/walkthrough/backup-operator.md
+++ b/doc/user/walkthrough/backup-operator.md
@@ -53,8 +53,43 @@ the secret will be used later to save etcd backup into S3.
 
 2. Create secret `aws`:
 
+    ```sh
+    kubectl create secret generic aws \
+        --from-file=$AWS_DIR/credentials \
+        --from-file=$AWS_DIR/config
     ```
-    kubectl create secret generic aws --from-file=$AWS_DIR/credentials --from-file=$AWS_DIR/config
+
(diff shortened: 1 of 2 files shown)


doc/user/walkthrough/backup-operator.md
```diff
@@ -65,6 +100,7 @@ Create EtcdBackup CR:
 ```sh
 sed -e 's|<full-s3-path>|jenkins-testing-operator/etcd.backup|g' \
     -e 's|<aws-secret>|aws|g' \
+    -e 's|<etcd-client-secret>|etcd-client|g' \
```
hongchaodeng: The current `example-etcd-cluster` is an non-TLS cluster.

khrisrichardson: Hi @hongchaodeng. Thanks for the review. Would you suggest that I create another example cluster YAML with TLS, and add a separate sed command for that example?

xiang90:
@khrisrichardson kindly ping.

khrisrichardson:
Hi @xiang90 I discovered several places in the docs where the distinction between non-TLS and TLS clusters would need to be made, and unfortunately that needs to be postponed while I work on upgrading our Tectonic clusters from 1.7 to 1.8. I should be able to resume this within a week.

xiang90:
@khrisrichardson thanks for the update!

PR #1861, opened 2018-01-17
Add PersistVolumne Support

rjtsdl:
Successor of #1695 

Copy the plan:

 - [x] add PVC for etcd pod (#1861)
 - [ ] handle node restart case
        Behavior: first, it will has the eviction behavior. After that, when the nodes back. The **Unknown** pod(s) would disappear in the pod list. 
 - [ ] handle node partition case
 - [ ] handle node eviction case
        Behavior: the pod(s) scheduled on the evicted node will become **Unknown** after a certain period. Then another pod(s) will be scheduled on the healthy nodes to reach the desired replica count. The **Unknown** will not disappear though.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -328,6 +328,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "ff4a4b53ef454edf0ae2ec3bc497baf74d561af328faa123eb472d87954a906c"
+  inputs-digest = "b19769641bf117041022e3de0c490313047f3b7aaef54e00a7354573fff9b738"
   solver-name = "gps-cdcl"
   solver-version = 1
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -328,6 +328,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "b19769641bf117041022e3de0c490313047f3b7aaef54e00a7354573fff9b738"
+  inputs-digest = "ff4a4b53ef454edf0ae2ec3bc497baf74d561af328faa123eb472d87954a906c"
   solver-name = "gps-cdcl"
   solver-version = 1
--- a/pkg/apis/etcd/v1beta2/cluster.go
+++ b/pkg/apis/etcd/v1beta2/cluster.go
@@ -141,6 +141,8 @@ type PodPolicy struct {
 
 	// PersistentVolumeClaimSpec is the spec to describe PVC for the etcd container
 	// This field is optional. If no PVC spec, etcd container will use emptyDir as volume
+	// Note. This feature is in alpha stage. It is currently only used as non-stable storage,
+	// not the stable storage. Future work need to make it used as stable storage.
 	PersistentVolumeClaimSpec *v1.PersistentVolumeClaimSpec `json:"persistentVolumeClaimSpec,omitempty"`
 }
 
--- a/pkg/apis/etcd/v1beta2/cluster.go
+++ b/pkg/apis/etcd/v1beta2/cluster.go
@@ -141,10 +141,6 @@ type PodPolicy struct {
 
 	// PersistentVolumeClaimSpec ...
 	PersistentVolumeClaimSpec *v1.PersistentVolumeClaimSpec `json:"persistentVolumeClaimSpec,omitempty"`
-
-	// By default, kubernetes will mount a service account token into the etcd pods.
-	// AutomountServiceAccountToken indicates whether pods running with the service account should have an API token automatically mounted.
-	AutomountServiceAccountToken *bool `json:"automountServiceAccountToken,omitempty"`
 }
 
 func (c *ClusterSpec) Validate() error {
--- a/pkg/apis/etcd/v1beta2/cluster.go
+++ b/pkg/apis/etcd/v1beta2/cluster.go
@@ -139,7 +139,8 @@ type PodPolicy struct {
 	// This field cannot be updated.
 	EtcdEnv []v1.EnvVar `json:"etcdEnv,omitempty"`
 
-	// PersistentVolumeClaimSpec ...
+	// PersistentVolumeClaimSpec is the spec to describe PVC for the etcd container
+	// This field is optional. If no PVC spec, etcd container will use emptyDir as volume
 	PersistentVolumeClaimSpec *v1.PersistentVolumeClaimSpec `json:"persistentVolumeClaimSpec,omitempty"`
 }
 
--- a/pkg/apis/etcd/v1beta2/cluster.go
+++ b/pkg/apis/etcd/v1beta2/cluster.go
@@ -138,6 +138,13 @@ type PodPolicy struct {
 	// bootstrap the cluster (for example `--initial-cluster` flag).
 	// This field cannot be updated.
 	EtcdEnv []v1.EnvVar `json:"etcdEnv,omitempty"`
+
+	// PersistentVolumeClaimSpec ...
+	PersistentVolumeClaimSpec v1.PersistentVolumeClaimSpec `json:"persistentVolumeClaimSpec,omitempty"`
+
+	// By default, kubernetes will mount a service account token into the etcd pods.
+	// AutomountServiceAccountToken indicates whether pods running with the service account should have an API token automatically mounted.
+	AutomountServiceAccountToken *bool `json:"automountServiceAccountToken,omitempty"`
 }
 
 func (c *ClusterSpec) Validate() error {
--- a/pkg/apis/etcd/v1beta2/cluster.go
+++ b/pkg/apis/etcd/v1beta2/cluster.go
@@ -140,7 +140,7 @@ type PodPolicy struct {
 	EtcdEnv []v1.EnvVar `json:"etcdEnv,omitempty"`
 
 	// PersistentVolumeClaimSpec ...
-	PersistentVolumeClaimSpec v1.PersistentVolumeClaimSpec `json:"persistentVolumeClaimSpec,omitempty"`
+	PersistentVolumeClaimSpec *v1.PersistentVolumeClaimSpec `json:"persistentVolumeClaimSpec,omitempty"`
 
 	// By default, kubernetes will mount a service account token into the etcd pods.
 	// AutomountServiceAccountToken indicates whether pods running with the service account should have an API token automatically mounted.
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -254,6 +254,13 @@ func (c *Cluster) run() {
 				continue
 			}
 
+			pvcs, err := c.pollPVCs()
+			if err != nil {
+				c.logger.Errorf("failed to poll pvcs: %v", err)
+				reconcileFailed.WithLabelValues("failed to poll vcs").Inc()
+				continue
+			}
+
 			if len(pending) > 0 {
 				// Pod startup might take long, e.g. pulling image. It would deterministically become running or succeeded/failed later.
 				c.logger.Infof("skip reconciliation: running (%v), pending (%v)", k8sutil.GetPodNames(running), k8sutil.GetPodNames(pending))
@@ -274,7 +281,7 @@ func (c *Cluster) run() {
 					break
 				}
 			}
-			rerr = c.reconcile(running)
+			rerr = c.reconcile(running, pvcs)
 			if rerr != nil {
 				c.logger.Errorf("failed to reconcile: %v", rerr)
 				break
@@ -354,6 +361,10 @@ func (c *Cluster) isSecureClient() bool {
 	return c.cluster.Spec.TLS.IsSecureClient()
 }
 
+func (c *Cluster) IsPodPVEnabled() bool {
+	return c.cluster.Spec.Pod != nil && c.cluster.Spec.Pod.PV != nil
+}
+
 // bootstrap creates the seed etcd member for a new cluster.
 func (c *Cluster) bootstrap() error {
 	return c.startSeedMember()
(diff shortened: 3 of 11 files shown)


xiang90:
/cc @msohn @georgekuruvillak 

Can you guys also help on this PR.

As a side note, I noticed the SAP fork of etcd operator. Now etcd operator can already work with swift through its s3 compatible API. We want to get PV in.

We want to put some love in the project to move everyone back to upstream.

hongchaodeng:
@etcd-bot ok to test

rjtsdl:
@hongchaodeng , can you explain a bit, what is the node partition case? 
For other cases, i have tested, and documented the behavior in the description. Check if it makes sense.

brendandburns:
LGTM.

xiang90:
@rjtsdl 

i have not really looked into the code. but want to mention that if PV is enabled, etcd operator should 

1. always use write once mode to avoid two pods use the same etcd metadata. this can happen if k8s thinks a node is dead (however it is not) and creates a pod on some other node.
2. disable disaster detection. even if there is 1 out of 3 pods left there (or even 0/3 pods), the data is on PV. we can just wait for the other two member comes back with data.

pkg/apis/etcd/v1beta2/cluster.go
```diff
@@ -138,6 +138,13 @@ type PodPolicy struct {
 	// bootstrap the cluster (for example `--initial-cluster` flag).
 	// This field cannot be updated.
 	EtcdEnv []v1.EnvVar `json:"etcdEnv,omitempty"`
+
+	// PersistentVolumeClaimSpec ...
+	PersistentVolumeClaimSpec *v1.PersistentVolumeClaimSpec `json:"persistentVolumeClaimSpec,omitempty"`
+
+	// By default, kubernetes will mount a service account token into the etcd pods.
+	// AutomountServiceAccountToken indicates whether pods running with the service account should have an API token automatically mounted.
+	AutomountServiceAccountToken *bool `json:"automountServiceAccountToken,omitempty"`
```
hongchaodeng: Can you remove the changes related to `AutomountServiceAccountToken`?
This PR should be about `PersistentVolumeClaimSpec`.

pkg/controller/restore-operator/sync.go
```diff
@@ -219,6 +219,7 @@ func (r *Restore) createSeedMember(ec *api.EtcdCluster, svcAddr, clusterName str
 	}
 	ms := etcdutil.NewMemberSet(m)
 	backupURL := backupapi.BackupURLForRestore("http", svcAddr, clusterName)
+
```
hongchaodeng: remove blank line

pkg/util/k8sutil/k8sutil.go
```diff
@@ -264,10 +295,12 @@ func newEtcdPod(m *etcdutil.Member, initialCluster []string, clusterName, state,
 		livenessProbe,
 		readinessProbe)
 
-	volumes := []v1.Volume{
-		{Name: "etcd-data", VolumeSource: v1.VolumeSource{EmptyDir: &v1.EmptyDirVolumeSource{}}},
+	if cs.Pod != nil {
```
hongchaodeng: This changes aren't needed?

xiang90:
@rjtsdl 

it seems that this PR only adds the PV as the non-stable storage (same as empty dir) for etcd pod, not use PV as a stable storage? 

if we consider PV as stable storage, we should not add/remove members if the size field is not changed. if one pod dies, we should create another pod with the same FQDN + PV. no member changes need to be involved. when size is changed, we need to add/remove member as before.

pkg/apis/etcd/v1beta2/cluster.go
rjtsdl: Sure thing

rjtsdl:
@xiang90 you are right. It is treated as non-stable storage. We should make it as a stable storage.

xiang90:
@rjtsdl 

i am fine marking this as alpha which uses PV as temp storage. then we can iterate on it.

hongchaodeng:
@rjtsdl 
Let me clarify the four points in the plan:

- 1. Add PVC for etcd pod. This is basically what is achieved in this PR.
- 2. When node restart happens, current etcd pods would fail and won't come up anymore. This PR DOES NOT handle it. Besides changing the restart policy to "Always", following cases need be addressed:
  - 2.1. kubelet couldn't report to master and node would become "unready". etcd-operator would need to tolerate this case for some period of time. If node restarts and then etcd pods restarts and everything turns to healthy state, then etcd-operator wouldn't do anything. Otherwise, etcd-operator needs to take the nodes and thus etcd pods on it as unhealthy and do healing.
  - 2.2. Precious case assumes etcd pods still exist. If not, e.g. evicted, we will handle it like case 4.
- 3. If node partition happens, it could encounter case 2.1 that node would become "unready" and case 2.2 that etcd pods get deleted. The solution applies the same. But we need to think about and test this case.
- 4. Node eviction happens, and etcd pods get deleted. Let's assume PV are still there. In this case, etcd-operator should have the knowledge of current membership and reschedule the "failed" members onto other healthy nodes, and mount its corresponding PV back to it, and then everything should be fine because network and storage are still the same.

Gopkg.lock
```diff
@@ -328,6 +328,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "ff4a4b53ef454edf0ae2ec3bc497baf74d561af328faa123eb472d87954a906c"
+  inputs-digest = "b19769641bf117041022e3de0c490313047f3b7aaef54e00a7354573fff9b738"
```
hongchaodeng: No need to change this?

xiang90:
> When node restart happens, current etcd pods would fail and won't come up anymore.

it is unfair to say this PR does not handle it. this PR adds PV as non-stable storage. 

if the node goes down, the operator will still remove the member and add a new one as the empty-dir case, no?

hongchaodeng:
I'm fine to use it to replace ephemeral storage initially.

xiang90:
> I'm fine to use it to replace ephemeral storage initially.

just to be clear. not to replace current ephemeral storage, but another ephemeral storage options besides current empty-dir, right? eventually we should make PV a stable storage option.

hongchaodeng:
@xiang90 
Yes. That's the plan in https://github.com/coreos/etcd-operator/issues/1323#issuecomment-346872347 which is explained above. Those items are expected to be solved one by one.

rjtsdl:
That would be awesome. I will address comments, then we can merge it as initial step.

pkg/util/k8sutil/k8sutil.go
rjtsdl: It is taken care by function `AddEtcdVolumeToPod` in the same file.

Gopkg.lock
rjtsdl: It is updated automatically after I ran `update_vendor.sh`. Do I need to revert?

pkg/util/k8sutil/k8sutil.go
hongchaodeng: Sorry. I don't understand your comment?
Why add this change here?

Gopkg.lock
hongchaodeng: yes

pkg/util/k8sutil/k8sutil.go
rjtsdl: Here is 
```
func AddEtcdVolumeToPod(pod *v1.Pod, pvc *v1.PersistentVolumeClaim) {
	vol := v1.Volume{Name: etcdVolumeName}
	if pvc != nil {
		vol.VolumeSource = v1.VolumeSource{
			PersistentVolumeClaim: &v1.PersistentVolumeClaimVolumeSource{ClaimName: pvc.Name},
		}
	} else {
		vol.VolumeSource = v1.VolumeSource{EmptyDir: &v1.EmptyDirVolumeSource{}}
	}
	pod.Spec.Volumes = append(pod.Spec.Volumes, vol)
}
```

If it is not PV, it will just append the EmptyDir.

pkg/cluster/cluster.go
```diff
@@ -375,9 +375,26 @@ func (c *Cluster) setupServices() error {
 	return k8sutil.CreatePeerService(c.config.KubeCli, c.cluster.Name, c.cluster.Namespace, c.cluster.AsOwner())
 }
 
+func (c *Cluster) IsPodPVEnabled() bool {
```
xiang90: please add doc string on public method.

pkg/util/k8sutil/k8sutil.go
```diff
@@ -78,6 +78,10 @@ func GetPodNames(pods []*v1.Pod) []string {
 	return res
 }
 
+func PVCNameFromMemberName(memberName string) string {
```
xiang90: please add doc string on public method.

pkg/util/k8sutil/k8sutil.go
```diff
@@ -209,6 +213,18 @@ func newEtcdServiceManifest(svcName, clusterName, clusterIP string, ports []v1.S
 	return svc
 }
 
+func AddEtcdVolumeToPod(pod *v1.Pod, pvc *v1.PersistentVolumeClaim) {
```
xiang90: please add doc string on public method.

pkg/util/k8sutil/k8sutil.go
```diff
@@ -231,6 +249,19 @@ func NewSeedMemberPod(clusterName string, ms etcdutil.MemberSet, m *etcdutil.Mem
 	return pod
 }
 
+func NewEtcdPodPVC(m *etcdutil.Member, pvcSpec v1.PersistentVolumeClaimSpec, clusterName, namespace string, owner metav1.OwnerReference) *v1.PersistentVolumeClaim {
```
xiang90: please add doc string on public method.

pkg/apis/etcd/v1beta2/cluster.go
```diff
@@ -138,6 +138,9 @@ type PodPolicy struct {
 	// bootstrap the cluster (for example `--initial-cluster` flag).
 	// This field cannot be updated.
 	EtcdEnv []v1.EnvVar `json:"etcdEnv,omitempty"`
+
+	// PersistentVolumeClaimSpec ...
```
xiang90: complete the comment?

xiang90:
lgtm defer to @hongchaodeng

pkg/apis/etcd/v1beta2/cluster.go
```diff
@@ -138,6 +138,10 @@ type PodPolicy struct {
 	// bootstrap the cluster (for example `--initial-cluster` flag).
 	// This field cannot be updated.
 	EtcdEnv []v1.EnvVar `json:"etcdEnv,omitempty"`
+
+	// PersistentVolumeClaimSpec is the spec to describe PVC for the etcd container
+	// This field is optional. If no PVC spec, etcd container will use emptyDir as volume
+	PersistentVolumeClaimSpec *v1.PersistentVolumeClaimSpec `json:"persistentVolumeClaimSpec,omitempty"`
```
xiang90: Please mention that this field is alpha, and not use PV as stable storage as it should yet.

pkg/util/k8sutil/k8sutil.go
```diff
@@ -78,6 +78,11 @@ func GetPodNames(pods []*v1.Pod) []string {
 	return res
 }
 
+// PVCNameFromMemberName the way we get PVC name from the member name
+func PVCNameFromMemberName(memberName string) string {
```
hongchaodeng: PVCNameFromMemberName -> PVCNameFromMember

pkg/util/k8sutil/k8sutil.go
hongchaodeng: I still don't understand it.
Why is this related to PV or emptyDir? This is what I saw on this PR on github:
```
	if cs.Pod != nil {
		container = containerWithRequirements(container, cs.Pod.Resources)
	}
```
It only deals with resources.

pkg/util/k8sutil/self_hosted.go
```diff
@@ -141,14 +141,13 @@ done
...
 			// Self-hosted etcd pod need to endure node restart.
 			// If we set it to Never, the pod won't restart. If etcd won't come up, nor does other k8s API components.
-			RestartPolicy:                v1.RestartPolicyAlways,
-			Containers:                   []v1.Container{c},
-			Volumes:                      volumes,
-			HostNetwork:                  true,
-			DNSPolicy:                    v1.DNSClusterFirstWithHostNet,
-			Hostname:                     m.Name,
-			Subdomain:                    clusterName,
-			AutomountServiceAccountToken: func(b bool) *bool { return &b }(false),
```
hongchaodeng: revert this. It is not related to this PR.

rjtsdl: I need to delete `AutomountServiceAccountToken `. That is basically all the change here.

hongchaodeng: > I need to delete AutomountServiceAccountToken

What's the reason for it?

pkg/util/k8sutil/k8sutil.go
rjtsdl: Ah, i thought you are asking ln 267 & 268.

This line I don't think it has anything to do with this PR scope. It was there mostly because of my cherry pick from the old PR. 

I do think it is a nice thing to have. But not belong to this PR's scope. I can remove it though

pkg/util/k8sutil/self_hosted.go
rjtsdl: Because we want to 
> Can you remove the changes related to AutomountServiceAccountToken?
This PR should be about PersistentVolumeClaimSpec.

The comment from your previous review.

hongchaodeng: Can you explain why AutomountServiceAccountToken is related to PV?

rjtsdl: I just get it back. 

I deleted it while I was trying to remove changes to AutomountServiceAccountToken. Apparently, it shouldn't be removed here.

pkg/util/k8sutil/k8sutil.go
hongchaodeng: Please remove this.
It's duplicate. Resource requirements/pod policy are applied outside this func:
https://github.com/coreos/etcd-operator/blob/788063849dac25805e0bcdc637400ba55aebe9a3/pkg/util/k8sutil/k8sutil.go#L338

rjtsdl:
@hongchaodeng thx for explaining. Removed the line.

hongchaodeng:
Thanks everyone!
Especially @sgotti for the initiate and @rjtsdl for the finish!

softblade-c:
Could someone kindly give an example how to mount persistent volume instead of emptyDir in yaml?  
I haven't found any infomation on internet, and tried several ways like the following:
```
  volumes:
    - name: vol-etcd-data
      hostPath:
        path: /mnt/cluster/85a6410fc1b/etcd-data
        type: ''
...
  containers:
    - name: etcd
      volumeMounts:
      - name: vol-etcd-data
        mountPath: /var/etcd
```
They have no effect. voloumes etcd-data is still emptyDir.  
Thanks.

PR #1863, opened 2018-01-17
k8sutil: run etcd pods as non-root user

hasbro17:
ref: #806 

With the given pod security context, the etcd containers will run as the non-root user 9000, and any volumes will have permissions for the group 9000 allowing the etcd process to read/write to them.


hongchaodeng:
lgtm

hongchaodeng:
cc @xiang90

PR #1865, opened 2018-01-18
Enable users to specify a ServiceAccount for their etcd pods

edulop91:
Clusters with locked down default service accounts are unable to
mount secrets into the etcd pods. Allow users to specify a
service account to use for the etcd pods.

### DIFF
--- a/pkg/util/k8sutil/pod_util.go
+++ b/pkg/util/k8sutil/pod_util.go
@@ -93,6 +93,8 @@ func applyPodPolicy(clusterName string, pod *v1.Pod, policy *api.PodPolicy) {
 		return
 	}
 
+	pod.Spec.ServiceAccountName = policy.ServiceAccountName
+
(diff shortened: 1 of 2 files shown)


hongchaodeng:
@edulop91 
Can you create an issue first to describe your problem and how to reproduce it?

edulop91:
@hongchaodeng Yes! https://github.com/coreos/etcd-operator/issues/1866

pkg/util/k8sutil/pod_util.go
```diff
@@ -93,6 +93,8 @@ func applyPodPolicy(clusterName string, pod *v1.Pod, policy *api.PodPolicy) {
 		return
 	}
 
+	pod.Spec.ServiceAccountName = policy.ServiceAccountName
```
mat285: You should probably check to make sure that this is filled before adding it to the pod spec

PR #1880, opened 2018-01-23
CHANGELOG: update

hasbro17:
[skip ci]

/cc @hongchaodeng

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,31 @@
 
 ### Added
 
+- Added the option to use PersistentVolume as non-stable storage for etcd pods. This feature is still alpha and subject to change in future releases [#1861](https://github.com/coreos/etcd-operator/pull/1861)
+
 ### Changed
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Changed
 
+- Changed etcd pod member names to be unique by having a random suffix instead of a sequence number e.g `example-etcd-cluster-2885z`
```
hongchaodeng: Compare what it looked like before and looks now.
Also mention about compatibility.

hongchaodeng:
Also need to mention that we fixed a bug that restore might fail.

hongchaodeng:
It's mystic convention. But object name could not exceed 64 characters:
https://github.com/kubernetes/apimachinery/blob/1168e538ea3ccf444854d1fdd5681d2d876680a7/pkg/util/validation/validation.go#L33

We also need to subtract the five characters from it.

fanminshi:
lgtm

PR #1896, opened 2018-02-01
k8sutil: make pod annotations configurable

akrennmair:
Annotations are a way to attach metadata to Kubernetes objects. This PR extends the operator configuration with an option to attach annotations to pods that the operator creates for the cluster. A
use case for this is prometheus using annotations to identify pods from which metrics can be scraped, including pod-specific scraping configuration. The changes in this PR would enable users to configure
etcd pods created by the operator to be annotated for scraping by prometheus.

### DIFF
--- a/pkg/util/k8sutil/k8sutil.go
+++ b/pkg/util/k8sutil/k8sutil.go
@@ -372,7 +372,6 @@ func newEtcdPod(m *etcdutil.Member, initialCluster []string, clusterName, state,
 			},
 		},
 	}
-	setAnnotations(pod, cs.Pod.Annotations)
 	SetEtcdVersion(pod, cs.Version)
 	return pod
 }
@@ -483,10 +482,4 @@ func UniqueMemberName(clusterName string) string {
 		clusterName = clusterName[:maxNameLength]
 	}
 	return clusterName + "-" + suffix
-}
-
-func setAnnotations(pod *v1.Pod, annotations map[string]string) {
-	for key, value := range annotations {
-		pod.ObjectMeta.Annotations[key] = value
-	}
 }
--- a/pkg/util/k8sutil/k8sutil.go
+++ b/pkg/util/k8sutil/k8sutil.go
@@ -372,6 +372,7 @@ func newEtcdPod(m *etcdutil.Member, initialCluster []string, clusterName, state,
 			},
 		},
 	}
+	setAnnotations(pod, cs.Pod.Annotations)
 	SetEtcdVersion(pod, cs.Version)
 	return pod
 }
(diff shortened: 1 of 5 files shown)


hongchaodeng:
@etcd-bot ok to test

hongchaodeng:
> A
use case for this is prometheus using annotations to identify pods from which metrics can be scraped, including pod-specific scraping configuration.

Can you clarify this more?

akrennmair:
> Can you clarify this more?

Sure. Prometheus uses kubernetes's API to find kubernetes objects within a kubernetes cluster. By default, it doesn't know whether it can scrape metrics from an object or not. In order to identify an object such as a pod to be scrapeable, the standard pattern that is used in prometheus-provided examples (e.g. https://github.com/prometheus/prometheus/blob/master/documentation/examples/prometheus-kubernetes.yml) is to use annotations such as `prometheus.io/scrape: "true"` to indicate that it can be scraped, and to provide additional information like the metrics port and the URL path via annotations like `prometheus.io/port: "2379"` and `prometheus.io/path: "/metrics"`. This is not hard-coded, but the recommended way of configuring prometheus.

Since `etcd-operator` spawns `etcd` pods, and `etcd` exposes prometheus metrics, I would like to scrape them, as well. Following the recommended prometheus way of doing this, I need to set annotations on these pods to indicate to prometheus that they can be scraped. `etcd-operator` doesn't provide a configuration option for this yet, hence this PR.

Again, as mentioned before, the annotations used by prometheus are merely convention as suggested by example configuration that comes with prometheus, and it's not hard-coded. I can imagine there are other use cases that people may come up with, so that's why the PR keeps the addition of annotations as simple and generic as possible.

I hope this clarifies my intention and motivation behind this PR.

akrennmair:
I just realized that the tests have failed. I didn't take into account that cs.Pod could be nil, and so I implemented a fix that also does a bit of restructuring. Should I put this into a separate commit, or amend the previous commit and force-push?

pkg/util/k8sutil/k8sutil.go
```diff
@@ -372,6 +372,7 @@ func newEtcdPod(m *etcdutil.Member, initialCluster []string, clusterName, state,
 			},
 		},
 	}
+	setAnnotations(pod, cs.Pod.Annotations)
```
hongchaodeng: All pod policy setting should be done in applyPodPolicy(): 
https://github.com/coreos/etcd-operator/blob/83d7ae64f92992b05818919fb682830225201bf1/pkg/util/k8sutil/pod_util.go#L91:6

This would also avoid the nil pointer

hongchaodeng:
@akrennmair 
Are you scraping from the pods or the endpoints? Note that this would only change the pod annotations, not the endpoint.

pkg/util/k8sutil/k8sutil.go
akrennmair: Yep, I realized this right after I saw the failed tests. Fixed it in my latest commit.

akrennmair:
@hongchaodeng prometheus is scraping from the pods directly.

hongchaodeng:
Please also update generated code:
https://github.com/coreos/etcd-operator/blob/master/hack/k8s/codegen/README.md

hongchaodeng:
LGTM
@hasbro17

hasbro17:
LGTM

Issue #1906, opened 2018-02-05
backup: make backup timeout configurable 


fanminshi:
Currently, etcd backup operator has a default non-configuration backup timeout of 60s. This can causing issue when retrieving and saving backup is more than 60s as seen in https://github.com/coreos/etcd-operator/issues/1903.

TODOs:
- [x] Define a new backup field to take in a backup timeout. https://github.com/coreos/etcd-operator/pull/1907
- [x] Implement backup timeout in etcd backup operator. https://github.com/coreos/etcd-operator/pull/1908
- [x] Add ctx to BackupWriter.Write() method. https://github.com/coreos/etcd-operator/pull/1910
- [x] Manual testing plan.

xiang90:
sgtm

PR #1908, opened 2018-02-05
*: implement configurable backup timeout

fanminshi:
ref: https://github.com/coreos/etcd-operator/issues/1906

### DIFF
--- a/pkg/apis/etcd/v1beta2/backup_types.go
+++ b/pkg/apis/etcd/v1beta2/backup_types.go
@@ -84,8 +84,8 @@ type BackupSource struct {
 
 // BackupPolicy defines backup policy.
 type BackupPolicy struct {
-	// timeout is the maximal time of retriving plus saving an etcd backup.
-	Timeout int64 `json:"timeout,omitempty"`
+	// TimeoutInSecond is the maximal allowed time in second of the entire backup process.
+	TimeoutInSecond int64 `json:"timeoutInSecond,omitempty"`
 }
 
 // BackupStatus represents the status of the EtcdBackup Custom Resource.
--- a/pkg/backup/backup_manager.go
+++ b/pkg/backup/backup_manager.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 
 	"github.com/coreos/etcd-operator/pkg/backup/writer"
+	"github.com/coreos/etcd-operator/pkg/util/constants"
 
 	"github.com/coreos/etcd/clientv3"
 	"github.com/sirupsen/logrus"
@@ -91,10 +92,11 @@ func getClientWithMaxRev(ctx context.Context, endpoints []string, tc *tls.Config
 	maxRev := int64(0)
 	errors := make([]string, 0)
 	for _, endpoint := range endpoints {
+		// TODO: update clientv3 to 3.2.x and then use ctx as in clientv3.Config.
 		cfg := clientv3.Config{
-			Endpoints: []string{endpoint},
-			Context:   ctx,
-			TLS:       tc,
+			Endpoints:   []string{endpoint},
+			DialTimeout: constants.DefaultDialTimeout,
+			TLS:         tc,
 		}
 		etcdcli, err := clientv3.New(cfg)
 		if err != nil {
--- a/pkg/backup/backup_manager.go
+++ b/pkg/backup/backup_manager.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 
 	"github.com/coreos/etcd-operator/pkg/backup/writer"
-	"github.com/coreos/etcd-operator/pkg/util/constants"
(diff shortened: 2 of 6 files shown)


pkg/controller/backup-operator/sync.go
```diff
@@ -119,15 +121,21 @@ func (b *Backup) handleBackup(spec *api.BackupSpec) (*api.BackupStatus, error) {
...
+	// When BackupPolicy.Timeout <= 0, use default DefaultBackupTimeout.
+	backupTimeout := time.Duration(constants.DefaultBackupTimeout)
+	if spec.BackupPolicy != nil && spec.BackupPolicy.Timeout > 0 {
+		backupTimeout = time.Duration(spec.BackupPolicy.Timeout) * time.Second
+	}
+
 	switch spec.StorageType {
 	case api.BackupStorageTypeS3:
-		bs, err := handleS3(b.kubecli, spec.S3, spec.EtcdEndpoints, spec.ClientTLSSecret, b.namespace)
+		bs, err := handleS3(b.kubecli, spec.S3, backupTimeout, spec.EtcdEndpoints, spec.ClientTLSSecret, b.namespace)
```
hongchaodeng: I rethink about it.
Instead of defining the timeout to be "saving backup", we should define the timeout to be the entire process. Each type could spend different amount of time doing other work not on saving backup. As a user, I might just define 60s and want the backup to be made around that limit without caring about getting secret, etc.

hongchaodeng: In code:

```
ctx, cancel := context.WithTimeout(context.Background(), backupTimeout)
handleS3(ctx)
```

fanminshi: @hongchaodeng yeah, that's my original thinking. However, the issue with an timeout for entire process is that kube client doesn't take-in context so I can't include kubecli activities such as retrieving secret as part of the total timeout. Hence,  the `Timeout` only applies to retrieving+saving backup.

hongchaodeng:
> kube client doesn't take-in context

k8s client upstream issue ref: https://github.com/kubernetes/kubernetes/issues/46503

It also looks like in the REST API level client-go provides context support: https://github.com/kubernetes/client-go/blob/4def1285ff0e4d1fee7cc9e2684ef9923dca591f/rest/request.go#L411
But this is not propogated onto upper level.

There is a default 30s Dialer.Timeout: 
https://github.com/kubernetes/client-go/blob/f4d24c8a47016d9ffcc8a44c4011d153cb413367/transport/cache.go#L84-L87
But no default timeout for http request.

My current conclusion is that this is not limited to backup. This is a general issue we need to work around or solve in upstream. Otherwise if it keeps hanging, anything using client-go will have problems.

My current suggest would be to leave a TODO signifying the current limitation of timeout.

fanminshi:
cc/ @hongchaodeng @hasbro17

pkg/apis/etcd/v1beta2/backup_types.go
```diff
@@ -84,7 +84,7 @@ type BackupSource struct {
 
 // BackupPolicy defines backup policy.
 type BackupPolicy struct {
-	// timeout is the maximal time of retriving plus saving an etcd backup.
+	// timeout is the maximal allowed time in second of the entire backup process.
 	Timeout int64 `json:"timeout,omitempty"`
```
hongchaodeng: After a second thought, maybe we should rename this TimeoutInSecond or TimeoutSecond

pkg/backup/backup_manager.go
```diff
@@ -82,20 +78,21 @@ func (bm *BackupManager) SaveSnap(s3Path string) (int64, string, error) {
...
 }
 
-func getClientWithMaxRev(endpoints []string, tc *tls.Config) (*clientv3.Client, int64, error) {
+func getClientWithMaxRev(ctx context.Context, endpoints []string, tc *tls.Config) (*clientv3.Client, int64, error) {
 	mapEps := make(map[string]*clientv3.Client)
 	var maxClient *clientv3.Client
 	maxRev := int64(0)
 	errors := make([]string, 0)
 	for _, endpoint := range endpoints {
+		// TODO: update clientv3 to 3.2.x and thenuse ctx as in clientv3.Config.
```
hongchaodeng: thenuse -> then use

pkg/apis/etcd/v1beta2/backup_types.go
fanminshi: `TimeoutInSecond` seems good.

fanminshi:

hongchaodeng:
After offline discussion, we are going to add `ctx` to BackupWriter.Write() method too.

hongchaodeng:
Let's move forward this and change BackupWriter in following PR.

PR #1919, opened 2018-02-07
pkg/util: Change check-dns init container image

lander2k2:
The image currently used is busybox:1.28.0 which uses uclibc.  This is
causing nslookup to fail name resolution when it shouldn't in some
environments.  The version of busybox that uses glibc does not suffer the
same problem.

Related:
https://github.com/docker-library/busybox/issues/9
https://github.com/docker-library/busybox/issues/27

### DIFF
--- a/pkg/util/k8sutil/k8sutil.go
+++ b/pkg/util/k8sutil/k8sutil.go
@@ -346,7 +346,7 @@ func newEtcdPod(m *etcdutil.Member, initialCluster []string, clusterName, state,
 		},
 		Spec: v1.PodSpec{
 			InitContainers: []v1.Container{{
-				Image: "busybox:1.28.0",
+				Image: "busybox:1.28.0-glibc",
 				Name:  "check-dns",
 				// In etcd 3.2, TLS listener will do a reverse-DNS lookup for pod IP -> hostname.
 				// If DNS entry is not warmed up, it will return empty result and peer connection will be rejected.


hongchaodeng:
@etcd-bot ok to test

hongchaodeng:
@lander2k2 
Can you provide a failure scenario and let me see if I can reproduce it.

lander2k2:
@hongchaodeng 
I've been unable to reproduce the error in AWS but here's the test I employed in the VMware environment where this is happening.
```
# bbox-po.yaml
apiVersion: v1
kind: Pod
metadata:
  name: bbox
spec:
  containers:
  - image: busybox
    name: bbox
    command: ["sleep", "86400"]
```

```
# bbox-glibc-po.yaml
apiVersion: v1
kind: Pod
metadata:
  name: bbox-glibc
spec:
  containers:
  - image: busybox:glibc
    name: bbox-glibc
    command: ["sleep", "86400"]
```

```
$ kubectl create -f bbox-po.yaml
$ kubectl create -f bbox-glibc-po.yaml
$ kubectl exec bbox-glibc /bin/nslookup tectonic-lb.tectonic-system.svc
Server:    10.3.0.10
Address 1: 10.3.0.10 kube-dns.kube-system.svc.cluster.local

Name:      tectonic-lb.tectonic-system.svc
Address 1: 10.3.155.251 tectonic-lb.tectonic-system.svc.cluster.local

$ kubectl exec bbox /bin/nslookup tectonic-lb.tectonic-system.svc  nslookup: 
can't resolve 'tectonic-lb.tectonic-system.svc'
Server:    10.3.0.10
Address 1: 10.3.0.10 kube-dns.kube-system.svc.cluster.local
```

Sorry I can't give a more universally reproducible example.  I did think this was related to a bug in uclibc related to nameserver search path length [1] but I've been unable to induce the error outside the VMware environment by manipulating the search path.  :/

[1] https://mailman.uclibc-ng.org/pipermail/devel/2016-February/000775.html

pkg/util/k8sutil/k8sutil.go
```diff
@@ -346,7 +346,7 @@ func newEtcdPod(m *etcdutil.Member, initialCluster []string, clusterName, state,
 		},
 		Spec: v1.PodSpec{
 			InitContainers: []v1.Container{{
-				Image: "busybox:1.28.0",
+				Image: "busybox:1.28.0-glibc",
```
hongchaodeng: Can you add a comment about the search path issue on VMWare environment and why use glibc instead?

hongchaodeng:
Going to do a release later today. So I will merge this and add the comment myself.

PR #1922, opened 2018-02-08
pkg/util: Add comment regarding glibc image for check-dns container

lander2k2:


pkg/util/k8sutil/k8sutil.go
```diff
@@ -348,6 +348,8 @@ func newEtcdPod(m *etcdutil.Member, initialCluster []string, clusterName, state,
 		},
 		Spec: v1.PodSpec{
 			InitContainers: []v1.Container{{
+				// busybox:latest uses uclibc which contains a bug that sometimes prevents name resolution
+				// More info: https://github.com/docker-library/busybox/issues/9
```
hongchaodeng: This issue seemed to be claimed "fixed in https://github.com/docker-library/busybox/pull/10".

Maybe we should refer to https://github.com/docker-library/busybox/issues/27 instead?

lander2k2: Strange.  I checked uclibc source code [1] to verify that patch [2] never was merged. :confused: Nonetheless, I will reference the more recent short name issue.

[1] https://git.uclibc.org/uClibc/tree/libc/inet/resolv.c
[2] https://mailman.uclibc-ng.org/pipermail/devel/2016-February/000775.html

hongchaodeng:
LGTM

PR #1932, opened 2018-03-01
change log level on commandline

prune998:
Setup Logrus so user can control the log level on commandline


hongchaodeng:
Hi @prune998 
Thanks for the patch.
As far as I know this patch only sets the global level of logrus, where etcd-operator loggers are not. Can you fix that?

prune998:
I had no time to scrub through the code. The logging things sounds to be pretty, well, distributed, and of course, we should go through each package to change the logging behaviour.

I'm no Go expert and I'm not sure to exactly know how to achieve that.
My intuition would point me to pass the `cmd` logrus logger to each sub-packages so we can control it from one place. Maybe it would be better to add a variable into each sub-package to handle that at creation time ? 

This should be trivial for people knowing the code well, which I'm not...

Issue #1927, opened 2018-02-14
Unable to use etcd-operator in disconnected Kubernetes cluster with private registry


yehiyam:
When working in a disconnected k8s cluster, we use a private docker registry for all of the images.
While the image name for the etcd itself can be specified, the busybox image name is hard coded. 
Can the option to specify the image/tag/registry for the busybox image be added?

Thanks

hongchaodeng:
Same issue for the restore container: https://github.com/coreos/etcd-operator/blob/0a4c2c15d05411deaf2dbb1555bc183dd051f4cb/pkg/util/k8sutil/k8sutil.go#L102

PR #1934, opened 2018-03-05
fixes #1933, allow tutum/curl image to be overridden

### LINKED ISSUE #1933: tutum/curl image can't be overridden
The operator allows for the base image to be overridden but not the curl image. This makes is difficult to run on a private network with no internet connections.

daniel-garcia:


daniel-garcia:
Looks like this is already being handled in #1927

pkg/util/k8sutil/k8sutil.go
```diff
@@ -96,10 +96,14 @@ func PVCNameFromMember(memberName string) string {
 }
 
 func makeRestoreInitContainers(backupURL *url.URL, token, repo, version string, m *etcdutil.Member) []v1.Container {
+	curlImage := "tutum/curl"
+	if s := os.Getenv("CURL_IMAGE"); len(s) > 0 {
```
hongchaodeng: Why get it from ENV?

daniel-garcia: i was basing this MR on an older version of etcd operator (0.8.2). What is the recommended approach? Do you want this to be a CLI option on the main process?

hongchaodeng: > Do you want this to be a CLI option on the main process?

I'm not sure what you meant by "CLI option"?
This would be better in the EtcdRestore CR spec.

daniel-garcia: I'll reimplement as an option to the EtcdRestore CR spec.

PR #1947, opened 2018-04-05
Avoid probe failure from auth enable

Seth-Miller:
Change the probe to check `endpoint health` instead of `get foo` so that enabling authorization does not take down the cluster.

### DIFF
--- a/pkg/util/k8sutil/pod_util.go
+++ b/pkg/util/k8sutil/pod_util.go
@@ -69,7 +69,7 @@ func containerWithRequirements(c v1.Container, r v1.ResourceRequirements) v1.Con
 }
 
 func newEtcdProbe(isSecure bool) *v1.Probe {
-	// etcd pod is alive only if a linearizable get succeeds.
+	// etcd pod is healthy only if it can participate in consensus
 	cmd := "ETCDCTL_API=3 etcdctl endpoint health"
 	if isSecure {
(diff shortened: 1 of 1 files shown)


hasbro17:
@Seth-Miller Sorry for the delay on this. Taking a look at this now. If you don't have time to follow up on this then I can take over.

hasbro17:
@etcd-bot test this please

pkg/util/k8sutil/pod_util.go
```diff
@@ -70,10 +70,10 @@ func containerWithRequirements(c v1.Container, r v1.ResourceRequirements) v1.Con
 
 func newEtcdProbe(isSecure bool) *v1.Probe {
 	// etcd pod is alive only if a linearizable get succeeds.
```
hasbro17: Update this comment as well.
```
// etcd pod is healthy only if it can participate in consensus
```

hasbro17:
I've created an issue for this to better track and discuss this.
ref: #1957

raoofm:
@hasbro17 can this be moved forward?

raoofm:
@fanminshi ping

pkg/util/k8sutil/pod_util.go
fanminshi: @Seth-Miller can you address the above comment.

fanminshi:
lgtm

Issue #1948, opened 2018-04-12
Remove the default non-root SecurityContext


hasbro17:
The default non-root SecurityContext for etcd pods was added so that the etcd pod containers would run as non-root.
https://github.com/coreos/etcd-operator/pull/1863

However hardcoding the default value for UID and FSGroup 9000 is problem because the allowed SecurityContext values for a pod depends is specific to the Admission Control Policy of a cluster.
So the etcd-operator is potentially not be able to create etcd-pods on clusters that don't allow pods with (UID,FSGroup) 9000.

To resolve this issue, the operator should resort back to not setting any SecurityContext for the etcd-pods and instead let the user specify that per EtcdCluster as part of the PodPolicy configuration.

E.g:
```YAML
apiVersion: "etcd.database.coreos.com/v1beta2"
kind: "EtcdCluster"
metadata:
  name: "example-etcd-cluster"
spec:
  size: 3
  version: "3.2.13"
  pod:
    securityContext:
      runAsNonRoot: true
      runAsUser: 1234
      fsGroup: 1234
```

PR #1949, opened 2018-04-12
*: make SecurityContext configurable via PodPolicy

hasbro17:
ref: #1948 
/cc @fanminshi


fanminshi:
lgtm after passing ci.

PR #1955, opened 2018-04-22
Add a time-out for waiting on DNS

meyskens:
Adding a timeout for waiting on the DNS entry to exists. This prevents the init container to be in an inifinite loop in case of a failure of registering the DNS on startup

We have seen this happening on GKE with preemptive nodes where the pod got created before the network was able to assign it an IP causing the init container to be stuck forever in this loop.

### DIFF
--- a/pkg/util/k8sutil/k8sutil.go
+++ b/pkg/util/k8sutil/k8sutil.go
@@ -365,11 +365,20 @@ func newEtcdPod(m *etcdutil.Member, initialCluster []string, clusterName, state,
 				Name:  "check-dns",
 				// In etcd 3.2, TLS listener will do a reverse-DNS lookup for pod IP -> hostname.
 				// If DNS entry is not warmed up, it will return empty result and peer connection will be rejected.
+				// In some cases the DNS is not created correctly so we need to time out after a given period.
 				Command: []string{"/bin/sh", "-c", fmt.Sprintf(`
+					TIMEOUT_READY=%d
 					while ( ! nslookup %s )
 					do
-						sleep 2
-					done`, m.Addr())},
+						# If TIMEOUT_READY is 0 we should never time out and exit 
+						TIMEOUT_READY=$(( TIMEOUT_READY-1 ))
+                        if [ $TIMEOUT_READY -eq 0 ];
+				        then
+				            echo "Timed out waiting for DNS entry"
+				            exit 1
+				        fi
+						sleep 1
+					done`, cs.Pod.DNSTimeoutInSecond, m.Addr())},
 			}},
 			Containers:    []v1.Container{container},
 			RestartPolicy: v1.RestartPolicyNever,
--- a/pkg/apis/etcd/v1beta2/cluster.go
+++ b/pkg/apis/etcd/v1beta2/cluster.go
@@ -152,6 +152,11 @@ type PodPolicy struct {
 	// SecurityContext specifies the security context for the entire pod
 	// More info: https://kubernetes.io/docs/tasks/configure-pod-container/security-context
 	SecurityContext *v1.PodSecurityContext `json:"securityContext,omitempty"`
+
+	// DNSTimeoutInSecond is the maximum allowed time for the init container of the etcd pod to
+	// reverse DNS lookup its IP given the hostname.
+	// The default is to wait indefinitely and has a vaule of 0.
+	DNSTimeoutInSecond int64 `json:"DNSTimeoutInSecond,omitempty"`
 }
 
 // TODO: move this to initializer


lawrencejones:
Hey @fanminshi, hope you're doing ok!

We're trying to use the etcd operator at GoCardless, but are running into problems in our lab environment. We deliberately use GCP's preemptible nodes in lab to test robustness to individual node failure. Unfortunately the etcd clusters managed by the operator survive for less than 24hrs in this environment, due to the bug we describe in this PR.

We think it's probably something to do with https://github.com/kubernetes/kubernetes/issues/44273, but timings out after 5m of failing DNS makes sense without a longterm fix to that issue being available.

Handling this edge case will help us and anyone else running on preemptible machines. If we can do anything to get this PR merged, or need to adapt it to better suit the style of the project, then please let us know.

fanminshi:
@lawrencejones @meyskens I am taking a look.

fanminshi:
I'd suggest to make the timeout configurable because we don't really know what's a good timeout for different scenarios.  

You can add a `DNSTimeoutInSecond` to the PodPolicy of the `ClusterSpec`.
Feel free to use a better name if you come up with one.

```go 
type PodPolicy struct {
...
// DNSTimeoutInSecond is the maximum allowed time for the init container of the etcd pod to
// reverse DNS lookup its IP given the hostname.
// The default is to wait indefinitely and has a vaule of 0. 
 DNSTimeoutInSecond int64 `json:"DNSTimeoutInSecond,omitempty"`
}
```
https://github.com/coreos/etcd-operator/blob/master/pkg/apis/etcd/v1beta2/cluster.go#L105

And use DNSTimeoutInSec to set the init container's `TIMEOUT_READY` period. And, Make sure that the default value is 0 and it means to wait forever.

meyskens:
@fanminshi I have added the `DNSTimeoutInSecond` option 😄Thanks for that! Sure let me know if you have any other feedback!

pkg/util/k8sutil/k8sutil.go
```diff
@@ -365,11 +365,19 @@ func newEtcdPod(m *etcdutil.Member, initialCluster []string, clusterName, state,
...
-						sleep 2
-					done`, m.Addr())},
+						# If TIMEOUT_READY is 0 we should never time out and exit 
+						if [ $(( TIMEOUT_READY-- )) -eq 0 ];
+				            then
+				                echo "Timed out waiting for DNS entry"
+				                exit 1
+				            fi
+						sleep 1
+					done`, m.Addr(), cs.Pod.DNSTimeoutInSecond)},
```
fanminshi: the ordering of the args`fmt.Sprintf("%d %s", m.Addr(),cs.Pod.DNSTimeoutInSecond)` doesn't seem right. Should it be `fmt.Sprintf("%d %s", cs.Pod.DNSTimeoutInSecond, m.Addr())`?

fanminshi:
@meyskens were you able to test this pr out to ensure that ti works as intended?

pkg/util/k8sutil/k8sutil.go
meyskens: You're right! Let me fix this

meyskens:
@fanminshi I am now preparing a test build of it to test out in our lab cluster

fanminshi:
@meyskens sounds good. Let me know how the test goes. Once this PR passes your test, then we can merge this PR.

meyskens:
The test proved that my shell scripting skills are not the best 😉 
I have patched the loop and it works!

pkg/util/k8sutil/k8sutil.go
```diff
@@ -365,11 +365,20 @@ func newEtcdPod(m *etcdutil.Member, initialCluster []string, clusterName, state,
...
+				// In some cases the DNS is not created correctly so we time out after 10 minutes in this case.
 				Command: []string{"/bin/sh", "-c", fmt.Sprintf(`
+					TIMEOUT_READY=%d
 					while ( ! nslookup %s )
 					do
-						sleep 2
-					done`, m.Addr())},
+						# If TIMEOUT_READY is 0 we should never time out and exit 
+						TIMEOUT_READY=$(( TIMEOUT_READY-1 ))
+                        if [ $TIMEOUT_READY -eq 0 ];
```
fanminshi: Tap here? make the spacing of `if` align with the `then`.

fanminshi:
@etcd-bot ok to test

pkg/util/k8sutil/k8sutil.go
meyskens: 👍

pkg/util/k8sutil/k8sutil.go
```diff
@@ -365,11 +365,20 @@ func newEtcdPod(m *etcdutil.Member, initialCluster []string, clusterName, state,
 				Name:  "check-dns",
 				// In etcd 3.2, TLS listener will do a reverse-DNS lookup for pod IP -> hostname.
 				// If DNS entry is not warmed up, it will return empty result and peer connection will be rejected.
+				// In some cases the DNS is not created correctly so we time out after 10 minutes in this case.
```
fanminshi: one last nit.
`In some cases the DNS is not created correctly so we time out after 10 minutes in this case.` ->
`In some cases the DNS is not created correctly so we need to time out after a given period`

fanminshi:
lgtm after nits

pkg/util/k8sutil/k8sutil.go
meyskens: 👍

fanminshi:
@meyskens Thanks for all the hardwork. The code looks great. Currently, some of our CI is down. Let me fix that first and then re-run it against PR. Afterward, I'll merge the PR.

fanminshi:
@etcd-bot ok to test

hasbro17:
@etcd-bot retest this please

fanminshi:
I will fix the CI failure in a subsequent PR.

PR #1961, opened 2018-05-15
e2e/framework: support on-cluster as well as kubeconfig

colhom:
If kubeconfig is not provided to e2e framwork binary, assume that the framework is running on-cluster

\cc @hasbro17


hasbro17:
@colhom As we discussed offline the current framework already falls back on the in-cluster config if no kubeconfig is provided.
https://github.com/coreos/etcd-operator/blob/master/test/e2e/framework/framework.go#L64:27
https://github.com/kubernetes/client-go/blob/b379a2676756c2a4ecf7923ac191008bf9843a9e/tools/clientcmd/client_config.go#L528-L531

I think we can close this unless anything else is needed.

PR #1970, opened 2018-06-11
Support arbitrary etcd image names

justinsb:
We allow a simple template to be defined in the PodPolicy, and we
perform substitution of the repository and version into that template.
The default template keeps the existing behaviour, so no user changes
should be needed.

This allows for registries that may not follow the same format as
quay.io - for example when mirroring to a private repo for no-internet
installs, the limited depth allowed by docker often means that users
will embed some of the repo name into the image name.

### DIFF
--- a/pkg/apis/etcd/v1beta2/cluster.go
+++ b/pkg/apis/etcd/v1beta2/cluster.go
@@ -23,8 +23,8 @@ import (
 )
 
 const (
-	defaultRepository  = "quay.io/coreos/etcd"
-	DefaultEtcdVersion = "3.2.13"
+	DefaultEtcdRepository = "quay.io/coreos/etcd"
+	DefaultEtcdVersion    = "3.2.13"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


crazyleeyang:
@justinsb  can this spec  solve your problem:
```
 spec:
   size: 5
   version: "test2"
   repository: "xxx.xx.x/test/etcd"
   pod:
     busyboxImage: "xxx.xx.x/test/busybox:1.28.0-glibc"
```

perotinus:
@crazyleeyang @justinsb Is there anything preventing this PR from merging? We would like to have this functionality.

ianchakeres:
CC: @hasbro17 @fanminshi

jpbetz:
I'm in favor of this.  We'd use it, and it seams generally valuable for any organizations that already have etcd images in a repo and would like to be able to point etcd-operator to those images.

philips:
Overall this looks OK to me. I think these changes would be necessary to merge it:

1. Please have it print an error if the template doesn't have the version variable
1. Please rename the field to EtcdImageTemplate because we don't want people not providing a way for the operator to update the version
1. Please write a doc on how to use the field and the template substitution. I know we don't have a schema doc but inevitably someone will want to do this and I don't want to point them at comments in code.

PR #1976, opened 2018-07-04
close s3 reader after saving snapshot

### LINKED ISSUE #1965: etcd backup operator TCP socket memory leak
We just found that create `EtcdBackup` custom resource will cause TCP socket memory leak which can be observed by watching the output of command `cat /proc/net/sockstat`.

In our case, each time we create a `EtcdBackup` custom resource backups to S3 bucket will leave ~3000 unit TCP memory. Eventually the TCP memory usage will hit the max (from `/proc/sys/net/ipv4/tcp_mem`) if we backup periodically (eg. via cronjob) and the system network performance becomes extremely slow.

Kubernetes 1.9.7 on AWS, created by kops 1.9.0
Both etcd-operator 0.8.4 and 0.9.2 have this issue.

etcd-backup-operator `Deployment`:
```yaml
kind: Deployment
metadata:
  namespace: holosense
  name: etcd-backup-operator
spec:
  replicas: 1
  template:
    metadata:
      labels:
        name: etcd-backup-operator
    spec:
      containers:
      - name: etcd-backup-operator
        image: quay.io/coreos/etcd-operator:v0.9.2
        command:
        - etcd-backup-operator
        env:
        - name: MY_POD_NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace
        - name: MY_POD_NAME
          valueFrom:
            fieldRef:
              fieldPath: metadata.name
```

`EtcdBackup` custom resource:
```yaml
kind: EtcdBackup
metadata:
    namespace: my-namespace
    name: etcd-backup-01
spec:
    etcdEndpoints: ["etcd-cluster-client:2379"]
    storageType: S3
    s3:
    path: etcd-backup/etcd.backup.01
    awsSecret: my-aws-secret
```

daniel-garcia:
After uploading the snapshot, the backup operator performs a get
to ensure that the snapshot was actually saved. The problem is that
the resulting get request body is not closed. This causes a leak
of http connections.

fixes #1965


drewwells:
Definitely need this, we're seeing memory leaks in backup operator

fanminshi:
lgtm thanks!

fanminshi:
@daniel-garcia one last thing, could you update the changelog with this fix https://github.com/coreos/etcd-operator/blob/master/CHANGELOG.md#fixed

PR #1979, opened 2018-07-09
allow the curl init container to be variable

kodieGlosser:


kodieGlosser:
@hexfusion for us, we use a curl image in a different registery. All of our images come from the same registry so we have to rely on less depencies when running in production

PR #1842, opened 2018-01-09
*: add ABS support for backup and restore

rjtsdl:
Add ABS support for backup-operator and restore-operator
Fix #1784

PR #1981, opened 2018-08-02
Add cloud environment support for ABS backup

feiskyer:
PR #1842 adds support for ABS (Azure Blob Service) backup. But it always using AzurePublicCloud, so that other sovereign clouds (e.g. mooncake) won't work.

This PR adds an extra option `cloud` to ABS credential and use `storage.NewBasicClientOnSovereignCloud()`  to fix the issue. If `cloud` is not set, then AzurePublicCloud is used (same as before).

Example ABS credential now is

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: abs-credentials
type: Opaque
stringData:
  storage-account: <storage-account-name>
  storage-key: <storage-key>
  cloud: AzurePublicCloud
```

### DIFF
--- a/pkg/util/azureutil/absfactory/client.go
+++ b/pkg/util/azureutil/absfactory/client.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 
 	"github.com/Azure/azure-sdk-for-go/storage"
+	"github.com/Azure/go-autorest/autorest/azure"
 	api "github.com/coreos/etcd-operator/pkg/apis/etcd/v1beta2"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


feiskyer:
cc @xiang90

pkg/util/azureutil/absfactory/client.go
```diff
@@ -18,6 +18,7 @@ import (
 	"fmt"
 
 	"github.com/Azure/azure-sdk-for-go/storage"
+	"github.com/Azure/go-autorest/autorest/azure"
```
cpuguy83: Right now it looks like go-autorest is a transient dependency, this seems to make it a direct dependency.
Should this include an update to Gopkg.toml?

feiskyer: Run `dep ensure` doesn't add that. Should I add it explicitly?

fanminshi: No need to add go-autorest. it is vendored already via https://github.com/coreos/etcd-operator/blame/master/Gopkg.lock#L17

fanminshi:
I don't see a doc on how to use the ABS as the backup backend. Could you add a doc under `doc/user/backup` on how to do backup using ABS? We can do that at a separate pr.

fanminshi:
lgtm on this code.

mwieczorek:
@fanminshi there’s 
https://github.com/coreos/etcd-operator/blob/master/doc/design/abs_backup.md
But of course it should be updated and maybe moved to doc/user folder?

fanminshi:
@mwieczorek good find. I think we should leave the design doc alone. Let's just copy and paste the design content and update it with more details on how to use ABS as the backend for public cloud, sovereign cloud, and etc.

feiskyer:
@fanminshi @mwieczorek Thanks. Let me add the user guide for it in a separate PR.

PR #1986, opened 2018-08-09
added test to add label on etcd cluster services

prune998:
When creating Etcd clusters, we want to be able to differentiate the `peer` service and the `client` service using labels.
The main reason is that Prometheus-Operator can't easily differentiate on service names, only on labels. 
This PR is adding a specific label that tels the `peer` service can be used to scrape metrics.


prune998:
poor @etcd-bot, all alone managing the PRs 🤖

PR #1988, opened 2018-08-13
Add documentation for ABS backup

### LINKED ISSUE #1987: Document ABS backup
We have a [design doc](https://github.com/coreos/etcd-operator/pull/1981#issuecomment-411841840) for ABS backup which can be outdated if we add new features to ABS backup.  We should have a `abs_backup.md` under `doc/user/` to document how to use ABS as the backup backend.

cc/ @mwieczorek @feiskyer

feiskyer:
Closes #1987

/cc @fanminshi @mwieczorek

### DIFF
--- a/doc/user/abs_backup.md
+++ b/doc/user/abs_backup.md
@@ -1,0 +1,164 @@
+# Backups using Azure Blob Service (ABS)
+
+Etcd backup operator backups the data of an etcd cluster running on Kubernetes to a remote storage such as Azure Blob Service (ABS). If it is not deployed yet, please follow [here](walkthrough/backup-operator.md#deploy-etcd-backup-operator) to deploy it, e.g. by running
+
+```sh
+kubectl apply -f example/etcd-backup-operator/deployment.yaml
+```
+
+## Setup Azure Secret
+
+Create a secret file `abs-secret.yaml`:
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: abs-credentials
+  namespace: etcd
+type: Opaque
+stringData:
+  storage-account: "<storage-account>"
+  storage-key: "<storage-key>"
+  cloud: "<cloud>"
+```
+
+In above secret, following data should be provided:
+
+- `storage-account`: the name of Azure storage account.
+- `storage-key`: the storage key of storage account.
+- `cloud`: the name of cloud environment, e.g. `AzurePublicCloud` or `AzureChinaCloud`. Optional. If not specified, it is defaulted to `AzurePublicCloud`.
+
+And then create the secret:
+
+```sh
+kubectl apply -f abs-secret.yaml
+```
+
+## Create EtcdBackup CR
+
+First, create a container in the storage account:
+
+```sh
+az storage container create -n etcd-backups --account-name <storage-account> --account-key <storage-key>
+```
+
+Then create a `EtcdBackup` CR file `etcdbackup.yaml` which uses secret `abs-credentials`:
+
+```yaml
+apiVersion: "etcd.database.coreos.com/v1beta2"
+kind: "EtcdBackup"
+metadata:
+  name: example-etcd-cluster-backup
+  namespace: etcd
+spec:
+  etcdEndpoints: ["http://example-etcd-cluster-client:2379"]
+  storageType: ABS
+  abs:
+    # The format of the path must be: "<abs-container-name>/<path-to-backup-file>"
+    path: etcd-backups/etcd.backup
+    absSecret: abs-credentials
+```
+
+Finally apply it to kubernetes cluster:
+
+```sh
+kubectl apply -f etcdbackup.yaml
+```
+
+Check the `status` section of the `EtcdBackup` CR:
+
+```sh
+$ kubectl get EtcdBackup example-etcd-cluster-backup -o yaml
+apiVersion: etcd.database.coreos.com/v1beta2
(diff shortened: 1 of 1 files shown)


doc/user/abs_backup.md
```diff
@@ -0,0 +1,95 @@
...
+  storage-account: "<storage-account>"
+  storage-key: "<storage-key>"
+  cloud: "<cloud>"
+```
+
+In above secret, following data should be provided:
+
+- `storage-account`: the name of Azure storage account.
+- `storage-key`: the storage key of storage account.
+- `cloud`: the name of cloud environment, e.g. `AzurePublicCloud` or `AzureChinaCloud`.
```
fanminshi: If `cloud` is not specified, it is default to `AzurePublicCloud` right?

mwieczorek: And should be marked as optional field, not required

doc/user/abs_backup.md
```diff
@@ -0,0 +1,95 @@
...
+
+And then create the secret:
+
+```sh
+kubectl apply -f abs-secret.yaml
+```
+
+## Create EtcdBackup CRD
```
mwieczorek: It’s just a naming, and I’m just nitpicking, but is it CRD (definition) or just an instance of CRD? So maybe just resource instead of CRD?

mwieczorek:
And general comment: do we want also EtcdRestore describe in the same doc?

doc/user/abs_backup.md
fanminshi: Agree. it should be CR instead of CRD.

feiskyer: yep, it should be CR

doc/user/abs_backup.md
feiskyer: Let me address it as optional.

feiskyer:
Addressed comments in the new commit.

@mwieczorek  @fanminshi While backup works properly, restore doesn't work in my cluster. Some foundings:

Etcd restore CR has changed its status to success

```
kubectl get etcdrestore example-etcd-cluster -o yaml 
kind: EtcdRestore
...
status:
  succeeded: true
```

And new pod has also been created

```
 kubectl get pod -l app=etcd,etcd_cluster=example-etcd-cluster
NAME                              READY     STATUS     RESTARTS   AGE
example-etcd-cluster-khv927lqq6   0/1       Init:0/3   0          6m36s
```

But the pod is always in PodInitializing state because `check-dns` init container failed to resolve dns 'example-etcd-cluster-khv927lqq6.example-etcd-cluster.etcd.svc':

```
nslookup: can't resolve 'example-etcd-cluster-khv927lqq6.example-etcd-cluster.etcd.svc'
nslookup: can't resolve 'example-etcd-cluster-khv927lqq6.example-etcd-cluster.etcd.svc'
Server:    10.0.0.10
Address 1: 10.0.0.10 kube-dns.kube-system.svc.cluster.local

Server:    10.0.0.10
Address 1: 10.0.0.10 kube-dns.kube-system.svc.cluster.local

nslookup: can't resolve 'example-etcd-cluster-khv927lqq6.example-etcd-cluster.etcd.svc'
Server:    10.0.0.10
Address 1: 10.0.0.10 kube-dns.kube-system.svc.cluster.local

nslookup: can't resolve 'example-etcd-cluster-khv927lqq6.example-etcd-cluster.etcd.svc'
```

@mwieczorek  @fanminshi Do you know why this is happening and how to fix this?

mwieczorek:
@feiskyer I haven't seen such issues. Which versions are you using (etcd-operator/etcd)?

@fanminshi I guess restore operator sets success status before etcd cluster gets healthy. Is it expected behaviour?

feiskyer:
@mwieczorek I'm using operators build from master branch. Seems headless service is removed by etcd-operator before restoring etcdcluster.

feiskyer:
@fanminshi Yep, it always fail in my testings. Should I add headless service creation first before creating pod `example-etcd-cluster-khv927lqq6`?

fanminshi:
@feiskyer 
`Seems headless service is removed by etcd-operator`
In restore operator, it [deletes the cluster CR](https://github.com/coreos/etcd-operator/blob/master/pkg/controller/restore-operator/sync.go#L131), and in return will delete any resources that's has the ownerRef of the cluster CR. I think that the headless svc has the ownerref of the cluster CR. Hence, headless service is also removed.

` Should I add headless service creation first before creating pod example-etcd-cluster-khv927lqq6` No, you don't need to do that. if you take a look of test for restore [testEtcdRestoreOperatorForS3Source](https://github.com/coreos/etcd-operator/blob/master/test/e2e/e2eslow/backup_restore_test.go#L170), the user doesn't need to create the headless service. The etcd-operator should have handled that.

I am currently investigating this issue. FYI. 

this pr looks good. The restore failure is another issue. Could you open it as a separate issue and so we can track it.

fanminshi:
@mwieczorek `I guess restore operator sets success status before etcd cluster gets healthy. Is it expected behaviour?`
The current code reports succeed if the seed pod is successfully created, However, it doesn't check pod is running or not. I would say the the behavior is not expected. The restore operator should only report succeed if the deployed pod is running. Could you file an issue to track that.

mwieczorek:
@fanminshi I created #1993 to track the Etcdrestore status issue.

fanminshi:
Thanks for the PR. merging!

PR #2021, opened 2018-12-06
Add support to use IAM roles without credentials on s3 backup and restore

### LINKED ISSUE #1682: docs: etcd backup/restore IAM roles
Users will want to scope credentials down for backup/restore to only the rights that are required. We need to document that scope for users and add it to the [backup](https://github.com/coreos/etcd-operator/blob/master/doc/user/walkthrough/backup-operator.md) and [restore](https://github.com/coreos/etcd-operator/blob/master/doc/user/walkthrough/restore-operator.md) docs.

Example docs

- https://docs.databricks.com/user-guide/cloud-configurations/aws/iam-roles.html
- https://aws.amazon.com/blogs/security/iam-policies-and-bucket-policies-and-acls-oh-my-controlling-access-to-s3-resources/

slok:
Fixes #1682 

This PR allows not using credentials and use instead node IAM roles or applications like Kiam or Kube2iam.

- Add new S3 client without credentials if not `awsSecret` present.
- The region when not using `awsSecret` will be inferred from the S3 endpoint.
- Add documentation and example of how to use based on IAM roles.
- Format some parts of the docs markdown.

I added Minio-go dependency only to infer the region,  I thought it would be cleaner and had fewer corner cases/errors than implementing myself as it has a little [bit of logic](https://github.com/minio/minio-go/blob/c085651199f9403929babaa1e6524a53085d2395/pkg/s3utils/utils.go#L99) depending on the kind of regions/cases.

### DIFF
--- a/pkg/util/awsutil/s3factory/client.go
+++ b/pkg/util/awsutil/s3factory/client.go
@@ -17,15 +17,18 @@ package s3factory
 import (
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"path"
 
-	api "github.com/coreos/etcd-operator/pkg/apis/etcd/v1beta2"
-
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/minio/minio-go/pkg/s3utils"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+
+	api "github.com/coreos/etcd-operator/pkg/apis/etcd/v1beta2"
 )
 
 const (
@@ -38,8 +41,61 @@ type S3Client struct {
 	configDir string
 }
 
-// NewClientFromSecret returns a S3 client based on given k8s secret containing aws credentials.
-func NewClientFromSecret(kubecli kubernetes.Interface, namespace, endpoint, awsSecret string) (w *S3Client, err error) {
+// ClientConfig is the configuration for a S3 client.
+type ClientConfig struct {
+	Endpoint  string
+	Namespace string
+	AWSSecret string
+}
+
+// NewClient returns a new S3Client. This client can be based on profiles, keys...
+// (from secret), or IAM Role based (nodes, Kube2iam, Kiam...), the client
+// type will be infered based on the missing (or not missing) data in the
+// client configuration.
+func NewClient(cfg ClientConfig, kubecli kubernetes.Interface) (*S3Client, error) {
+	// If credentials not required then use the S3 client based on IAM roles.
+	if cfg.AWSSecret == "" {
+		return newClientForIAMRole(cfg.Endpoint)
+	}
+
+	return newClientFromSecret(kubecli, cfg.Namespace, cfg.Endpoint, cfg.AWSSecret)
+}
+
+// newClientForIAMRole returns a S3 client that doesn't have credentials set and will do that
+// AWS S3 client fallback to machines assigned IAM role (or in case of using Kube2iam, Kiam...,
+// use the methods used to act on befhalf of the assumed roles given by these systems).
+func newClientForIAMRole(endpoint string) (w *S3Client, err error) {
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("new S3 client failed: %v", err)
+		}
+	}()
+
+	// Get the region based on the endpoint.
+	region := s3utils.GetRegionFromURL(url.URL{
+		Host: endpoint,
+	})
+
+	if region == "" {
+		return nil, fmt.Errorf("could not infer region from '%s' s3 endpoint, use an endpoint with region like `s3.us-east-1.amazonaws.com`", endpoint)
+	}
+
+	// Create AWS session default the config files, profile...
+	sess, err := session.NewSession(&aws.Config{
+		Endpoint: aws.String(endpoint),
+		Region:   aws.String(region),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return &S3Client{
+		S3: s3.New(sess),
+	}, nil
+}
+
+// newClientFromSecret returns a S3 client based on given k8s secret containing aws credentials.
+func newClientFromSecret(kubecli kubernetes.Interface, namespace, endpoint, awsSecret string) (w *S3Client, err error) {
 	defer func() {
 		if err != nil {
 			err = fmt.Errorf("new S3 client failed: %v", err)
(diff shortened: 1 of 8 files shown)


hexfusion:
@etcd-bot ok to test

pkg/util/awsutil/s3factory/client.go
```diff
@@ -38,8 +41,61 @@ type S3Client struct {
...
+type ClientConfig struct {
+	Endpoint  string
+	Namespace string
+	AWSSecret string
+}
+
+// NewClient returns a new S3Client. This client can be based on profiles, keys...
+// (from secret), or IAM Role based (nodes, Kube2iam, Kiam...), the client
+// type will be infered based on the missing (or not missing) data in the
+// client configuration.
```
hexfusion: s/infered/inferred

> based on the missing (or not missing) data in the
// client configuration

based on the ClientConfig.

pkg/util/awsutil/s3factory/client.go
```diff
@@ -38,8 +41,61 @@ type S3Client struct {
...
 }
 
-// NewClientFromSecret returns a S3 client based on given k8s secret containing aws credentials.
-func NewClientFromSecret(kubecli kubernetes.Interface, namespace, endpoint, awsSecret string) (w *S3Client, err error) {
+// ClientConfig is the configuration for a S3 client.
+type ClientConfig struct {
+	Endpoint  string
+	Namespace string
+	AWSSecret string
+}
```
hexfusion: Should AWSSecret be more generically something like Auth and direct traffic/validate with for example a switch? Just thinking how we make this more flexible moving forward.

```golang
switch cfg.Auth.Type {
	case "IAMRole":
            return newClientForIAMRole(cfg.Endpoint)
        case "Secret":
           return newClientFromSecret(...
        default:
		// oops  
	}
```

slok: I don't know if I understood correctly.

Who would set that information? should the upper layer decide what client type it wants and tell to the factory method what client wants?

(Independently of the type we need `AWSSecret` in case of config base auth to know the name of the secret to reference when downloading and creating an AWS config file)

slok:
I didn't know that this app was used like (or affects as) a library so we mark the internal changes as breaking. Sorry about that. I'll add to the changelog!

pkg/util/awsutil/s3factory/client.go
hexfusion: @slok let me rephrase. Since we now create ClientConfig object (which I like) what happens if a 3rd authentication option needs to be added? Currently we only have 2 options and perhaps we should make this more extensible since we are already making this change.

So I was thing something more like this might work.
```
type ClientConfig struct {
	Endpoint  string
	Namespace string
	Auth
}

type Auth struct {
	Type  string
	Value string
}
```
What do you think? Then you can set Auth{Type: "AWSSecret", Value: "Foo"} we need to think about this a bit but I would like to move in this direction in general if possible and I would like to hear your thoughts.

slok: Oh! I see I didn't understand at first. I like it, The only caveat that I see here is that depending on the Auth-type could require a different number and kind of fields.

hexfusion: Yes this requires thought to do correctly so we should not be in a rush to change this.

danbeaulieu:
@hexfusion From my pov IAM role based access is much more typical than using keys and secrets. That said the auth config for the go sdk is well defined and IAM is documented as preferred https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html.

This code is already a bit odd with the requirement to create a shared config file as a secret and then write that out to the fs when it **isn't** one of the default credential providers for the go sdk.

Here is one proposal to help push this PR forward:
- Do the file management stuff automatically if the optional secret is set
- Create a [custom credential chain](https://docs.aws.amazon.com/sdk-for-go/api/aws/credentials/#NewChainCredentials) that includes the shared config file and IAM provider.
- Create the S3 client from the custom credential provider

Doing it this way delegates to the SDK and means a custom Auth object isn't needed?

pkg/util/awsutil/s3factory/client_test.go
```diff
@@ -38,3 +38,45 @@ func TestSetupAWSConfig(t *testing.T) {
...
+			name: "A configuration without AWS secret, should return a client without credentials.",
+			cfg: ClientConfig{
+				Endpoint: "s3.eu-west-1.amazonaws.com",
+			},
+			expErr: false,
+		},
+		{
+			name: "A configuration without AWS secret and bad endpoint without region, should return a error.",
+			cfg: ClientConfig{
+				Endpoint: "s3.amazonaws.com",
```
danbeaulieu: Curious why this is considered a bad endpoint? This is the default endpoint in the SDK, as an S3 user I'd expect it to be valid.

slok: Although S3 is multi-region, you need to specify the region, so we don't need to add another option like the `region` (this way improving the configuration usage experience), and you provide directly the region on the url (is a common practice). The [error](https://github.com/coreos/etcd-operator/pull/2021/files#diff-cc5b6d77b89214ec48aaee598700dd5bR78) is self-descriptive to let the user know what's happening.

danbeaulieu: This may be out of scope for this PR but having the default endpoint cause an error is counter intuitive. Why do you need to specify the region if S3 is multi-region and s3.amazonaws.com is valid? It is not required to pass in an endpoint or a region when constructing a client.

Ideally the logic would support using straight defaults for everything or an optional configured endpoint and if that endpoint specifies a region then also the region. 

Maybe one solution is to use the IsAmazonEndpoint function in that minio utility. If that is true set the region to us-east-1?

slok: Replication on different AWS regions could be hours. In that case, a random default zone could be failing for hours although you know what's your main region.

And a region is required always for clients, another thing is that people have on env vars, or profiles their default zones at the moment of executing applications: https://github.com/aws/aws-sdk-go/blob/master/aws/config.go#L48-L49

dkhenry: I agree with @danbeaulieu on this one. Making the default fail doesn't make sense. I should be able to configure this without specifying an endpoint.

yzargari:
Does anyone have ETA on this? I need to setup backups and really don't want to create a user account with keys just to do that...

Thanks!

jescarri:
Hey Any progress on this?.

Br3wm4ster:
Hey guys any help required here to get this merged?

I've been running this PR + Prometheus exporter one in prod for a month and it's been working w/o problems.

slok:
@Br3wm4ster We have also been running this fork for the backups for several months (since the PR) and we didn't have any problems neither.

Apparently, this seems to be blocked/stopped because it wasn't optimal (although _optimal_ is a matter of opinion, like _perfection_ or software design, I guess) and we are waiting so the maintainers have some spare cycles to revisit this.

By the way, I can help fixing the things that are needed so we can have this feature on master or solving the conflicts with master at this moment.

PR #2025, opened 2018-12-13
example: added "root secure" deployment yaml

qvicksilver:
Allows deploying in clusters with PodSecurityPolicy enforcing
runAsNonRoot without getting the following error:
“Error: container has runAsNonRoot and image has non-numeric
user (etcd-operator), cannot verify user is non-root”


hexfusion:
lgtm @qvicksilver thanks!

PR #2029, opened 2018-12-18
Use endpoint status for probe checks

### LINKED ISSUE #1957: etcd pod health/liveliness probe fails if auth is enabled
The current probe used for determining etcd liveliness/health is a linearizable get.
https://github.com/coreos/etcd-operator/blob/3ceb27aafe674420af6767c946734cf839c0efbc/pkg/util/k8sutil/pod_util.go#L72-L73
This ensures that the etcd pod is ready to participate in consensus and serve requests.

After auth is enabled on a deployed etcd cluster the etcd-pod's probe will fail since the request is not authenticated.

One solution is to replace the linearizable get with `endpoints health`.
https://github.com/coreos/etcd/tree/master/etcdctl#endpoint-health

For this to work we need to ensure that `endpoints health` would give the same guarantees as a linearizable get, and that it works on an auth enabled cluster.

avorima:
`endpoint status` doesn't require a username when authentication is enabled.

Fixes #1957


hexfusion:
@etcd-bot ok to test

hexfusion:
@avorima I appreciate the PR this seems fine the result is the same and as you noted it allows auth. Can you please update the commit message to read `pkg/*: Use endpoint status for probes` also update [CHANGELOG](https://github.com/coreos/etcd-operator/blob/master/CHANGELOG.md) in a separate commit. Thanks!

avorima:
@hexfusion Thanks for the quick response. Hope this is ok.

hexfusion:
@avorima OK we are really close, but we need to get rid of the merge commit. And we would like all commit messages to be `$package/: $message`

try
```
# remove merge commit
git reset --hard HEAD~1

# rename CHANGELOG commit
git commit --amend -m'CHANGELOG: updated to reflect probe fix'

# sanity check
git log

git push -f

```
Thanks!

avorima:
@hexfusion What about the merge conflict?

hexfusion:
lgtm, thanks @avorima!

hexfusion:
@avorima I have run into an issue merging your commit because it has no email address.

please update with interactive rebase.
```
git rebase -i HEAD~2
```
you will see something like
```bash
pick 7c7141550 pkg/*: Use endpoint status for probes
pick a580ec454 CHANGELOG: updated to reflect probe fix
```
change **pick** to **edit** for both and save this will start the rebase.
# for each 
git commit --amend --author="Mario Valderrama <my@email.com>"
git rebase --continue
when done git push -f
```
Then please make sure to update your email address for future commits.
https://help.github.com/articles/setting-your-commit-email-address-in-git

Once this is done I can merge, thanks!

avorima:
My bad, here you go.

PR #2037, opened 2019-01-10
pkg/*: allow to add extra arguments to cluster members

aanm:
With this changes, users will be able to specifiy extra arguments on all
cluster members running in the etcd cluster.

Signed-off-by: André Martins <aanm90@gmail.com>

### DIFF
--- a/pkg/apis/etcd/v1beta2/cluster.go
+++ b/pkg/apis/etcd/v1beta2/cluster.go
@@ -99,6 +99,9 @@ type ClusterSpec struct {
 
 	// etcd cluster TLS configuration
 	TLS *TLSPolicy `json:"TLS,omitempty"`
+
+	// ExtendedArgs is a form to add more arguments that can be used for all etcd instances
+	ExtendedArgs string `json:"extendedArgs,omitempty"`
 }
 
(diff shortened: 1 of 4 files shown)


hexfusion:
@aanm thanks for the PR can you give a use case for this change?

aanm:
@hexfusion absolutely, we were trying to have automatic compaction set in the etcd-cluster by default, instead of adding a field to enable/ disable the automatic compaction I thought it would be better to allow to add any extra argument for all etcd instances. This way we can have something like `ExtendedArgs: "--auto-compaction-mode=revision --auto-compaction-retention=25000"` which will work perfectly for us.

More info:
https://github.com/cilium/cilium-etcd-operator

hexfusion:
@aanm we currently allow passing ENV args to etcd like below would that work for you?

```
  pod:
    etcdEnv:
    - name: ETCD_AUTO_COMPACTION_MODE
      value: "revision"
    - name: ETCD_AUTO_COMPACTION_RETENTION
      value: 25000
```

aanm:
@hexfusion Awesome! That work perfectly!

PR #2059, opened 2019-02-27
dep: bump k8s to 1.12.6 and regenerate codegen

hexfusion:
This PR bumps k8s to 1.12.6, golang to 1.11.5 and regenerated codegen.

Signed-off-by: Sam Batschelet <sbatsche@redhat.com>

### DIFF
--- a/pkg/controller/backup-operator/util.go
+++ b/pkg/controller/backup-operator/util.go
@@ -46,3 +46,12 @@ func isPeriodicBackup(ebSpec *api.BackupSpec) bool {
 	}
 	return false
 }
+
+func containsString(slice []string, s string) bool {
+	for _, item := range slice {
+		if item == s {
+			return true
+		}
+	}
+	return false
+}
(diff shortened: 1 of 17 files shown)


hexfusion:
@hasbro17 PTAL

hasbro17:
SGTM. We should also update the CHANGELOG to point out the update to k8s 1.12.6.

pkg/controller/backup-operator/util.go
```diff
@@ -46,3 +46,15 @@ func isPeriodicBackup(ebSpec *api.BackupSpec) bool {
 	}
 	return false
 }
+
+func containsString(slice []string, s string, modifier func(s string) string) bool {
```
hasbro17: Since we're not really using the modifier argument anywhere right now, probably better to simplify this function to a simpler `func containsString(slice []string, s string) bool` and add the modifier later when we actually need it.

hexfusion: ack will do.

PR #2063, opened 2019-03-04
pkg/util: Don't expose unready nodes via client service

### LINKED ISSUE #2030: Client Service "tolerate-unready-endpoints" annotation causes connection issues
I don't know whether this is just a question, or a bug in etcd-operator, or a bug in kubernetes itself.

In any case, I was just playing with a 3 node etcd cluster and the impact on clients when I delete one of the cluster members.

I use a simple command like `etcdctl get --prefix /` in a loop, while I delete one pod, and wait for a replacement pod to appear.

The problem happens when the new pod appears, it is not yet Ready, but the Service endpoints include the new pod:

```
$ kubectl get ep example-etcd-cluster-client
[...]
example-etcd-cluster-client   10.125.27.21:2379,10.125.6.10:2379,10.126.8.18:2379   41m
```

But at the same time the pod is still initialising:

```
example-etcd-cluster-kbn69vn2xn   0/1     Init:0/1   0          6s    10.125.6.10    hex-48b-pm.k2.gambit   <none>           <none>
```

As a result, my etcdctl gets either delayed for 1 second or so, or gets an error, or works fast, randomly.

As soon as the new pod finishes initialising, the service is restored.

To be honest, to be this seems like a Kubernetes bug -- it shouldn't add a pod to a Service endpoints list before the pod is Ready -- but I could be missing something.  Otherwise, how can you have zero downtime with etcd, with expected node maintenance requiring me to evict pods from nodes, once in a while?

Any thoughts?

ckoehn:
Previously unready etcd nodes were already receiving client connections
although they are still in the initiation phase and not able to accept
any traffic. This caused connection failure or high latency.

Fixes #2030

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Fixed
 
+- Don't expose unready nodes via client service. [#2063](https://github.com/coreos/etcd-operator/pull/2063)
+
 ### Deprecated
 
 ### Security
--- a/pkg/util/k8sutil/k8sutil.go
+++ b/pkg/util/k8sutil/k8sutil.go
@@ -20,6 +20,7 @@ import (
 	"net"
 	"net/url"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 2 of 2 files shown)


hexfusion:
@etcd-bot ok to test

ckoehn:
@hexfusion Any plans to merge this?

kannanvr:
Hi All,
   I think, this code change may  solve  the problem which i faced recently on my cluster....
Request you to merge this code change as early as possible....


Thanks,

hexfusion:
@ckoehn sorry for the delay I will take a final look at this before the end of the week.

hasbro17:
@ckoehn Seems good to me. Can you please rebase from the master.
And can you also please add a line to the CHANGELOG to point out this fix.
https://github.com/coreos/etcd-operator/blob/master/CHANGELOG.md#fixed

ckoehn:
@hasbro17 Done. Thanks for your review.

PR #2071, opened 2019-04-03
Azure blob storage: use correct list prefix

kapouille:
Azure bloc storage implementation didn't use the right prefix resulting on no backup cleanup being performed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixed
 
 - Don't expose unready nodes via client service. [#2063](https://github.com/coreos/etcd-operator/pull/2063)
+- Azure blob storage: use correct list prefix [#2071](https://github.com/coreos/etcd-operator/pull/2071)
 
 ### Deprecated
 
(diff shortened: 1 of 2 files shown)


hexfusion:
@etcd-bot ok to test

hexfusion:
@kapouille thanks for the PR. As we do not have testing for Azure setup can you provide a bit more details on how this resolves the situation? 

@rjtsdl I see you are with Oracle now but as a maintainer of Azure ABS feature could you please comment/review.

kapouille:
of course @hexfusion . In order to list the existing backups and see which ones are out of date or not, it used to retrieve the wrong url, *including* the container name (`<my_container>/<path>`). That would return an empty list and therefore never clean up older backups as the azure API only expects `<path>`

kapouille:
(apologies about the force-push, really should have done that work in the branch)

kapouille:
Putting on hold, noticed an odd behaviour where it is keeping the **oldest** backups rather than the most recent. Investigating.

kapouille:
Could not repro issue appears to be a one-off due to changing file naming conventions.

hasbro17:
@kapouille Sorry this has been open for a while. This seems good to me. Can you please rebase and also add a line in the Fixed section of the [CHANGELOG](https://github.com/coreos/etcd-operator/blob/master/CHANGELOG.md#fixed) for this.

hasbro17:
@kapouille pinging again to see if you have time to update the PR. If not then I can close this out and resubmit the bugfix.

kapouille:
Apologies, will get on that tomorrow.

kapouille:
I believe this is now done.

PR #2075, opened 2019-04-05
backup operator: fixed incorrect backup file naming convention.

kapouille:
When limiting the number of backups, if the revision would move to needing another digit, the logic would never select a new entry as _ is > to 9 in ascii terms.

Note that this is backwards incompatible as now files receive a correct padding and old revisions will still be incorrectly handled.


hexfusion:
@etcd-bot ok to test

mikenorgate:
@kapouille it looks like we missed this line in the tests

https://github.com/coreos/etcd-operator/blob/86211e9f3bc32cd97187c7ee9beb32be60264fed/test/e2e/e2eslow/backup_restore_test.go#L55

Issue #1974, opened 2018-06-25
Version selection assumes etcd image versions will begin with "v"


perotinus:
The code that constructs the etcd image version from the CRD spec prefixes the image version with a "v". This assumption breaks in some contexts where the released etcd images from quay.io are not beins used, and it would be useful to have a way to tell the operator to take the image version as-is, without inserting the extra "v" into the versioning scheme.

https://github.com/coreos/etcd-operator/blob/7fbe23cf50900e6282255978ae9930a7750d9238/pkg/util/k8sutil/k8sutil.go#L138

perotinus:
Looks like #1970 addresses this.

prune998:
made a PR for the `v` addition to the version : https://github.com/coreos/etcd-operator/pull/2079

Issue #1960, opened 2018-05-04
Deploying a cluster with different repository/version fails on pull


carlosedp:
I'm deploying an etcd-operator on ARM64 (compiled and built my own images) but when trying to deploy the etcd cluster with the Google provided etcd images from _gcr.io/google-containers/etcd-arm64_, I specify the image version "3.2.18" but etcd-operator adds a "v" to the version trying to pull ` gcr.io/google-containers/etcd-arm64:v3.2.18 ` failing because gcr doesn't prefix the versions with "v".

I understand this was modeled after CoreOS image store on Quay where the images have the "v" after before the SemVer but it would be more flexible if the repository and version could just be concatenated.

prune998:
Have the exact same issue here... 

Also, when using private repo + auth, we need to add to the Deployment Spec : 
```
"imagePullSecrets": [
          {
            "name": "your-auth-secret"
          }
        ],
```

Which is not taken care of in the `PodPolicy` spec of the `etcdCluster` (I could provide a patch for this and the `v` issue above if maintainers agree for this change.

Thanks

prune998:
Will also work on a PR to support imagePullSecrets

prune998:
The `ImagePullSecrets` can be set in a `ServiceAccount` as described in https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-imagepullsecrets-to-a-service-account

Would it be a better solution to allow to set a `ServiceAccount` in the `etcdCluster` resource ?

PR #2079, opened 2019-04-17
removed auto-adding v to the etcd image version

prune998:
When creating an `etcdCluster` resource, you provide an `Image` and a `Version.
`Version` is a semver version depending on the EtcD image version you wish to use. 

The default is to use Quay images, which are versioned using a pattern of `v` + `semver`, like `v3.0.12`.

So defining :
```
kind: "EtcdCluster"
metadata:
  name: "example-etcd-cluster"
spec:
  size: 3
  version: "3.1.10"
  repository: "quay.io/coreos/etcd"
```

Will create a Pod using image `quay.io/coreos/etcd:v3.1.10`.

This may be fine with `Quay` images, but when you are working with other Docker Registries, you may not want to add the `v` to the version.

The Operator should not force you to use a specific version pattern.

This PR removes the `v` addition to the version, allowing to use a real version name in the CRD.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ metadata:
   name: "example-etcd-cluster"
 spec:
   size: 5
-  version: "3.2.13"
+  version: "v3.2.13"
 ```
 
 Apply the size change to the cluster CR:
@@ -129,7 +129,7 @@ metadata:
   name: "example-etcd-cluster"
 spec:
   size: 3
-  version: "3.2.13"
+  version: "v3.2.13"
 ```
 ```
 $ kubectl apply -f example/example-etcd-cluster.yaml
@@ -222,7 +222,7 @@ metadata:
   name: "example-etcd-cluster"
 spec:
   size: 3
-  version: "3.1.10"
+  version: "v3.1.10"
   repository: "quay.io/coreos/etcd"
 ```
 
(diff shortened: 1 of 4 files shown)


prune998:
issue referenced in 
- https://github.com/coreos/etcd-operator/issues/1974
- https://github.com/coreos/etcd-operator/issues/1960

alaypatel07:
I am wondering if instead of composing the image name using `repository` and `version`, it would be simpler to replace it with a key called `image`? The CR will look like:

```
kind: "EtcdCluster"
metadata:
  name: "example-etcd-cluster"
spec:
  size: 3
  image: "quay.io/coreos/etcd:v3.1.10"
```

prune998:
Maybe doing this will make it harder to manage the upgrade process when only the image version is changed ? 
That would also imply an API change, while this PR does not.

alaypatel07:
I could be wrong, but I thought if we can compare the image instead of version [here](https://github.com/coreos/etcd-operator/blob/86211e9f3bc32cd97187c7ee9beb32be60264fed/pkg/cluster/reconcile.go#L215), the upgrade process can be handled.

Agreed, this would mean an API change, and I do not know if we would want it for simplicity or not.

xmlking:
we also have issue with `auto-adding v `
our `Pod Security Policy`  restrict images using with version number. we have to sign them and we have to use tag with digest something like this  `"sha256:4567464gfjfjfgjfjfjfjfjjfjfjfjfj354345534"` if it adds "v" , then it breaks

NormanFoxLee:
Is it possible to add `digest:` so we can use `image@digest` instead `image:version`?

PR #2082, opened 2019-04-30
Allow specifying Cluster Domain for clientURLs

mikkeloscar:
This allows setting the cluster domain as a flag on the operator. The
cluster domain is used as a suffix in the Client URLs for the etcd
members.

The ability to set a custom cluster domain is desirable when running in
clusters with a custom DNS configuration.

In our case we need this because we have decided to default the DNS config to use `ndots:2` instead of the default `ndots:5` setting which can result in up to 10 additional DNS queries for every single name lookup, which causes an unnecessary overhead for the DNS infrastructure and for applications.

Ref: https://github.com/kubernetes/dns/issues/159#issuecomment-336449931

An alternative way to fix this would be to allow specifying the DNS config of the pod spec: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-config

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -316,6 +316,9 @@ func (c *Cluster) startSeedMember() error {
 		SecurePeer:   c.isSecurePeer(),
 		SecureClient: c.isSecureClient(),
 	}
+	if c.cluster.Spec.Pod != nil {
+		m.ClusterDomain = c.cluster.Spec.Pod.ClusterDomain
+	}
 	ms := etcdutil.NewMemberSet(m)
 	if err := c.createPod(ms, m, "new"); err != nil {
 		return fmt.Errorf("failed to create seed member (%s): %v", m.Name, err)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Added
 
+- Added `spec.Pod.ClusterDomain` to explicitly set the cluster domain used for the etcd member URLs. [#2082](https://github.com/coreos/etcd-operator/pull/2082)
+
 ### Changed
 
 ### Removed
--- a/pkg/apis/etcd/v1beta2/cluster.go
+++ b/pkg/apis/etcd/v1beta2/cluster.go
@@ -157,6 +157,11 @@ type PodPolicy struct {
 	// reverse DNS lookup its IP given the hostname.
 	// The default is to wait indefinitely and has a vaule of 0.
 	DNSTimeoutInSecond int64 `json:"DNSTimeoutInSecond,omitempty"`
+
+	// ClusterDomain is the cluster domain to use for member URLs E.g.
+	// '.cluster.local'.
+	// The default is to not set a cluster domain explicitly.
+	ClusterDomain string `json:"ClusterDomain"`
 }
 
 // TODO: move this to initializer
--- a/pkg/cluster/member.go
+++ b/pkg/cluster/member.go
@@ -51,12 +51,17 @@ func (c *Cluster) updateMembers(known etcdutil.MemberSet) error {
 
 func (c *Cluster) newMember() *etcdutil.Member {
 	name := k8sutil.UniqueMemberName(c.cluster.Name)
-	return &etcdutil.Member{
+	m := &etcdutil.Member{
 		Name:         name,
 		Namespace:    c.cluster.Namespace,
 		SecurePeer:   c.isSecurePeer(),
 		SecureClient: c.isSecureClient(),
 	}
+
+	if c.cluster.Spec.Pod != nil {
+		m.ClusterDomain = c.cluster.Spec.Pod.ClusterDomain
+	}
+	return m
 }
 
 func podsToMemberSet(pods []*v1.Pod, sc bool) etcdutil.MemberSet {
--- a/pkg/controller/restore-operator/sync.go
+++ b/pkg/controller/restore-operator/sync.go
@@ -217,6 +217,9 @@ func (r *Restore) createSeedMember(ec *api.EtcdCluster, svcAddr, clusterName str
 		SecurePeer:   ec.Spec.TLS.IsSecurePeer(),
 		SecureClient: ec.Spec.TLS.IsSecureClient(),
 	}
+	if ec.Spec.Pod != nil {
+		m.ClusterDomain = ec.Spec.Pod.ClusterDomain
+	}
 	ms := etcdutil.NewMemberSet(m)
 	backupURL := backupapi.BackupURLForRestore("http", svcAddr, clusterName)
 	ec.SetDefaults()
--- a/pkg/util/etcdutil/member.go
+++ b/pkg/util/etcdutil/member.go
@@ -33,10 +33,13 @@ type Member struct {
 
 	SecurePeer   bool
 	SecureClient bool
+
+	// ClusterDomain is the DNS name of the cluster. E.g. .cluster.local.
+	ClusterDomain string
 }
 
 func (m *Member) Addr() string {
-	return fmt.Sprintf("%s.%s.%s.svc", m.Name, clusterNameFromMemberName(m.Name), m.Namespace)
+	return fmt.Sprintf("%s.%s.%s.svc%s", m.Name, clusterNameFromMemberName(m.Name), m.Namespace, m.ClusterDomain)
 }
 
 // ClientURL is the client URL for this member


hasbro17:
@etcd-bot ok to test

mikkeloscar:
Any update on this?

hasbro17:
@mikkeloscar Sorry for the delay on this. Appending the ClusterDomain seems okay to me, however I'm wondering how this is fixed via the DNS config?

Also I'm considering if we should pass in the ClusterDomain suffix as a flag to the operator, vs specifying it as a field in the [PodPolicy](https://github.com/coreos/etcd-operator/blob/master/pkg/apis/etcd/v1beta2/cluster.go#L105) of the EtcdCluster CR i.e `spec.pod.clusterDomainSuffix`?

Currently the restore-operator also creates a new seed member for an existing EtcdCluster. And with this change it won't be setting the member's ClusterDomain suffix:
https://github.com/coreos/etcd-operator/blob/master/pkg/controller/restore-operator/sync.go#L214-L223

So instead of passing in another flag to the restore-operator we can set it once in the EtcdCluster PodPolicy and use it across the two operators.
WDYT?

mikkeloscar:
@hasbro17 thanks for the feedback

> @mikkeloscar Sorry for the delay on this. Appending the ClusterDomain seems okay to me, however I'm wondering how this is fixed via the DNS config?

In our setup we default to `ndots:2` but if the user specifies another DNS config on the pod then we respect that, so you would be able to "opt-out" by specifying:

```yaml
dnsConfig:
    options:
      - name: ndots
        value: "5"
```

Obviously you wouldn't benefit from reducing the number of useless DNS queries, which is why I think setting the suffix is better.

> So instead of passing in another flag to the restore-operator we can set it once in the EtcdCluster PodPolicy and use it across the two operators.
> WDYT?

My motivation for setting this as a flag is that you would usually want to have the same clusterDomain for all etcd clusters within a single Kubernetes cluster, so it's simpler to set it centrally rather than configure it for each etcd cluster.

If you prefer the podPolicy solution then I could add it there instead. Otherwise I would add the flag to the restore-operator as well.

hasbro17:
@mikkeloscar Thanks for the explanation.

I would still lean towards setting the ClusterDomain in the EtcdCluster's PodPolicy just so the restore operator can reuse the existing EtcdCluster's setting.
Plus making it part of the API would more clearly document this setting for each cluster.

PS can you also please add a line to the [CHANGELOG](https://github.com/coreos/etcd-operator/blob/master/CHANGELOG.md#added) to mention the new API.

mikkeloscar:
@hasbro17 ok, I changed it now to use a `ClusterDomain` field on the `PodPolicy`.

pkg/cluster/cluster.go
```diff
@@ -311,10 +311,11 @@ func isSpecEqual(s1, s2 api.ClusterSpec) bool {
...
 	m := &etcdutil.Member{
-		Name:         k8sutil.UniqueMemberName(c.cluster.Name),
-		Namespace:    c.cluster.Namespace,
-		SecurePeer:   c.isSecurePeer(),
-		SecureClient: c.isSecureClient(),
+		Name:          k8sutil.UniqueMemberName(c.cluster.Name),
+		Namespace:     c.cluster.Namespace,
+		SecurePeer:    c.isSecurePeer(),
+		SecureClient:  c.isSecureClient(),
+		ClusterDomain: c.cluster.Spec.Pod.ClusterDomain,
```
hasbro17: Check if `c.cluster.Spec.Pod` is `nil` before setting this field. Otherwise we get a segfault:

```console
time="2019-05-09T20:39:40Z" level=info msg="etcd-operator Version: 0.9.4+git"
time="2019-05-09T20:39:40Z" level=info msg="Git SHA: 8d9b1e5d"
time="2019-05-09T20:39:40Z" level=info msg="Go Version: go1.10"
time="2019-05-09T20:39:40Z" level=info msg="Go OS/Arch: linux/amd64"
time="2019-05-09T20:39:40Z" level=info msg="Event(v1.ObjectReference{Kind:\"Endpoints\", Namespace:\"e2e-etcd-operator-e2e-pr-3371\", Name:\"etcd-operator\", UID:\"91df3254-729a-11e9-8ce3-42010a800271\", APIVersion:\"v1\", ResourceVersion:\"111408674\", FieldPath:\"\"}): type: 'Normal' reason: 'LeaderElection' etcd-operator became leader"
time="2019-05-09T20:39:45Z" level=info msg="creating cluster with Spec:" cluster-name=test-etcd-lntbc cluster-namespace=e2e-etcd-operator-e2e-pr-3371 pkg=cluster
time="2019-05-09T20:39:45Z" level=info msg="{" cluster-name=test-etcd-lntbc cluster-namespace=e2e-etcd-operator-e2e-pr-3371 pkg=cluster
time="2019-05-09T20:39:45Z" level=info msg="    \"size\": 3," cluster-name=test-etcd-lntbc cluster-namespace=e2e-etcd-operator-e2e-pr-3371 pkg=cluster
time="2019-05-09T20:39:45Z" level=info msg="    \"repository\": \"quay.io/coreos/etcd\"," cluster-name=test-etcd-lntbc cluster-namespace=e2e-etcd-operator-e2e-pr-3371 pkg=cluster
time="2019-05-09T20:39:45Z" level=info msg="    \"version\": \"3.2.13\"" cluster-name=test-etcd-lntbc cluster-namespace=e2e-etcd-operator-e2e-pr-3371 pkg=cluster
time="2019-05-09T20:39:45Z" level=info msg="}" cluster-name=test-etcd-lntbc cluster-namespace=e2e-etcd-operator-e2e-pr-3371 pkg=cluster
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x98 pc=0x10c9589]

goroutine 73 [running]:
github.com/coreos/etcd-operator/pkg/cluster.(*Cluster).startSeedMember(0xc4202345a0, 0xc42048fe00, 0x30)
	/go/src/github.com/coreos/etcd-operator/pkg/cluster/cluster.go:318 +0x149
github.com/coreos/etcd-operator/pkg/cluster.(*Cluster).bootstrap(0xc4202345a0, 0x0, 0x3)
	/go/src/github.com/coreos/etcd-operator/pkg/cluster/cluster.go:344 +0x2b
github.com/coreos/etcd-operator/pkg/cluster.(*Cluster).prepareSeedMember(0xc4202345a0, 0x0, 0x0)
	/go/src/github.com/coreos/etcd-operator/pkg/cluster/cluster.go:167 +0x5a
github.com/coreos/etcd-operator/pkg/cluster.(*Cluster).create(0xc4202345a0, 0xc420472f60, 0x10d95a7)
	/go/src/github.com/coreos/etcd-operator/pkg/cluster/cluster.go:161 +0x111
github.com/coreos/etcd-operator/pkg/cluster.(*Cluster).setup(0xc4202345a0, 0xc4202bfa70, 0xc420472f80)
	/go/src/github.com/coreos/etcd-operator/pkg/cluster/cluster.go:148 +0x83
github.com/coreos/etcd-operator/pkg/cluster.New.func1(0xc4202345a0)
	/go/src/github.com/coreos/etcd-operator/pkg/cluster/cluster.go:105 +0x35
created by github.com/coreos/etcd-operator/pkg/cluster.New
	/go/src/github.com/coreos/etcd-operator/pkg/cluster/cluster.go:104 +0x33b
```

mikkeloscar:
@hasbro17 added the `nil` checks!

mikkeloscar:
Can you help me understand why `jenkins-ci` is failing?

I can build it locally:

```
./hack/build/build
building operator...
building backup-operator...
building restore-operator...
```

hasbro17:
@etcd-bot retest this please

mikkeloscar:
@hasbro17 thanks!

mikkeloscar:
Friendly ping! :) If you need anything from me let me know!

hasbro17:
@mikkeloscar Just waiting on @hexfusion but I think we'll be able to merge this later today.
Apologies for the delay but can you please rebase this again?

mikkeloscar:
> Apologies for the delay but can you please rebase this again?

Done!

hexfusion:
@mikkeloscar one last thing can we make this a single commit something like below.

`*: Allow specifying Cluster Domain for clientURLs`

After this is done we can merge right away.

mikkeloscar:
There you go!

aermakov-zalando:
@hexfusion 
Is there an ETA on when this would make it into a release?

PR #2092, opened 2019-06-10
Add curl image parameter

atiannicelli:
cluster: add option to specify curl image

add an option in EtcdCluster.spec:
CurlImage. Defaults to "tutum/curl:latest"

see https://github.com/coreos/etcd-operator/issues/1933

### DIFF
--- a/pkg/apis/etcd/v1beta2/cluster.go
+++ b/pkg/apis/etcd/v1beta2/cluster.go
@@ -18,7 +18,7 @@ import (
 	"errors"
 	"strings"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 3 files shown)


marcostvz:
hi @hasbro17, I need this feature too, could you or any other maintainer take a look? Thanks! :)

pkg/apis/etcd/v1beta2/cluster.go
```diff
@@ -162,6 +162,11 @@ type PodPolicy struct {
 	// '.cluster.local'.
 	// The default is to not set a cluster domain explicitly.
 	ClusterDomain string `json:"ClusterDomain"`
+
+	// curl init container image. default is tutum/curl.
+	// pulling tutum/curl:latest requires external access.
+	// More info: https://github.com/coreos/etcd-operator/issues/1933
+	CurlImage string `json:"curlImage,omitempty"`
```
alaypatel07: Better to add this to `EtcdRestore` struct

vaneyckt:
hi @hasbro17, I just ran into this today as well. What's the status of this PR? @atiannicelli do you need someone to take over from you on this?

atiannicelli:
We have been using this change for a few months and it works fine. I don't know how to "take it from here" as I don't have merge ability and it seems that the tests are not being run above. I don't know who to contact to get an approving review from someone with write access.

pkg/apis/etcd/v1beta2/cluster.go
alaypatel07: I thought the `busyboxImage` parameter is used for all etcd cluster pods and not just etcd restore https://github.com/coreos/etcd-operator/blob/b2a63ce924602be64a156361f3ee85498843026a/pkg/util/k8sutil/k8sutil.go#L374

PR #2099, opened 2019-07-02
fix sort, with a helper so the sort is done 'naturally'

jurgenweber:

I hit the edge case that when it sorts, it things

etcd_v1066060_2019-07-02-02:51:51
is smaller/older than
etcd_v999523_2019-06-30-14:56:49

because it is sorting on the first digit only.

Added all the stuff needed to sort this properly.

### DIFF
--- a/pkg/backup/backup_manager.go
+++ b/pkg/backup/backup_manager.go
@@ -19,7 +19,10 @@ import (
 	"crypto/tls"
 	"fmt"
 	"sort"
+	"strconv"
+	"strings"
 	"time"
+	"unicode"
 
 	"github.com/coreos/etcd-operator/pkg/backup/writer"
 	"github.com/coreos/etcd-operator/pkg/util/constants"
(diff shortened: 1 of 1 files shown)


jurgenweber:
so I built the image and used it overnight, it works as expected now.

Thanks

pkg/backup/backup_manager.go
```diff
@@ -83,13 +86,69 @@ func (bm *BackupManager) SaveSnap(ctx context.Context, s3Path string, isPeriodic
...
+	}
+	sort.Sort(strSort)
+}
+
+type strSorter struct {
+	strs []string
+	cmp  func(str1, str2 string) bool
+}
+
+func extractNumberFromString(str string, size int) (num int) {
```
daxmc99: Might be worth mentioning where you got his function from?  Appears to be a direct copy from https://www.socketloop.com/tutorials/golang-natural-string-sorting-example

jurgenweber: yeah, fair enough. What is the etiquette here? A comment and the link.

PR #2103, opened 2019-07-24
Add tmpfs option for etcd-data volume

### LINKED ISSUE #2053: Option to set emptyDir.medium to "Memory"
For testing purposes I'd like to be able to specify to use 
```yaml
emptyDir:
  medium: "Memory"
```

When leaving out the PVC from `pod` it's using an emptyDir already. Can we add the option to use that from Memory?

chriscoester:
This adds an additional optional field to the etcdcluster.spec.pod field, allowing etcdpods to use tmpfs for etcd-data.

Fixes #2053

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -377,9 +377,9 @@ func (c *Cluster) createPod(members etcdutil.MemberSet, m *etcdutil.Member, stat
 		if err != nil {
 			return fmt.Errorf("failed to create PVC for member (%s): %v", m.Name, err)
 		}
-		k8sutil.AddEtcdVolumeToPod(pod, pvc)
+		k8sutil.AddEtcdVolumeToPod(pod, pvc, false)
 	} else {
-		k8sutil.AddEtcdVolumeToPod(pod, nil)
+		k8sutil.AddEtcdVolumeToPod(pod, nil, c.cluster.Spec.Pod.Tmpfs)
 	}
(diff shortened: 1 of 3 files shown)


Marlinc:
Any chance someone could look into this? Anything I can do to help with getting it in?

pkg/cluster/cluster.go
```diff
@@ -377,9 +377,9 @@ func (c *Cluster) createPod(members etcdutil.MemberSet, m *etcdutil.Member, stat
 		if err != nil {
 			return fmt.Errorf("failed to create PVC for member (%s): %v", m.Name, err)
 		}
-		k8sutil.AddEtcdVolumeToPod(pod, pvc)
+		k8sutil.AddEtcdVolumeToPod(pod, pvc, false)
 	} else {
-		k8sutil.AddEtcdVolumeToPod(pod, nil)
+		k8sutil.AddEtcdVolumeToPod(pod, nil, c.cluster.Spec.Pod.Tmpfs)
```
Marlinc: I have tested this and till will cause the operator to crash with a nil error when `c.cluster.Spec.Pod` is nil.

Marlinc: I have tested this and till will cause the operator to crash with a nil error when `c.cluster.Spec.Pod` is nil.

Marlinc: I've pushed a fix in cbws/etcd-operator#22

PR #2123, opened 2019-10-05
chore: add sas token support for abs.

andyliuliming:
in some enterprise scenario, we would not give whole storage account permission to the etcd operator. we need to support the scenario only some container permission granted.


andyliuliming:
@hongchaodeng could you please help do a review? thanks.

andyliuliming:
Hi @hexfusion , could you please help do a review? thank very much.

PR #2108, opened 2019-08-05
tuning the liveness check time of etcd pod

armstrongli:
liveness probe needs to take care of the following steps' time:
* time to transmit a 4GB snapshot from leader to member -- 4000MB / 50MB/s = 80s
* time to load snapshot —— 4000MB / 200MB/s = 20s
* time to set up streams and follow up leader revisions -- 100s * 10 heartbeat/s * 1KB/heartbeat / 1024KB/MB / 50MB/s ~= 20s

total time -- 120s

Fix: https://github.com/coreos/etcd-operator/issues/2109

Please read https://github.com/coreos/etcd-operator/blob/master/CONTRIBUTING.md#contribution-flow

PR #2124, opened 2019-10-05
etcd: expose more configs for the etcd pods.

andyliuliming:
people may have requirements that have control on the etcd behaviour in more details.

### DIFF
--- a/pkg/apis/etcd/v1beta2/cluster.go
+++ b/pkg/apis/etcd/v1beta2/cluster.go
@@ -64,6 +64,26 @@ func (c *EtcdCluster) AsOwner() metav1.OwnerReference {
 	}
 }
 
+type EtcdConfig struct {
+	// Heartbeat timeout setting for etcd pod
+	HeartbeatTimeout int `json:"heartbeatTimeout,omitempty"`
+
+	// Election timeout setting for etcd pod
+	ElectionTimeout int `json:"electionTimeout,omitempty"`
+
+	// Snapshot count setting for etcd pod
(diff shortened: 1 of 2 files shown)


andyliuliming:
this will also address the issue:
https://github.com/coreos/etcd-operator/pull/2108

andyliuliming:
Hi @hexfusion , could you please help do a review? thank very much.

ArneMancofi:
Will this fix be released, and if so, when can we expect it?
We are having timeout problems every time an etcd pod is created, due to disk attachment times; is there any workaround possible to increase the total liveness and readyness timeouts?

pkg/apis/etcd/v1beta2/cluster.go
```diff
@@ -64,6 +64,39 @@ func (c *EtcdCluster) AsOwner() metav1.OwnerReference {
...
 }
 
+type ProbeConfig struct {
+	InitialDelaySeconds int `json:"initialDelaySeconds,omitempty"`
+	TimeoutSeconds      int `json:"timeoutSeconds,omitempty"`
+	PeriodSeconds       int `json:"periodSeconds,omitempty"`
+	FailureThreshold    int `json:"failureThreshold,omitempty"`
+}
+
+type EtcdConfig struct {
```
nvtkaszpir: most of this parameters can be controlled directly via env vars, and etcd-opreator already supports in as `env` vars in .spec.pod.etcdEnv.

I would leave in here only k8s specific ones - liveness and readiness probes, but also would rather reconsider it and put that in pod spec.

PR #2127, opened 2019-10-10
Fix no matches for kind "Deployment" in version "extensions/v1beta1"

longfei-zhang:
As deployment in the extensions/v1beta1 was deprecated from
K8S 1.16, so change the deployment to suit K8S 1.16 version
and later.

Ref: https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/


oliverhu:
Verified and works for me.

Issue #1410, opened 2017-09-17
Finish informer work


hongchaodeng:
Since upstream provides good examples of informer for CRD, we made use of that and already switched our watching code to informer. But some work is left to be done:

- [ ] Use workqueue. Nothing should block event handler.
- [ ] [here](https://github.com/coreos/etcd-operator/blob/master/pkg/controller/informer.go) converts the informer event to internal event. We shouldn't need the duplicate events. Should clean it up . Otherwise it is confusing.

PR #2151, opened 2020-01-18
fix use workqueue. nothing should block event handler(#1410)

xigang:


xigang:
cc @hongchaodeng @hexfusion please review:)

Issue #2156, opened 2020-02-03
etcd-operator-builder deleted?


dbavatar:
Can no longer build from master, builder container is gone?

$ docker pull gcr.io/coreos-k8s-scale-testing/etcd-operator-builder:0.4.1-2
Error response from daemon: Get https://gcr.io/v2/coreos-k8s-scale-testing/etcd-operator-builder/manifests/0.4.1-2: denied: Project coreos-k8s-scale-testing has been deleted.

Still referenced here:
https://github.com/coreos/etcd-operator/blob/master/hack/build/build

PR #2158, opened 2020-02-11
Provide documentation on etcd-operator-builder.  Remove references to gcr.io image.

pjsully:
This addresses #2156 and likely will close it.

if there is a need to use the builder image just build it locally and push it where you need it.

Added documentation for generating builder and using it.


pjsully:
Note that for the ci build script to work, etcd-operator-builder:latest has to be available from the local docker registry...not ideal, but gives a way to make this work (and steps to build the builder are part of this PR as well...)

